PR #9, opened 2015-02-05
readd jshintrc

fippo:
jshintrc was AWOL


andresusanopinto:
Samples is now on its own repository with its own set of linting rules.
So there is no need to change the adapter.js inline settings whenever we change code that depends on this repo.

fippo:
I tend to disagree in general, mostly because a jshintrc is part of my standard skeleton for new projects.

However, for adapter.js it actually makes sense to inline this under the assumption that people will download just adapter.js and it shouldn't break their jshint runs.

PR #13, opened 2015-02-05
allow adapter to be required

fippo:
minor change from the previous PR: I don't think it makes sense to export trace

### DIFF
--- a/adapter.js
+++ b/adapter.js
@@ -8,8 +8,8 @@
 
 /* More information about these options at jshint.com/docs/options */
 /* jshint browser: true, camelcase: true, curly: true, devel: true,
-   eqeqeq: true, forin: false, globalstrict: true, quotmark: single,
-   undef: true, unused: strict */
+   eqeqeq: true, forin: false, globalstrict: true, node: true,
+   quotmark: single, undef: true, unused: strict */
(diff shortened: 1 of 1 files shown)


fippo:
hey @googlebot, didn't you see I was already a contributor. But I've updated my CLA and will get you a corporate one even ;-)

adapter.js
```diff
@@ -228,3 +228,16 @@ function requestUserMedia(constraints) {
...
+
+if (typeof module !== 'undefined') {
+  module.exports = {
+    RTCPeerConnection: RTCPeerConnection,
+    getUserMedia: getUserMedia,
+    attachMediaStream: attachMediaStream,
+    reattachMediaStream: reattachMediaStream,
+    webrtcDetectedBrowser: webrtcDetectedBrowser,
+    webrtcDetectedVersion: webrtcDetectedVersion,
+    requestUserMedia: requestUserMedia
```
juberti: i would prefer that requestUserMedia not be used as well, this should become an overload of getUserMedia

juberti:
lgtm with one comment

fippo:
Done. commented out so they're not accidentially added again.

PR #16, opened 2015-02-10
Fix bower metadata (missing ignore).

andresusanopinto:
Bower was complaining with:
 bower adapter.js#\*        invalid-meta adapter.js is missing "ignore" entry in bower.json

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -10,6 +10,7 @@
     "Jesús Leganés Combarro \"piranna\" <piranna@gmail.com>"
   ],
   "main": "adapter.js",
+  "ignore": [],
   "moduleType": [
     "globals"
   ],


andresusanopinto:
@piranna - Is it common to always have a ignore, even if empty?
I was a bit surprised with bower complaining as "invalid-meta" for missing it.

piranna:
> Is it common to always have a ignore, even if empty?

Honestly I'm not sure, I'm not a bower guy, sorry... :-/

> I was a bit surprised with bower complaining as "invalid-meta" for missing it.

There are other fields with similar warning messages on bower.json, there are some issues regarding to don't make it so alarmist... ;-) I wouldn't worry about it.

andresusanopinto:
Ok... will not merge then... noticed most polymer and paper bower packages complaing on a lot of missing "Recommended" (but not required fields).. such as main, ignore list, etc...

PR #17, opened 2015-02-11
Renamed bower module to webrtc-adapter.

andresusanopinto:
I believe `webrtc-adapter` is a better name than the existent `adapter.js`
Not only makes the bower install more meaningful. (e.g. no one can guess what `adapter.js` package is from).
But it also removes the `.js` from the name avoiding directories with file-like names such as:
`<script src="bower_components/adapter.js/adapter.js">`


andresusanopinto:
@samdutton @juberti - Giving that we have recently changed the adapter.js location, the repo and so on.. I think this is the right moment to fix the package name.

PR #18, opened 2015-03-10
firefox 38+ supports .urls for RTCIceServer

fippo:
Made me wonder if some of the "cruft" can be cleaned up. I don't think I care about FF < 27...

For Chrome, apprtc doesn't care about versions < 33 that don't suppport .urls, since it doesn't call createIceServer, only createIceServer**s**


adapter.js
```diff
@@ -116,16 +118,24 @@ if (navigator.mozGetUserMedia) {
...
 
   window.createIceServers = function(urls, username, password) {
-    var iceServers = [];
-    // Use .url for FireFox.
-    for (var i = 0; i < urls.length; i++) {
-      var iceServer =
-        window.createIceServer(urls[i], username, password);
-      if (iceServer !== null) {
-        iceServers.push(iceServer);
+    if (webrtcDetectedVersion < 38) {
```
juberti: since we do the fixup in the ctor, do we need any of this anymore?

fippo: the behaviour differs since createIceServers also tries to fixup old cruft like ?transport stuff in createIceServer (for firefox -- it doesn't for chrome). I do not want to think about what else breaks with versions as old as chrome 33 and firefox 27...

I'd tend to just fix it up in the constructor for FF < 38 and remove createIceServer(s).

juberti: that's my thinking as well. probably will also need an apprtc update to stop using createIceServers.

fippo:
surprise... the FF constructor didn't do a proper urls->url expansion. I don't fully understand what it was doing, just urls->url does not make much sense.
Probably relied on createIceServers to do the right thing. Anyway...

have an apprtc patch to go along with this.

fippo:
just noticed that technically .urls is

```
(DOMString or sequence<DOMString>)
```

the current code always expands to an array, since this is temporary I'd leave it at that. @juberti: ping

alvestrand:
Hi, this has been sleeping for a while, but not forgotten.... I have just added some testing framework that should at least tell if the javascript's parseable; would you mind rebasing and seeing if you can add a simple check that this code does what you want?

alvestrand:
(sorry, tests, not merged yet - please check that PR #25 is merged before resubmitting)

fippo:
ah... I think i got the rebase wrong. let me resubmit...

PR #20, opened 2015-03-12
jshint 2.6.3 doesnt define Promise anymore

fippo:
which broke the build over in #18 for reasons unrelated to that PR.


samdutton:
LGTM

PR #23, opened 2015-04-09
mediaDevices.enumerateDevices() polyfill in adapter.js

jan-ivar:
Add a mediaDevices.enumerateDevices polyfill for Chrome using existing MediaStreamTrack.getSources().

### DIFF
--- a/adapter.js
+++ b/adapter.js
@@ -208,9 +208,9 @@ if (navigator.mozGetUserMedia) {
     to.src = from.src;
   };
 
-  navigator.mediaDevices = navigator.mediaDevices || {
-    getUserMedia: requestUserMedia,
-    enumerateDevices: function() {
+  if (!navigator.mediaDevices) {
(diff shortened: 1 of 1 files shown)


fippo:
@jan-ivar do you think it's possible to shim MST.getSources with this? Would be nice if people who used it would get the functionality and a warning :-)

jan-ivar:
@fippo sorry just saw your comment. I'd rather not shim non-standard APIs because I worry it works against convergence by allowing people to never update. We want to converge on the spec, not on something else.

For instance, my latest patch here actually removes the MST.getSources() shim from Firefox (that never did anything) in favor of the spec one.

jan-ivar:
Don't people normally copy adapter.js anyway? (I would think linking directly to it would be a cross-origin warning in many browsers).

jan-ivar:
@juberti - please let me know if anything else is needed here as well. I'm excited for feedback.

alvestrand:
Guess this too needs a rebase and a test....

PR #25, opened 2015-05-22
Adding Javascript tests using testling.

alvestrand:
This gives the hope of doing a cross-browser test.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -6,6 +6,6 @@ m = require('../adapter.js');
 
 test('Browser identified', function(t) {
   t.plan(2);
-  t.ok(m.webrtcDetectedBrowser);
-  t.ok(m.webrtcDetectedVersion);
+  t.ok(m.webrtcDetectedBrowser, 'Browser detected');
+  t.ok(m.webrtcDetectedVersion, 'Browser version detected');
 });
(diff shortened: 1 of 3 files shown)


test/test.js
```diff
@@ -0,0 +1,11 @@
+// This is a basic test file for use with testling.
+// The test script language comes from tape.
+var test = require('tape');
+
+m = require('../adapter.js');
+
+test('Browser identified', function(t) {
+  t.plan(2);
+  t.ok(m.webrtcDetectedBrowser);
```
phoglund: This currently prints 

ok 1 (unnamed assert)
ok 2 (unnamed assert)

which isn't very helpful, so I suggest you add a description (you can apparently do that as the second argument to ok). For instance, 'adapter.js should detect a browser'.

Have you ensure the test fails, i.e. that webrtcDetectedBrowser is indeed false if we don't detect a browser?

phoglund:
One comment

alvestrand:
I made the messages a little simpler "adapter.js should" is kind of implicit, since adapter.js is what it tests.

I also checked that if I misspell the variable name, the test fails. Should be good enough to demonstrate that the framework does what it should.

phoglund:
lgtm

Issue #19, opened 2015-03-11
adapter.js versioning policy + minimum required versions


fippo:
In order to make adapter.js dependable, there needs to be some kind of versioning system.

Currently, it is rather unclear what you need to do when upgrading adapter which keeps even me from using it. In the past, things have been removed without notice. And I want to do that in this PR https://github.com/webrtc/adapter/pull/18#discussion_r26180825 as well, so this is not a theoretic problem.

Let's use [semver](http://semver.org) as follows:
Patch version: backward compatible fixes you absolutely want to pull in and that don't break anything.
Minor version: addition of new features.
Major version: removal of old APIs/functionality/workarounds in adapter.js.

The major version is a little tricky. We use adapter to shim functionality that is yet there or only exist newer versions of Chrome and Firefox. But we don't want to keep "old cruft" around forever. For example the shim of MediaStream.prototype.getAudioTracks (mid-2013)... 

On talky, I am seeing chrome versions which do not support DTLS and some other ... interesting... errors. I do not want to support those versions anymore. If you use a 6+ month old browser you are probably getting an experience which is worse than what you should get...
Chromium 37 (possibly older versions of Chrome as well) seem to have an [issue with opusfec](https://code.google.com/p/webrtc/issues/detail?id=4378).

Hence we should also require a minimum browser version (both FF and Chrome) and bump the major version if we change that.

thoughts @samdutton? (cc @bemasc @damonoehlman)
The final version will be an empty file because we don't need a shim anymore! Not sure if that is a realistic goal but... :-)

samdutton:
SGTM, with semver versioning: starting with adapter-1.0.0.js?

Would we keep old versions or just rename?

@juberti @jiayliu WDYT?

fippo:
@samdutton yes. Unless you want a fancy name that requires you to start at 2.0.0 :-)

Using npm for versioning automagically creates git tags which appear as releases on github (but only if one pushes with --tags...). I suppose bower can do something similar.

samdutton:
I think annotated git tags are probably the way to go, using the mechanism described [here](https://help.github.com/articles/creating-releases/). Creating the tags manually is probably OK.

@juberti @jiayliu If no objections, I'll go ahead and add tag/release semver versioning, starting with 1.0.0.

fippo:
@samdutton I can do a PR with the minimum version thing over the weekend

samdutton:
SGTM

fippo:
[...]

PR #26, opened 2015-05-26
firefox 38+ supports .urls for RTCIceServer

fippo:
see discussion in issue #18

@alvestrand: not sure if there is any way to really test the constructor. The only thing I can think of is "does it not throw an exception?"

### DIFF
--- a/adapter.js
+++ b/adapter.js
@@ -21,6 +21,7 @@ var attachMediaStream = null;
 var reattachMediaStream = null;
 var webrtcDetectedBrowser = null;
 var webrtcDetectedVersion = null;
+var webrtcMinimumVersion = null;
 
 function trace(text) {
   // This function is used for logging.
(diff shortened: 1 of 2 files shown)


adapter.js
```diff
@@ -79,55 +100,6 @@ if (navigator.mozGetUserMedia) {
     }, 0);
   };
 
-  // Creates ICE server from the URL for FF.
-  window.createIceServer = function(url, username, password) {
```
alvestrand, line 83: You're withdrawing an utility function. Was this ever used by users? Do we have any way of knowing?

alvestrand:
Sure, you can do some tests. Call "new RTCPeerConnection()" and see that you get back an Object, that has the setLocalDescription function defined. (Because of the nature of the shim, I don't think you can test the actual type of the returned object; it won't be RTCPeerConnection.)
This PR also exposes webrtcMinimumVersion - add a test saying that it has been defined.

I've done so many silly errrors in coding, I always add a really, really basic test.....

adapter.js
fippo, line 83: the only user I know of is apprtc here (and it only uses createIceServers) https://github.com/webrtc/apprtc/blob/master/src/web_app/js/util.js#L96-97
-- I had a PR to remove that when updating adapter.

Breaking changes are not good obviously, see issue #19 for that.

fippo:
Good point. Also that allows checking that the detected version is actually supported by adapter.js.

alvestrand:
#18 said that you were adding version numbers here, but I can't see them.
Still, merging. It looks good.

PR #27, opened 2015-05-26
add jscs

fippo:
@alvestrand more linting, copied over from the samples repo.
That should hopefully prevent errors when merging adapter there.

### DIFF
--- a/.jscsrc
+++ b/.jscsrc
@@ -1,0 +1,3 @@
+{
+  "preset": "google"
+}
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ before_script:
   - export DISPLAY=:99.0
   - sh -e /etc/init.d/xvfb start
 
--script:
+script:
   - grunt
 
 after_failure:
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,19 +19,14 @@ matrix:
     - env: BROWSER=chrome  BVER=unstable
     - env: BROWSER=firefox BVER=unstable
 
-install:
-  - npm install -g jshint jscs
-  - npm install
-
 before_script:
   - ./node_modules/travis-multirunner/setup.sh
   - export DISPLAY=:99.0
   - sh -e /etc/init.d/xvfb start
 
-script:
-   - jshint *.js
-   - jscs --preset google *.js
-   - npm test
+-script:
+  - grunt
 
 after_failure:
   - for file in *.log; do echo $file; echo "======================"; cat $file; done || true
+
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ before_script:
   - sh -e /etc/init.d/xvfb start
 
 script:
-  - grunt
+  - node_modules/.bin/grunt
 
 after_failure:
   - for file in *.log; do echo $file; echo "======================"; cat $file; done || true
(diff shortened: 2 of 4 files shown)


alvestrand:
Does this actually cause Travis to run jscs?

I have a call directly on jscs in the .travis.yml file, but I do _not_ have a call on "grunt" in the script section.

I'm not wedded to the method I currently use to invoke jscs, but I want Travis to do it, and I only want it to be done once.

Can you fix .travis.yml so that the Right Thing Happens?

fippo:
it did not... I assumed you had copied that over from the samples. It should now. 
Everything is running from grunt now. Not wedded either, but that's what the samples and apprtc do so...

alvestrand:
As a matter of curiosity (I'll find out when I merge this) - what happens when you run grunt locally?
Do you have to set BROWSER and BVER to make it complete?

(I've thought about making a contribution to travis-multirunner that stops it from polluting ~/bin and the unit directory when it runs - it does not make me happy.)

fippo:
Running grunt will make testling start the default browser. Good enough for presubmit checks. BROWSER and BVER are only used when running testling with -x I think

alvestrand:
The weird thing is that when I look at the output from the Travis run that purports to be from this version:

https://travis-ci.org/webrtc/adapter/jobs/64169882

it still runs "npm test", not "grunt". And jscs - the title of the CL - isn't visible in the log at all.
Is this a Travis weirdness, or is something still wrong?

fippo:
Odd indeed. npm test seems to be the default thing to run. I can reproduce something similar locally, running `npm i` removes the `script: -grunt` from the .travis.yml

@damonoehlman: shouldn't the travis-sync script in travis-multirunner check for an existing .travis.yml?

DamonOehlman:
Hey All - apologies for this, I very selfishly made `travis-multirunner` suit my own particular use case by overwriting the `.travis.yml` file (which does seem very strange in retrospect, but when porting a repo to using it did make sense at the time).  Anyway...

That's all fixed now and published as `travis-multirunner@2.2.0` - glad to see it's being useful to someone other than just myself these days :)

fippo:
uh... I had '-script:' instead of 'script:'. Thanks diff. And then no grunt-cli. But now it works:
https://travis-ci.org/webrtc/adapter/jobs/64364528#L405

alvestrand:
Still something strange going on.

https://travis-ci.org/webrtc/adapter/jobs/64364528

BROWSER=chrome

"This appears to be firefox"

Seems that testling must be invoked with "-x start-$BROWSER" in order to do its thing right - and it isn't.

DamonOehlman:
Yes, this seems to be a limitation in the `grunt-testling` package unfortunately which only spawns testling with empty args: 

https://github.com/bevacqua/grunt-testling/blob/master/tasks/testling.js#L13

fippo:
ah... this looked so easy :-/

easiest way out is to do jshint + jscs with grunt (like the samples) and run npm test afterwards, right? Which puts us where we started with a little more grunt and a bugfix. Or we just get rid of the grunt dep...

alvestrand:
You're caught in a twisty little maze of frameworks, all incompatible :-)
Yes, I think invoking both grunt and npm test explicitly should do the trick.
Don't know if grunt can have a target calling npm test as part of its default, but that bit of polish can wait until another day.

fippo:
right-hand rule to the rescue, that looks better.
I assume travis-semantic mean that if any of the scripts fail it with a non-zero status, the build fails.

alvestrand:
NOW I'm happy!

PR #28, opened 2015-05-26
add basic p2p test

fippo:
very basic test, if that works all the RTCSessionDescription, RTCIceCandidate shimming for Mozilla works.

I really wonder why adapter doesn't put RTCPeerConnection on window as well...

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -1,11 +1,101 @@
+'use strict';
 // This is a basic test file for use with testling.
 // The test script language comes from tape.
+/* jshint node: true */
 var test = require('tape');
 
-m = require('../adapter.js');
+var m = require('../adapter.js');
 
 test('Browser identified', function(t) {
   t.plan(2);
   t.ok(m.webrtcDetectedBrowser, 'Browser detected');
   t.ok(m.webrtcDetectedVersion, 'Browser version detected');
+});
+
+test('basic connection establishment', function(t) {
+  var pc1 = new m.RTCPeerConnection(null);
+  var pc2 = new m.RTCPeerConnection(null);
+
+  pc1.createDataChannel('somechannel');
+  pc1.oniceconnectionstatechange = function() {
+    if (pc1.iceConnectionState === 'connected') {
+      t.pass('P2P connection established');
+      t.end();
+    }
+  };
+
+  var addCandidate = function(pc, event) {
+    if (event.candidate) {
+      var cand = JSON.parse(JSON.stringify(event.candidate));
+      cand = new RTCIceCandidate(cand);
+      pc.addIceCandidate(cand,
+        function() {
+        },
+        function(err) {
+          t.fail('addIceCandidate ' + err.toString());
+        }
+      );
+    }
+  };
+  pc1.onicecandidate = function(event) {
+    addCandidate(pc2, event);
+  };
+  pc2.onicecandidate = function(event) {
+    addCandidate(pc1, event);
+  };
+
+  pc1.createOffer(
+    function(offer) {
+      pc1.setLocalDescription(offer,
+        function() {
+          t.pass('pc1.setLocalDescription');
+        },
+        function(err) {
+          t.fail('pc1.setLocalDescription ' + err.toString());
+        }
+      );
+
+      offer = JSON.parse(JSON.stringify(offer));
+      offer = new RTCSessionDescription(offer);
+      t.pass('created RTCSessionDescription from offer');
+      pc2.setRemoteDescription(offer,
+        function() {
+          t.pass('pc2.setRemoteDescription');
+          pc2.createAnswer(
+            function(answer) {
+              t.pass('pc2.createAnswer');
+              pc2.setLocalDescription(answer,
+                function() {
+                  t.pass('pc2.setLocalDescription');
+                },
+                function(err) {
+                  t.fail('pc2.setLocalDescription ' + err.toString());
+                }
+              );
+              answer = JSON.parse(JSON.stringify(answer));
+              answer = new RTCSessionDescription(answer);
+              t.pass('created RTCSessionDescription from answer');
+              pc1.setRemoteDescription(answer,
+                function() {
+                  t.pass('pc1.setRemoteDescription');
+                },
+                function(err) {
+                  t.fail('pc1.setRemoteDescription ' + err.toString());
+                }
+              );
+            },
+            function(err) {
+              t.fail('pc2.createAnswer ' + err.toString());
+            }
+      );
+        },
+        function(err) {
+          t.fail('pc2.setRemoteDescription ' + err.toString());
+        }
+      );
+    },
+    function(err) {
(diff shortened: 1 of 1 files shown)


fippo:
I think that should do it. Maybe checking for actual p2p connectivity is too flaky?

test/test.js
```diff
@@ -1,11 +1,106 @@
...
+
+test('basic connection establishment', function(t) {
+  var pc1 = new m.RTCPeerConnection(null);
+  var pc2 = new m.RTCPeerConnection(null);
+  var ended = false;
+
+  pc1.createDataChannel('somechannel');
+  pc1.oniceconnectionstatechange = function() {
+    if (pc1.iceConnectionState === 'connected' ||
+        pc1.iceConnectionState === 'completed') {
```
jan-ivar, line 23: Can iceConnectionState ever skip 'connected' and go straight to 'completed'? If not then this can be simplified.

test/test.js
```diff
@@ -1,11 +1,106 @@
...
+      if (!ended) {
+        ended = true;
+        t.end();
+      }
+    }
+  };
+
+  var addCandidate = function(pc, event) {
+    if (event.candidate) {
+      var cand = JSON.parse(JSON.stringify(event.candidate));
```
jan-ivar: Seems redundant to me.

test/test.js
```diff
@@ -1,11 +1,106 @@
...
+        },
+        function(err) {
+          t.fail('pc1.setLocalDescription ' + err.toString());
+        }
+      );
+
+      offer = JSON.parse(JSON.stringify(offer));
+      offer = new RTCSessionDescription(offer);
+      t.pass('created RTCSessionDescription from offer');
+      pc2.setRemoteDescription(offer,
```
jan-ivar: This is using the older _"signal from createOffer"_ pattern rather than the newer [signal with pc.localDescription](http://w3c.github.io/webrtc-pc/#simple-peer-to-peer-example) pattern.

test/test.js
fippo, line 23: i've seen a number of bugs here -- most recently https://code.google.com/p/webrtc/issues/detail?id=4483 as well as cases where things don't go to completed.

jan-ivar, line 23: Fair enough.

test/test.js
fippo: an attempt to guard against JSON.stringify issues (in early chrome 43). Not that I ever expected that to happen again, but this is slightly shorter than manually specifying the sdpMid, sdpMLineIndex and candidate.

test/test.js
fippo: pushed, thanks

test/test.js
fippo: as discussed in IRC: also to test that RTCIceCandidate is there in both browsers

test/test.js
```diff
@@ -1,11 +1,103 @@
...
+
+  pc1.createOffer(
+    function(offer) {
+      pc1.setLocalDescription(offer,
+        function() {
+          t.pass('pc1.setLocalDescription');
+
+          offer = new RTCSessionDescription(offer);
+          t.pass('created RTCSessionDescription from offer');
+          pc2.setRemoteDescription(offer,
```
phoglund, line 59: Drive-by (fix in a separate patch since this is submitted): I must quote Linus Torvalds here: "If you need more than 3 levels of indentation, you're screwed anyway, and should fix your program." :)

One suggestion is to move lines 59-91 into a new function that takes pc1, pc2, offer as arguments. It's a pretty natural place to do it since that's where the signaling barrier would be normally. Shorter, shallower functions = good in my book, and it's a bit easier to match up err functions and inspect which calls they match to.

test/test.js
```diff
@@ -1,11 +1,103 @@
...
+                  pc2.setLocalDescription(answer,
+                    function() {
+                      t.pass('pc2.setLocalDescription');
+                      answer = new RTCSessionDescription(answer);
+                      t.pass('created RTCSessionDescription from answer');
+                      pc1.setRemoteDescription(answer,
+                        function() {
+                          t.pass('pc1.setRemoteDescription');
+                        },
+                        function(err) {
```
phoglund, line 74: You can shorten these err functions a bit using a higher order function, like

function errorCallback(msg) {
  return function(err) {
    t.fail(msg + ': ' + err.toString());
  }
}

Then you can replace for instance this one with errorCallback('pc1.setRemoteDescription');

test/test.js
```diff
@@ -1,11 +1,103 @@
...
+    addCandidate(pc2, event);
+  };
+  pc2.onicecandidate = function(event) {
+    addCandidate(pc1, event);
+  };
+
+  pc1.createOffer(
+    function(offer) {
+      pc1.setLocalDescription(offer,
+        function() {
```
phoglund, line 54: Nit: We usually indent 4 on any line breaks, like in the middle of a function invocation like here. Indent 2 should be after {.

test/test.js
jan-ivar, line 59: Having tried both, I actually find the disjoint functions approach harder to follow and more verbose, which might matter as a desirable attribute of adapter.js seems to be to keep things short.

Can we use promises?

alvestrand, line 59: On 05/28/2015 02:42 PM, jan-ivar wrote:

> In test/test.js
> https://github.com/webrtc/adapter/pull/28#discussion_r31228505:
> 
> > -  pc1.onicecandidate = function(event) {
> > -    addCandidate(pc2, event);
> > -  };
> > -  pc2.onicecandidate = function(event) {
> > -    addCandidate(pc1, event);
> > -  };
> >   +
> > -  pc1.createOffer(
> > -    function(offer) {
> > -      pc1.setLocalDescription(offer,
> > -        function() {
> > -          t.pass('pc1.setLocalDescription');
> >   +
> > -          offer = new RTCSessionDescription(offer);
> > -          t.pass('created RTCSessionDescription from offer');
> > -          pc2.setRemoteDescription(offer,
> 
> Having tried both, I actually find the disjoint functions approach
> harder to follow and more verbose, which might matter as a desirable
> attribute of adapter.js seems to be to keep things short.
> 
> Can we use promises?

Not until we've polyfilled them for Chrome.

> —
> Reply to this email directly or view it on GitHub
> https://github.com/webrtc/adapter/pull/28/files#r31228505.

## 

Surveillance is pervasive. Go Dark.

jan-ivar, line 59: No, I meant using:

```
new Promise(function(resolve, reject) {
  pc2.setRemoteDescription(offer, resolve, reject);
})
.then(function() {
  t.pass('pc2.setRemoteDescription');
    pc2.createAnswer(resolve, reject);
  });
})
.then( ... and so forth.
```

But yes, the polyfill will fix this, so deferring cleanup until then makes sense to me.

fippo, line 59: actually I find this one so ugly that I really want promises

jan-ivar:
Nice test but it doesn't output anything. Maybe add some `t.ok`/`t.pass` tests to it?

PR #29, opened 2015-05-26
Modern constraints polyfill in adapter.js (rebase)

jan-ivar:
A rebase of #22 .

PR #30, opened 2015-05-28
Modern constraints polyfill in adapter.js (rebase 2)

jan-ivar:
There must be an easier way. From #29.


jan-ivar:
Sorry, I thought I had to create a new PR to get tests to run, but that doesn't seem to be the problem. closing.

PR #31, opened 2015-05-28
Modern constraints polyfill in adapter.js w/test.

jan-ivar:
Figured out the merge problem. Supersedes #29.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -23,6 +23,28 @@ test('create RTCPeerConnection', function (t) {
   t.ok(typeof(new m.RTCPeerConnection()) === 'object', 'RTCPeerConnection constructor');
 });
 
+test('call getUserMedia with constraints', function (t) {
+  t.plan(3);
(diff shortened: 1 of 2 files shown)


test/test.js
```diff
@@ -23,6 +23,28 @@ test('create RTCPeerConnection', function (t) {
   t.ok(typeof(new m.RTCPeerConnection()) === 'object', 'RTCPeerConnection constructor');
 });
 
+test('call getUserMedia with constraints', function (t) {
+  t.plan(3);
```
alvestrand: You're telling TAP that there will be 3 outputs, but only call t.ok / t.pass twice. This causes failure. Suggest not calling t.plan and calling t.end after the last t.ok - this is supposed to automatically tell tap that <whatever number of t.ok's have passed> is the plan.

jan-ivar:
Thanks, the plan error didn't show up locally for me.

PR #32, opened 2015-05-29
mediaDevices.enumerateDevices() polyfill in adapter.js w/test.

jan-ivar:
Supersedes #23.

### DIFF
--- a/adapter.js
+++ b/adapter.js
@@ -111,7 +111,7 @@ if (navigator.mozGetUserMedia) {
       try {
         return orgEnumerateDevices();
       } catch (e) {
-        if (e.name == "NotFoundError") {
+        if (e.name === 'NotFoundError') {
           return [];
         } else {
           throw e;
--- a/adapter.js
+++ b/adapter.js
@@ -103,7 +103,22 @@ if (navigator.mozGetUserMedia) {
       resolve(infos);
     });
   };
-
+  if (webrtcDetectedVersion < 41) {
+    // Work around http://bugzil.la/1169665
+    var orgEnumerateDevices =
+        navigator.mediaDevices.enumerateDevices.bind(navigator.mediaDevices);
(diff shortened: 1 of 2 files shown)


alvestrand:
Seems to work only on Firefox stable. Testing needed...

jan-ivar:
> not ok 16 At least one device found

The existing bogus polyfill kicks in on Firefox stable, whereas on 39+ it seems to correctly detect that whatever machine this test runs on has neither cameras nor microphones (firefox does not enumerate audio output devices yet, though I suspect it may not have that either).

I see no choice but to remove the failing tests.

alvestrand:
Having a test that accomodates all valid results is OK.
The current failure seems to be a Firefox bug - I can't see a line in the spec that permits enumerateDevices() to return NotFoundError. You may want to consider whether the adapter should hide the bug.

alvestrand:
This fails on Firefox beta (Mozilla Firefox 41.0a1)

https://travis-ci.org/webrtc/adapter/jobs/64608706

Do you care?
(I'm considering upgrading the tip-of-tree tests from "may fail" to "must pass". Letting them fail is asking for trouble, and developers from both teams are involved in this maintenance effort).

jan-ivar:
41 is Nightly, is it not? How often is that pulled?

I think I'll rather submit a fix there. http://bugzil.la/1169665

alvestrand:
Travis pulls the browser distro files every time it runs a try job.
I suspect that Travis keeps a webcache in front of the build machines, since the fetching is rather fast, but I don't think the caches have long expiry times.

If you think this PR is ready now, I'll merge it - but as a general policy, I think I'll merge stuff where the advisory runs fail if the requester explicitly tells me he thinks it's OK.

jan-ivar:
wfm.

alvestrand:
Merged.

PR #34, opened 2015-06-01
wrap webkitRTCPeerConnection

fippo:
this wraps the webkitRTCPeerConnection in a JS object. This _should_ be implementing all important properties and methods. Obviously, any new methods (e.g. ontrack) exposed by the native peerconnection need to be added here before they can become available.

For now, this should allow shimming ontrack and promise-based SLD/SRD. Still need some code to prove that.

### DIFF
--- a/adapter.js
+++ b/adapter.js
@@ -207,6 +207,8 @@ if (navigator.mozGetUserMedia) {
         this.peerconnection);
     this.getRemoteStreams = this.peerconnection.getRemoteStreams.bind(
         this.peerconnection);
+    this.addEventListener = this.peerconnection.addEventListener.bind(
+        this.peerconnection);
   };
 
   var properties = [
--- a/adapter.js
+++ b/adapter.js
@@ -179,8 +179,153 @@ if (navigator.mozGetUserMedia) {
   webrtcMinimumVersion = 38;
 
   // The RTCPeerConnection object.
+  // In Chrome this is a wrapper around the native peerconnection.
   RTCPeerConnection = function(pcConfig, pcConstraints) {
-    return new webkitRTCPeerConnection(pcConfig, pcConstraints);
+    var self = this;
+    this.peerconnection = new webkitRTCPeerConnection(pcConfig, pcConstraints);
+
+    var events = [
+        'onaddstream',
+        'onremovestream',
+        'ondatachannel',
+        'onicecandidate',
+        'onsignalingstatechange',
+        'oniceconnectionstatechange',
+        'onicegatheringstatechange',
+        'onnegotiationneeded'
+    ];
+    events.forEach(function(eventname) {
+      self[eventname] = null;
+      self.peerconnection[eventname] = function(event) {
+        if (self[eventname] !== null) {
+          self[eventname].apply(self, [event]);
+        }
+      };
+    });
+
+    this.getLocalStreams = this.peerconnection.getLocalStreams.bind(
+        this.peerconnection);
+    this.getRemoteStreams = this.peerconnection.getRemoteStreams.bind(
+        this.peerconnection);
+  };
+
+  var properties = [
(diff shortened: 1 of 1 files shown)


adapter.js
```diff
@@ -179,8 +179,153 @@ if (navigator.mozGetUserMedia) {
...
+        'onnegotiationneeded'
+    ];
+    events.forEach(function(eventname) {
+      self[eventname] = null;
+      self.peerconnection[eventname] = function(event) {
+        if (self[eventname] !== null) {
+          self[eventname].apply(self, [event]);
+        }
+      };
+    });
```
alvestrand, line 204: Question: Will this approach allow addEventListener to work?
If not - can you add a comment saying that it won't?

adapter.js
```diff
@@ -179,8 +179,153 @@ if (navigator.mozGetUserMedia) {
...
+  RTCPeerConnection.prototype.close = function() {
+    this.peerconnection.close();
+  };
+
+  RTCPeerConnection.prototype.createDataChannel = function(label, opts) {
+    return this.peerconnection.createDataChannel(label, opts);
+  };
+
+  RTCPeerConnection.prototype.setLocalDescription = function(description,
+      successCallback, failureCallback) {
```
alvestrand, line 246: For my information: Is it possible to make this function check how many arguments it has been called with, or is there some other way to make it polymorphic?
EDIT: Found http://www.markhansen.co.nz/javascript-optional-parameters/, which explains some options quite nicely. Guess that's your next PR :-)

alvestrand:
This should make some changes to the external behavior - the most obvious one should be that the class of the returned object is RTCPeerConnection, not WebkitRTCPeerConnection (I think).

Can you add a test verifying that this change happened?
(I _always_ want a test that failed before the change was made...)

EDIT: I was wrong. "new webkitRTCPeeerConnection(null)" in Chrome returns an object that calls itself an RTCPeerConnection. Will the scribbling over the RTCPeerConnection.prototype fields have any interesting effects because of this?

adapter.js
fippo, line 204: good catch, added in 45a75fe37eb5578696fc8773a3b826cb788dec12

adapter.js
```diff
@@ -179,8 +179,155 @@ if (navigator.mozGetUserMedia) {
...
+        'onaddstream',
+        'onremovestream',
+        'ondatachannel',
+        'onicecandidate',
+        'onsignalingstatechange',
+        'oniceconnectionstatechange',
+        'onicegatheringstatechange',
+        'onnegotiationneeded'
+    ];
+    events.forEach(function(eventname) {
```
jan-ivar, line 197: Nit: technically not eventname. Maybe oneventname?

adapter.js
```diff
@@ -179,8 +179,155 @@ if (navigator.mozGetUserMedia) {
...
+      self[eventname] = null;
+      self.peerconnection[eventname] = function(event) {
+        if (self[eventname] !== null) {
+          self[eventname].apply(self, [event]);
```
jan-ivar, line 201: If self[eventname] is undefined, this will try to call undefined as a function. Is that intentional?

fippo:
@alvestrand hah, "scribbling over" those fields (i.e.overriding webkitRTCPeerConnection.prototype.createOffer et al) will be much simpler, thanks! New PR incoming...

alvestrand:
Heh - the RTCPeerConnection prototype isn't exposed to Javascript, so it has to be done more circuitously.....

The thing I was worried about was that the way you wrote it, I was worried that the first time, it would go "wrap underlying RTCPeerConnection.xxx in my own RTCPeerConnection.xxx", and if called a second time, it would do the same, leading to an extremely long prototype chain of RTCPeerConnection.xxx ... is that a worry, or is it something we don't worry about?

PR #35, opened 2015-06-01
Add description of how to test with W3C tests.

alvestrand:


alvestrand:
@fippo @jan-ivar is this information that is worth adding?

jan-ivar:
I think so. I'll definitely have to try that. ;) Pity they can't run automatically. Thanks for the info!

alvestrand:
They can actually be run automatically - Chrome has a script that does exactly that. But I think that one depends on much Chrome scaffolding.

PR #36, opened 2015-06-02
promises support for chrome

fippo:
adds promise-support, with a much simpler approach than issue #34 would have resulted in. Still needs a little cleanup before merge, the wrapper functions can be generated I think.

### DIFF
--- a/adapter.js
+++ b/adapter.js
@@ -259,6 +259,27 @@ if (navigator.mozGetUserMedia) {
           );
       });
     };
+    var addIceCandidate = webkitRTCPeerConnection.prototype.addIceCandidate;
+    webkitRTCPeerConnection.prototype.addIceCandidate = function() {
+      var args = arguments;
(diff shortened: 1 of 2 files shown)


fippo:
Looks like a travis hickup... fails to install node?!

@phoglund did I get the indent right this time?
It also seems grunt doesn't lint test.js, will fix in another PR.

alvestrand:
Hooray!

Merging, with joy. I'll see if the w3c tests pass when using adapter after merging.

fippo:
woho. I did wonder if there needs to be a .bind of the method overridden like in https://github.com/webrtc/samples/pull/484/files#diff-0d5896d8ac938a743878fa854e2b7162R185
but the current behaviour of

```
var pc = new webkitRTCPeerConnection(null);
pc.createOffer(function (offer) { console.log(this); })
```

seems to be that this === window, not this === pc

PR #38, opened 2015-06-03
Publish this package to npmjs with a test package name.

alvestrand:
Another package is using the name "webrtc-adapter" on npm. In preparation
for making this the official name, we publish this version under a
different name, so that users can test before replacing the other.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,11 +1,11 @@
 {
-  "name": "webrtc-adapter",
-  "version": "0.1.0",
-  "description": "Hide browser differences in WebRTC APIs",
+  "name": "webrtc-adapter-test",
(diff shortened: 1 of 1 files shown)


alvestrand:
@bemasc @willscott Please comment on whether this is an appropriate step in resolving the name issue.

willscott:
Seems reasonable.

JacopoDaeli:
Seems good. Which is the npm module name at the moment? 

Should I use npm-test-publish or webrtc-adapter ?

alvestrand:
webrtc-adapter-test.
npm-test-publish was the branch name, not the package name.

PR #39, opened 2015-06-03
also run jshint and jscs on tests

fippo:
jshint and jscs were not run on test/test.js

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -160,12 +160,9 @@ test('basic connection establishment with promise', function(t) {
   var addCandidate = function(pc, event) {
     if (event.candidate) {
       var cand = new RTCIceCandidate(event.candidate);
-      pc.addIceCandidate(cand)
-          .then(function() {
-          })
-          .catch(function(err) {
-            t.fail('addIceCandidate ' + err.toString());
-          });
+      pc.addIceCandidate(cand).catch(function(err) {
+        t.fail('addIceCandidate ' + err.toString());
+      });
     }
   };
   pc1.onicecandidate = function(event) {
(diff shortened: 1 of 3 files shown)


test/test.js
```diff
@@ -159,9 +161,9 @@ test('basic connection establishment with promise', function(t) {
     if (event.candidate) {
       var cand = new RTCIceCandidate(event.candidate);
       pc.addIceCandidate(cand)
-          .then(function () {
+          .then(function() {
           })
-          .catch(function (err) {
+          .catch(function(err) {
```
alvestrand: Should we match these on style with the case below, and wrap .then and .catch up to the previous line?

test/test.js
```diff
@@ -173,39 +175,39 @@ test('basic connection establishment with promise', function(t) {
...
     devices.forEach(function(d) {
-      t.ok(d.kind == 'videoinput' ||
-           d.kind == 'audioinput' ||
-           d.kind == 'audiooutput', 'Known device kind');
+      t.ok(d.kind === 'videoinput' ||
+           d.kind === 'audioinput' ||
+           d.kind === 'audiooutput', 'Known device kind');
       t.ok(d.deviceId.length !== undefined, 'device id present');
       t.ok(d.label.length !== undefined, 'device label present');
     });
```
alvestrand, line 210: Didn't catch this when I first looked at it, but aren't we missing a ".catch(function (err) { t.fail(err.toString()); }) here?

test/test.js
fippo: removed the empty success callback (consistent with what the spec example shows) and moved up the catch to the previous line.

test/test.js
jan-ivar, line 210: No, it's there. The code just gets cut off here. ;) (click the "unfold" symbol to see it)

Issue #42, opened 2015-06-08
Capture screen in FF37-OSX not working: constraintsToFF37 mess up the constraints


igracia:
I've run into an issue trying to capture the screen in FF37. When I invoke getUserMedia, no screen or window is shared, no dialogue is shown, and this appears in the logs

```
spec: {"audio":false,"video":{"mandatory":{"maxWidth":640,"maxFrameRate":15,"minFrameRate":15},"mozMediaSource":"screen","mediaSource":"screen"}}
ff37: {"audio":false,"video":{"mandatory":{"maxWidth":640,"maxFrameRate":15,"minFrameRate":15},"advanced":[{"mozMediaSource":{"min":"screen","max":"screen"}},{"mediaSource":{"min":"screen","max":"screen"}}]}}
TypeError: 'mediaSource' member of MediaTrackConstraintSet '[object Object]' is not a valid value for enumeration MediaSourceEnum.
```

It looks like the mangling done by the function _constraintsToFF37_ is not working/needed for FF37, as commenting out the call to the function allows me to share the screen again. I've finally changed the code from

```
  getUserMedia = (webrtcDetectedVersion < 38)
```

to 

```
  getUserMedia = (webrtcDetectedVersion < 37) 
```

and so far it's working fine. Could you please confirm this issue?

Thanks,

jan-ivar:
Confirmed. Thanks for reporting this! I'm working on a fix now. facingMode is likely affected as well.

A side-note on your example: The constraints you're passing in for everything but mediaSource are not to spec and [don't work on Firefox](http://stackoverflow.com/questions/28282385/webrtc-firefox-constraints/28911694#28911694), even with the adapter. That's probably why you don't notice a difference with your workaround, which effectively disables adapter's spec-translation for 37.

jan-ivar:
2nd side-note: it's just `mediaSource`. No implementation ever used `mozMediaSource` fwiw (I checked).

igracia:
Thanks for confirming this, and for the prompt reply! Also thanks for the tips about the constraints. I'll check them against the specs to have them according to that. Thanks also for spotting the `mozMediaSource` issue, I'll fix that right away.

KaptenJansson:
@igracia does this work for you now?

igracia:
Yes, we have some automated tests running in Saucelabs and it's working.

[...]

PR #43, opened 2015-06-08
Unbreak FF37 handling of mediaSource and facingMode (regression)

jan-ivar:
`mediaSource` (screensharing) and `facingMode` should work again. Issue #42

### DIFF
--- a/adapter.js
+++ b/adapter.js
@@ -93,19 +93,31 @@ if (navigator.mozGetUserMedia) {
       }
       var require = [];
       Object.keys(c).forEach(function(key) {
+        if (key === 'require' || key === 'advanced' || key === 'mediaSource') {
+          return;
+        }
         var r = c[key] = (typeof c[key] === 'object') ?
             c[key] : {ideal: c[key]};
-        if (r.exact !== undefined) {
-          r.min = r.max = r.exact;
-          delete r.exact;
-        }
-        if (r.min !== undefined || r.max !== undefined) {
+        if (r.min !== undefined ||
+            r.max !== undefined || r.exact !== undefined) {
           require.push(key);
+        }
+        if (r.exact !== undefined) {
+          if (typeof r.exact === 'number') {
+            r.min = r.max = r.exact;
+          } else {
+            c[key] = r.exact;
+          }
+          delete r.exact;
         }
         if (r.ideal !== undefined) {
           c.advanced = c.advanced || [];
           var oc = {};
-          oc[key] = {min: r.ideal, max: r.ideal};
+          if (typeof r.ideal === 'number') {
+            oc[key] = {min: r.ideal, max: r.ideal};
+          } else {
+            oc[key] = r.ideal;
+          }
           c.advanced.push(oc);
           delete r.ideal;
           if (!Object.keys(r).length) {
@@ -231,7 +243,8 @@ if (navigator.mozGetUserMedia) {
       }
       var cc = {};
       Object.keys(c).forEach(function(key) {
-        if (key === 'require' || key === 'advanced') {
+        if (key === 'require' || key === 'advanced' || key === 'mediaSource' ||
+            key === mozMediaSource) {
           return;
         }
         var r = (typeof c[key] === 'object') ? c[key] : {ideal: c[key]};
--- a/adapter.js
+++ b/adapter.js
@@ -243,8 +243,7 @@ if (navigator.mozGetUserMedia) {
       }
       var cc = {};
       Object.keys(c).forEach(function(key) {
-        if (key === 'require' || key === 'advanced' || key === 'mediaSource' ||
-            key === mozMediaSource) {
+        if (key === 'require' || key === 'advanced' || key === 'mediaSource') {
           return;
         }
         var r = (typeof c[key] === 'object') ? c[key] : {ideal: c[key]};


jan-ivar:
I've manually verified the following on FF37 (and that it worked):

```
"spec: {"video":{"mediaSource":"screen","facingMode":"user"}}"
"ff37: {"video":{"mediaSource":"screen","advanced":[{"facingMode":"user"}]}}"

"spec: {"video":{"mediaSource":"screen","facingMode":{"exact":"user"}}}"
"ff37: {"video":{"mediaSource":"screen","facingMode":"user","require":["facingMode"]}}"
```

adapter.js
```diff
@@ -93,19 +93,31 @@ if (navigator.mozGetUserMedia) {
       }
       var require = [];
       Object.keys(c).forEach(function(key) {
+        if (key === 'require' || key === 'advanced' || key === 'mediaSource') {
+          return;
+        }
         var r = c[key] = (typeof c[key] === 'object') ?
             c[key] : {ideal: c[key]};
+        if (r.min !== undefined ||
```
alvestrand, line 101: Please write a test that exercises this code (and verifies that the right result is produced).

adapter.js
```diff
@@ -231,7 +243,7 @@ if (navigator.mozGetUserMedia) {
       }
       var cc = {};
       Object.keys(c).forEach(function(key) {
-        if (key === 'require' || key === 'advanced') {
+        if (key === 'require' || key === 'advanced' || key === 'mediaSource') {
```
alvestrand, line 246: Here too you're introducing special handling. Test it.

adapter.js
```diff
@@ -93,19 +93,31 @@ if (navigator.mozGetUserMedia) {
       }
       var require = [];
       Object.keys(c).forEach(function(key) {
+        if (key === 'require' || key === 'advanced' || key === 'mediaSource') {
+          return;
```
alvestrand, line 97: This line bothers me, because it means that when we hit it, we continue with an empty set of constraints, no matter what the argument is. That seems like a singularly bad idea.

Test.

alvestrand:
The oldest version currently being run by travis-multirunner is Firefox 38. Is there a possibility of getting Firefox versions older than 38 to test with? If so - could you work with travis-multirunner to get that included into the test suite somehow?
(I've looked for public repos of Chrome versions older than stable. They don't appear to exist.)

adapter.js
jan-ivar, line 97: No, this doesn't remove a constraint, rather it skips any translation of it, leaving it as it was. (`c` is what's ultimately returned in the firefox shim) This is intentional to not inadvertently remove any app requirements. This works in Firefox which ignores unknown dictionary keys.

Will add tests.

jan-ivar:
Earlier versions of firefox are [here](https://ftp.mozilla.org/pub/mozilla.org/firefox/releases). I know very little about how travis-multirunner downloads its stuff. @fippo do you know?

adapter.js
KaptenJansson, line 97: Maybe we should name it "constraints" rather than "c" to make it easier to read. Yes it might result in line breaks here and there but I think it's worth it.

alvestrand, line 97: Actually the symbol "c" is overloaded in the function; it is used in the getusermedia function's argument list and it is used in the constraintsToFF37 argument list. This is not a good overload.

The thing is - constraintToFF37 is called like this:

c.audio = constraintsToFF37(c.audio)

What value will c.audio have after this line when we hit the "return;" statement?

alvestrand:
travis-multirunner is here:

https://github.com/DamonOehlman/travis-multirunner

It uses DamonOehlman/djo-shell's browser-version.sh to detect the browser version. 

I suppose he would accept a patch that allows you to set BVER to a number instead of "stable".

adapter.js
jan-ivar, line 97: I see no bug there. The only `return;` statement is in the function passed to `forEach()`, which amounts to a "continue".

There are two functions with the argument `c`, one defined inside the other. The outer one does not have access to the inner one, and the inner one does not have access to the outer one. That seems nice and safe to me.

I tried it other ways, but found it too easy to use the wrong one. This way it's not possible to use the wrong one.

There are 34 hits for `c.` in this file. I tried it with `constraints` and `constraintSet` initially, but the verbosity was overpowering. I renamed it to `c` to not have the names get in the way of the code. YMMV.

In any case, this sort of cleanup seems out of scope for this regression PR.

alvestrand, line 97: And the reason I didn't see it was that I didn't see the third function level. Sigh.

I reject your argument about reusing the name "c". It makes it impossible to use the wrong one, but makes it entirely too easy to _read_ the wrong one.
Readability is the more important property. Naming the outer c "stream_constraints" would make this part readable for me.

alvestrand:
If you can write a test that exercises this, I'm approving it.

Legal test methods include overwriting webrtcDetectedVersion and rebinding mozGetUserMedia within the scope of the test.

KaptenJansson:
@jan-ivar ping about the https://github.com/webrtc/adapter/pull/43#issuecomment-111064451.

KaptenJansson:
@jan-ivar do you have cycles to add some tests for this as requested by @alvestrand ?

jan-ivar:
@KaptenJansson - yes, sorry, last I tried I had trouble rebinding navigator.mozGetUserMedia, but I'll try @fippo 's trick in https://github.com/webrtc/adapter/pull/74

KaptenJansson:
No worries, just trying to make sure we do NOT let good PR's get stale and die ;).

PR #44, opened 2015-06-08
add attachMediaStream / reattachMediaStream for edge

fippo:
Edge supports getUserMedia as of last month: http://blogs.windows.com/msedgedev/2015/05/13/announcing-media-capture-functionality-in-microsoft-edge/

GUM works just fine, but in order to make the samples work, we need to shim attachMediaStream.

### DIFF
--- a/adapter.js
+++ b/adapter.js
@@ -290,6 +290,7 @@ if (navigator.mozGetUserMedia) {
   }
 } else if (navigator.mediaDevices && navigator.userAgent.match(
     /Edge\/(\d+).(\d+)$/)) {
+  console.log('This appears to be Edge');
   webrtcDetectedBrowser = 'edge';
 
   webrtcDetectedVersion =
--- a/adapter.js
+++ b/adapter.js
@@ -154,14 +154,8 @@ if (navigator.mozGetUserMedia) {
       });
     };
   }
-  // Attach a media stream to an element.
-  attachMediaStream = function(element, stream) {
-    console.log('Attaching media stream');
-    element.mozSrcObject = stream;
-  };
 
   reattachMediaStream = function(to, from) {
-    console.log('Reattaching media stream');
     to.mozSrcObject = from.mozSrcObject;
   };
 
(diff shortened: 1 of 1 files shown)


fippo:
drive-by fix: moving attachMediaStream outside the browser switch. The chrome variant had implemented its own shim logic ever since.

adapter.js
```diff
@@ -316,10 +288,45 @@ if (navigator.mozGetUserMedia) {
       });
     }};
   }
+} else if (navigator.mediaDevices && navigator.userAgent.match(
+    /Edge\/(\d+).(\d+)$/)) {
+  webrtcDetectedBrowser = 'edge';
+
```
KaptenJansson, line 321: Add console.log('This appears to be MS Edge') as we do for other browsers.

KaptenJansson:
This LGTM, WDYT @alvestrand @juberti ?

adapter.js
```diff
@@ -316,10 +288,46 @@ if (navigator.mozGetUserMedia) {
       });
     }};
   }
+} else if (navigator.mediaDevices && navigator.userAgent.match(
+    /Edge\/(\d+).(\d+)$/)) {
+  console.log('This appears to be Edge');
+  webrtcDetectedBrowser = 'edge';
```
alvestrand, line 320: Any thoughts about how we can do testing with Edge?

adapter.js
```diff
@@ -316,10 +288,46 @@ if (navigator.mozGetUserMedia) {
...
+    parseInt(navigator.userAgent.match(/Edge\/(\d+).(\d+)$/)[2], 10);
+
+  // the minimum version still supported by adapter.
+  webrtcMinimumVersion = 12;
 } else {
   console.log('Browser does not appear to be WebRTC-capable');
 }
 
+// Attach a media stream to an element.
+attachMediaStream = function(element, stream) {
```
alvestrand: This smells wrong to me. This is a shim, we should only be shimming things if we know shimming is needed. Moving these functions out here makes us re-detect which browser we're facing on every use.

If Edge needs these functions, define them for Edge, in the Edge-specific way.

fippo: one of the problems of adapter is that it mixes shimming (navigator.getUserMedia = navigator.webkitGetUserMedia || ...) with utility functions like attachMediaStream. That's not a bad thing, just violates the "one thing" rule for small modules.

I can revert to 3374f5d63f82268dffd26aad15ed3d5f8f0534d1 -- and remove the mozSrcObject part of Chrome's attachMediaStream for clarity

alvestrand: Seems good to me. My next-scheduled-experiment is to see if I can use a Javascript setter to polyfill srcObject without the need for attachMediaStream.

At some major version change I'd like to drop the utility functions.

PR #45, opened 2015-06-09
shim addEventlistener with a no-op

fippo:
@jan-ivar -- that should take care of code that adds an eventlistener for 'devicechange' (in particular I found this in https://github.com/MicrosoftEdge/Demos/blob/master/photocapture/scripts/demo.js#L175)

### DIFF
--- a/adapter.js
+++ b/adapter.js
@@ -153,6 +153,7 @@ if (navigator.mozGetUserMedia) {
         throw e;
       });
     };
+    navigator.mediaDevices.addEventListener = function () { };
   }
   // Attach a media stream to an element.
   attachMediaStream = function(element, stream) {
@@ -315,6 +316,7 @@ if (navigator.mozGetUserMedia) {
         });
       });
     }};
+    navigator.mediaDevices.addEventListener = function () { };
   }
 } else {
   console.log('Browser does not appear to be WebRTC-capable');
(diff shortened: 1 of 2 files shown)


adapter.js
```diff
@@ -315,6 +319,9 @@ if (navigator.mozGetUserMedia) {
         });
       });
     }};
+    // in case someone wants to listen for the devicechange event.
+    navigator.mediaDevices.addEventListener = function() { };
```
alvestrand, line 323: Since this changes the API of adapter.js to not crash when addEventListener is called - can you add a test for it?

Of course that means having to add the shim to Chrome too. I don't want the functionality of the two to diverge.

fippo, line 323: test added. shim is in both chrome and for old versions of firefox. Newer versions have this (without functionality) automagically.

test/test.js
```diff
@@ -216,3 +216,13 @@ test('call enumerateDevices', function(t) {
     t.end();
   });
 });
+
+test('navigator.mediaDevices eventlisteners', function(t) {
+  navigator.mediaDevices.addEventListener('devicechanged', function() {
```
alvestrand: Would it be clearer to show what you are testing by using "t.doesNotThrow(function() {navigator.mediaDevices.addEventListener('devicechanged')})?

fippo: How about testing that addEventListener is a function instead?
And adding some documentation that the primary thing this is used for is the devicechanged event?

PR #46, opened 2015-06-09
Polyfill for Chrome getStats + test

milkypostman:
This is a migration of webrtc/samples#484.

Shim for https://code.google.com/p/webrtc/issues/detail?id=4117

//cc @KaptenJansson

### DIFF
--- a/adapter.js
+++ b/adapter.js
@@ -183,8 +183,41 @@ if (navigator.mozGetUserMedia) {
 
   // The RTCPeerConnection object.
   window.RTCPeerConnection = function(pcConfig, pcConstraints) {
-    return new webkitRTCPeerConnection(pcConfig, pcConstraints);
+    var pc = new webkitRTCPeerConnection(pcConfig, pcConstraints);
+    var origGetStats = pc.getStats.bind(pc);
+    pc.getStats = function(selector, successCallback, errorCallback) { // jshint ignore: line
+      // If selector is a function then we are in the old style stats so just
+      // pass back the original getStats format to avoid breaking old users.
+      if (typeof selector === 'function') {
+        return origGetStats(selector, successCallback);
+      }
+
+      var fixChromeStats = function(response) {
+        var standardReport = {};
+        var reports = response.result();
+        reports.forEach(function(report) {
+          var standardStats = {
+            id: report.id,
+            timestamp: report.timestamp,
+            type: report.type
+          };
+          report.names().forEach(function(name) {
+            standardStats[name] = report.stat(name);
+          });
+          standardReport[standardStats.id] = standardStats;
+        });
+
+        return standardReport;
+      };
+      var successCallbackWrapper = function(response) {
+        successCallback(fixChromeStats(response));
+      };
+      return origGetStats(successCallbackWrapper, selector);
+    };
+
+    return pc;
   };
+
   // add promise support
   ['createOffer', 'createAnswer'].forEach(function(method) {
     var nativeMethod = webkitRTCPeerConnection.prototype[method];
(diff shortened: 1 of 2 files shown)


adapter.js
```diff
@@ -179,8 +179,43 @@ if (navigator.mozGetUserMedia) {
...
+      var selector = arg1;
+
+      var fixChromeStats = function(response) {
+        var standardReport = {};
+        var reports = response.result();
+        reports.forEach(function(report) {
+          var standardStats = {
+              id: report.id,
+              timestamp: report.timestamp,
+              type: report.type
```
KaptenJansson: indent these by 2 instead of 4 and grunt will stop complaining.

milkypostman: thanks. my problem is that it doesn't print all the errors the first time. so i kept thinking i fixed them all and not re-running grunt. :facepunch:

KaptenJansson: Yeah that is a quirk of grunt (I think it's actually configurable on how many errors it will let by before stopping), try the --force flag next time.

KaptenJansson:
@alvestrand and @juberti WDYT?

KaptenJansson:
LGTM from me.

test/test.js
```diff
@@ -216,3 +216,18 @@ test('call enumerateDevices', function(t) {
...
+    pc1.getStats(null, resolve, reject);
+  })
+  .then(function(stats) {
+    t.ok(typeof(stats) === 'object', 'Stats object.');
+    t.end();
+  })
+  .catch(function(err) {
+    t.fail('getStats() should never fail with error: ' + err.toString());
+    t.end();
+  });
```
alvestrand, line 253: This tests that the happy path executes, but gives no indication which branch was taken (with or without fixChromeStats).
Can you make a test that actually checks that the stats are in the form that fixChromeStats produces?
Also - a test that verifies that the old format comes back in the old style call is a Good Thing.

I know I'm anal about testing....

adapter.js
```diff
@@ -179,8 +179,43 @@ if (navigator.mozGetUserMedia) {
 
   // The RTCPeerConnection object.
   window.RTCPeerConnection = function(pcConfig, pcConstraints) {
-    return new webkitRTCPeerConnection(pcConfig, pcConstraints);
+    var pc = new webkitRTCPeerConnection(pcConfig, pcConstraints);
+    var origGetStats = pc.getStats.bind(pc);
+    pc.getStats = function(arg1, arg2) {
```
alvestrand: This is not according to spec. The spec (at the moment) has selector, successCallback, errorCallback, in that order - 3 arguments, not 2.
Please use those names, not arg1 / arg2. Writing "if typeof selector === 'function' looks a bit weird, but it's more readable anyway.

test/test.js
milkypostman, line 253: Is there a way to have part of a test only run for chrome? because otherwise the test will fail in ff, which always does the "right" thing.

alvestrand, line 253: You can always write "if (webrtcDetectedBrowser === 'chrome')" around those tests that should run in Chrome only. The way we're using the test harness, we don't care how many test statements are actually run (that would matter if we called t.plan())

alvestrand:
Unfortunately jshint fails on this (unused variable).
https://travis-ci.org/webrtc/adapter/jobs/66500961

Can you fix?

milkypostman:
Yeah, sorry for the noise, I was actually still working out some things with Firefox but wanted to get the changes saved.

PTAL. I think I have fixed the other changes.

alvestrand:
Looks good. You need a rebase - it says "can't automatically merge this request". guess there are too many other changes....

milkypostman:
rebased. should I commit myself or do you want to do it after everything passes?

alvestrand:
I merged it. Going to increment version now.

PR #53, opened 2015-06-15
Add RequireJS support.

byronclark:


KaptenJansson:
Run "npm install" and then "grunt "locally to catch these types of errors.

adapter.js
```diff
@@ -361,4 +361,18 @@ if (typeof module !== 'undefined') {
     //requestUserMedia: not exposed on purpose.
     //trace: not exposed on purpose.
   };
+} else if ((typeof require === 'function') && (typeof define === 'function')) {
```
KaptenJansson, line 364: WDYT about adding a comment that this is for the RequireJS file and module loader?

PR #55, opened 2015-06-15
Npm test experiment

alvestrand:
Using a wrapper script for "npm test" so that it does something useful even without setting
environment variables.


KaptenJansson:
LGTM!

PR #64, opened 2015-06-26
Use RTCPeerConnection defined on the window 

xdumaine:
... to prevent reference error in a browser where it's not defined globally. In Safari, this was throwing a `ReferenceError: Can't find variable: RTCPeerConnection` and also happened to be a linter error when I opened up the file.


xdumaine:
I signed it!

fippo:
#63 beat you. Looks like we need a safari test :-(

xdumaine:
Fixed in #63

PR #67, opened 2015-07-02
Prevent ReferenceError in node context by checking for window

xdumaine:
Check for window existence to allow requiring the module without error in a node context

### DIFF
--- a/adapter.js
+++ b/adapter.js
@@ -38,6 +38,7 @@ function trace(text) {
 
 if (typeof window === 'undefined' || !window.navigator) {
   console.log('This does not appear to be a browser');
+  webrtcDetectedBrowser = 'not a browser';
 } else if (navigator.mozGetUserMedia) {
   console.log('This appears to be Firefox');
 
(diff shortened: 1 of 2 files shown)


xdumaine:
Any objections to this @alvestrand @KaptenJansson?

adapter.js
```diff
@@ -36,7 +36,9 @@ function trace(text) {
   }
 }
 
-if (navigator.mozGetUserMedia) {
+if (typeof window === 'undefined' || !window.navigator) {
+  console.log('This does not appear to be a browser');
+} else if (navigator.mozGetUserMedia) {
```
alvestrand, line 42: Can you set webrtcDetectedBrowser to 'not a browser' here?

alvestrand:
Good catch!

Because you can probably do this a lot more easily than I can: Can you add a test that verifies that the adapter _can_ be loaded in a node.js context?

This could be an extra task in npm's "test" label, an extra line in "test/run-tests" or an extra task in Gruntfile.js (you're caught in a twisty maze of packagers, all different).... just pick one.

This should have been caught earlier.... let's make sure it doesn't happen again.

alvestrand:
pinging @fippo

xdumaine:
Good call. I added a verify task to grunt that warns and stops if that fails.

alvestrand:
Thank you!

PR #70, opened 2015-07-09
add build handling for npm publishing

### LINKED ISSUE #66: Need publishing procedure for NPM packages
Discovered today "the hard way" that "npm publish" grabs everything in the homedir by default and includes it as part of the package, whether it's junk or not. This needs to not happen - one can use "files" in package.json and/or an ".npmignore" file, or one can use a procedure / script that does a clean checkout before the "npm publish".

KaptenJansson:
Fixes #66

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 node_modules/
+out/
 *~
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -22,6 +22,34 @@ module.exports = function(grunt) {
     },
     testling: {
       files: 'test/test.js'
+    },
+    copy: {
+      build: {
+        cwd: '.',
+        files: [
+          {src: [
+            // Make sure to add files that should be included for the NPM
+            // package here.
+            'adapter.js',
+            'bower.json',
+            'package.json',
+            'CONTRIBUTING.md',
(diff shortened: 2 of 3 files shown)


alvestrand:
What does this achieve, exactly?
I think it creates a new grunt command "grunt build", which will call "grunt clean" and "grunt copy" in that order, but why?

This seems to need an update to README to describe how one uses the functionality (and why).

KaptenJansson:
> What does this achieve, exactly?
> I think it creates a new grunt command "grunt build", which will call "grunt clean" and "grunt copy" in that order, but why?

It cleans out the out/ directory if it exists to make sure we start fresh, it then copies the files listed in the files array in Gruntfile.js to out/. This is done to ensure we explicitly choose which files should be part of the NPM published package, hence it fixes what you saw in #66.

I can change the name of the build command if it makes it more clear?

Do you want to add the npm publish step to this as well or keep it separate (i.e. grunt build > check that files are OK > npm Publishj)?

> This seems to need an update to README to describe how one uses the functionality (and why).

Sure I can add steps on how to prepare and publish this as an NPM package.

KaptenJansson:
I guess all of this can be handled by just populating .npmignore properly. Closing....

PR #72, opened 2015-07-12
test attachmediastream with travis-multirunner 2.5

fippo:
@alvestrand moar tests :-)

@jan-ivar I tried getting different resolutions from FF, but fake: true only returned 640x480. Shall I file a bug for fake with different resolutions?

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -35,7 +35,7 @@ test('attachMediaStream', function(t) {
     t.pass('got stream with w=' + video.videoWidth +
            ',h=' + video.videoHeight);
     t.end();
-  }
+  };
 
   var constraints = {video: true, fake: true};
   navigator.mediaDevices.getUserMedia(constraints)
--- a/test/test.js
+++ b/test/test.js
@@ -30,7 +30,7 @@ test('create RTCPeerConnection', function(t) {
 test('attachMediaStream', function(t) {
   var video = document.createElement('video');
   // if attachMediaStream works, we should get a video
-  // at some point. This will trigger onloadedmetadata
+  // at some point. This will trigger onloadedmetadata.
   video.onloadedmetadata = function() {
     t.pass('got stream with w=' + video.videoWidth +
            ',h=' + video.videoHeight);
--- a/test/test.js
+++ b/test/test.js
@@ -27,6 +27,29 @@ test('create RTCPeerConnection', function(t) {
       'RTCPeerConnection constructor');
 });
 
+test('attachMediaStream', function (t) {
+  var video = document.createElement('video');
+  var resized = false;
+  video.onresize = function() {
+    if (!resized) {
+      // in chrome, this is called twice?
+      t.pass('got stream with w=' + video.videoWidth + ',h=' + video.videoHeight);
+      t.end();
+      resized = true;
+    }
+  };
+  var constraints = {video: true};
+  if (m.webrtcDetectedBrowser === 'firefox') constraints.fake = true;
+  navigator.mediaDevices.getUserMedia(constraints)
+  .then(function(stream) {
+    t.pass('got stream.');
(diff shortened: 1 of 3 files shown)


jan-ivar:
@fippo please do. I've been meaning to add those.

test/test.js
```diff
@@ -27,6 +27,32 @@ test('create RTCPeerConnection', function(t) {
...
+    if (!resized) {
+      // in chrome, this is called twice?
+      t.pass('got stream with w=' + video.videoWidth +
+             ',h=' + video.videoHeight);
+      t.end();
+      resized = true;
+    }
+  };
+  var constraints = {video: true};
+  if (m.webrtcDetectedBrowser === 'firefox') {
```
alvestrand: constraints.fake = true should set "ideal" to true. So this should do no harm to Chrome. If it does do harm, it's a problem with the constraints rewriting code. Can you see if tests pass if you remove the browser test?

(I dislike having browser-specific code in tests when we're not testing browser-specific functionality.)

test/test.js
```diff
@@ -27,6 +27,32 @@ test('create RTCPeerConnection', function(t) {
...
+    constraints.fake = true;
+  }
+  navigator.mediaDevices.getUserMedia(constraints)
+  .then(function(stream) {
+    t.pass('got stream.');
+    m.attachMediaStream(video, stream);
```
alvestrand, line 44: What do you actually test here? That m.attachMediaStream doesn't throw an exception?

Please add a t.pass('attachMediaStream worked') after the call to attachMediaStream so that we're making this explicit.

test/test.js
fippo:
```
spec:   {"video":true,"fake":true}
chrome: {"video":true,"fake":true}
```

so it looks like it doesn't get transformed. Tests still pass, so i'm just going to set fake: true.

test/test.js
fippo, line 44: the actual test is whether I get a video stream -- added a comment and rewrote it to use onloadedmetadata which is called only once in both FF and Chrome.

test/test.js
```diff
@@ -27,6 +27,28 @@ test('create RTCPeerConnection', function(t) {
       'RTCPeerConnection constructor');
 });
 
+test('attachMediaStream', function(t) {
+  var video = document.createElement('video');
```
jan-ivar, line 31: Maybe add a `t.plan(3);` here? Otherwise nothing would catch e.g. loadedmetadata never firing.

jan-ivar, line 31: except `t.end()` not running would stall the test I suppose. hmm. I'll leave it to you.

fippo, line 31: I don't think we have timeouts on any tests. Might be a good idea though.

alvestrand, line 31: There's an overall timeout of 10 seconds or so on travis-ci in general.
So hanging tests get caught eventually, and stalling on error is not a big issue (as long as we fix errors on the main branch ASAP).

fippo, line 31: it's 10 minutes. But yeah, as long as we keep fixing those issues that shouldn't be a problem.

test/test.js
```diff
@@ -27,6 +27,28 @@ test('create RTCPeerConnection', function(t) {
...
+  // if attachMediaStream works, we should get a video
+  // at some point. This will trigger onloadedmetadata
```
KaptenJansson: Add full stop "." after comment.

KaptenJansson:
So we are fine with if loadedmetadata is never fired this test will hang for 10 min (travis timeout)?

alvestrand:
I think we're fine. Just add the missing full stop (Google style guide says "be a stickler for grammar") and let's ship it.

KaptenJansson:
LGTM after nit has been fixed.

PR #74, opened 2015-07-13
wrap console.log; dont log when used as a module

### LINKED ISSUE #68: Remove console.logs
Please remove console calls

fippo:
removes console.log when used in a module, fixes #68.
Also removes some inconsistent logging and debug in the resolution stuff.

Ideally we would provide modules with a way to set a logger but... maybe later.

### DIFF
--- a/adapter.js
+++ b/adapter.js
@@ -22,15 +22,16 @@ var reattachMediaStream = null;
 var webrtcDetectedBrowser = null;
 var webrtcDetectedVersion = null;
 var webrtcMinimumVersion = null;
-
-function webrtcLogger() {
-  // suppress console.log output when being included as a module
-  if (typeof module !== 'undefined' ||
-      ((typeof require === 'function') && (typeof define === 'function'))) {
-    return undefined;
+var webrtcUtils = {
+  log: function() {
+    // suppress console.log output when being included as a module.
+    if (!(typeof module !== 'undefined' ||
+        (typeof require === 'function') && (typeof define === 'function'))) { 
+      console.log.apply(console, arguments);
+    }
   }
-  return console.log.apply(console, arguments);
-}
+};
+
 
 function trace(text) {
   // This function is used for logging.
@@ -39,17 +40,17 @@ function trace(text) {
   }
   if (window.performance) {
     var now = (window.performance.now() / 1000).toFixed(3);
-    webrtcLogger(now + ': ' + text);
+    webrtcUtils.log(now + ': ' + text);
   } else {
-    webrtcLogger(text);
+    webrtcUtils.log(text);
   }
 }
 
 if (typeof window === 'undefined' || !window.navigator) {
-  webrtcLogger('This does not appear to be a browser');
+  webrtcUtils.log('This does not appear to be a browser');
   webrtcDetectedBrowser = 'not a browser';
 } else if (navigator.mozGetUserMedia) {
-  webrtcLogger('This appears to be Firefox');
+  webrtcUtils.log('This appears to be Firefox');
 
   webrtcDetectedBrowser = 'firefox';
 
@@ -130,8 +131,10 @@ if (typeof window === 'undefined' || !window.navigator) {
       }
       return c;
     };
+    webrtcUtils.log('spec: ' + JSON.stringify(c));
     c.audio = constraintsToFF37(c.audio);
     c.video = constraintsToFF37(c.video);
+    webrtcUtils.log('ff37: ' + JSON.stringify(c));
     return navigator.mozGetUserMedia(c, onSuccess, onError);
   } : navigator.mozGetUserMedia.bind(navigator);
 
@@ -178,7 +181,7 @@ if (typeof window === 'undefined' || !window.navigator) {
   };
 
 } else if (navigator.webkitGetUserMedia) {
-  webrtcLogger('This appears to be Chrome');
+  webrtcUtils.log('This appears to be Chrome');
 
   webrtcDetectedBrowser = 'chrome';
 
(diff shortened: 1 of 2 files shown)


adapter.js
```diff
@@ -23,6 +23,15 @@ var webrtcDetectedBrowser = null;
 var webrtcDetectedVersion = null;
 var webrtcMinimumVersion = null;
 
+function webrtcLogger() {
+  // suppress console.log output when being included as a module
+  if (typeof module !== 'undefined' ||
+      ((typeof require === 'function') && (typeof define === 'function'))) {
```
KaptenJansson: Remove parentheses around the 2nd and 3rd typeof operators, i.e:

```
(typeof module !== 'undefined' ||
  typeof require === 'function' && typeof define === 'function')
```

http://google-styleguide.googlecode.com/svn/trunk/javascriptguide.xml?showone=Parentheses#Parentheses

adapter.js
```diff
@@ -23,6 +23,15 @@ var webrtcDetectedBrowser = null;
 var webrtcDetectedVersion = null;
 var webrtcMinimumVersion = null;
 
+function webrtcLogger() {
+  // suppress console.log output when being included as a module
```
KaptenJansson: Add full stop "."

adapter.js
```diff
@@ -23,6 +23,15 @@ var webrtcDetectedBrowser = null;
...
+  if (typeof module !== 'undefined' ||
+      ((typeof require === 'function') && (typeof define === 'function'))) {
+    return undefined;
```
alvestrand: For my curiosity's sake: Why should this function return anything at all?

adapter.js
```diff
@@ -23,6 +23,15 @@ var webrtcDetectedBrowser = null;
 var webrtcDetectedVersion = null;
 var webrtcMinimumVersion = null;
 
+function webrtcLogger() {
```
alvestrand: Is this a new function in the global namespace? Can we wrap it inside some module object rather than continuing our tradition of polluting the global namespace with our utility functions?

alvestrand:
I'm not sure removing the logging of constraint rewrites is a Good Idea at this time... can you make that into a separate PR so that we can get Jan-Ivar's comment on that removal separately?

KaptenJansson:
Also could you add test that exercises this?

adapter.js
fippo: good question, especially since the default return value is undefined anyway.... removed.

adapter.js
fippo: Along the lines of 

```
webrtcUtils = {
   log: function () { ... }
};
```

? This still pollutes, but at least we can limit this in the future

adapter.js
jan-ivar: why not invert the if?

jan-ivar:
@alvestrand yeah it might be premature to remove the constraint conversion logging, if only because the conversion is imperfect (it could go further by implementing fallbacks in the advanced array to simulate the ideal algorithm).

fippo:
- added the constraint logs back
- introduced webrtcUtils object which currently only contains a log item.

@KaptenJansson testing this is somewhat tricky since the test requires this as a module. Might be possibly by overriding console and looking for an adapter-specific log but...

alvestrand:
Can we test this by letting the test function temporarily override console.log and verify that it's not called?

ie let the test be

savelog = console.log
console.log = function(args) { t.fail() }
actionThatCausesLog
console.log = savelog

Doesn't work for async/promise-driven actions, of course.

alvestrand:
(note - it's possible that my suggestion doesn't work because tape uses console.log for printing. We can figure that out by trying....)

fippo:
I did something evil... WDYT?

```
var logcount = 0;
var c = console.log.bind(console);
console.log = function() {
  logcount++;
  c.apply(c, arguments);
};
var m = require('../adapter.js');
console.log = c;

test('log suppression', function (t) {
  console.log('logcount', logcount);
});
```

When including adapter, console.log should not be called and that can be asserted as logcount === 0
Note that you need the additional .bind(console) when saving console.log.

alvestrand:
That sounds twisted enough for a Javascript test :-)
Let's do it! (if it works...)

adapter.js
```diff
@@ -22,6 +22,15 @@ var reattachMediaStream = null;
 var webrtcDetectedBrowser = null;
 var webrtcDetectedVersion = null;
 var webrtcMinimumVersion = null;
+var webrtcUtils = {
+  log: function() {
+    // suppress console.log output when being included as a module.
+    if (!(typeof module !== 'undefined' ||
+        (typeof require === 'function') && (typeof define === 'function'))) {
```
KaptenJansson: Remove parentheses around the 2nd and 3rd typeof operators, i.e:

```
(typeof module !== 'undefined' ||
  typeof require === 'function' && typeof define === 'function')
```

http://google-styleguide.googlecode.com/svn/trunk/javascriptguide.xml?showone=Parentheses#Parentheses

test/test.js
```diff
@@ -5,7 +5,22 @@
 /* global Promise */
 var test = require('tape');
 
+// adapter.js is not supposed to spill console.log
+// when used as a module. This temporarily overloads
+// console.log so we can assert this.
+var logcount = 0;
+var saveconsole = console.log.bind(console);
```
KaptenJansson: Use camelCase, applies below as well.

test/test.js
```diff
@@ -5,7 +5,22 @@
 /* global Promise */
 var test = require('tape');
 
+// adapter.js is not supposed to spill console.log
+// when used as a module. This temporarily overloads
+// console.log so we can assert this.
+var logcount = 0;
```
KaptenJansson: Use camelCase, applies below as well.

fippo:
@KaptenJansson done. too bad jscs can't check camel case (even though there is http://jscs.info/rule/requireCamelCaseOrUpperCaseIdentifiers.html ?)

KaptenJansson:
It should be able to and I believe I've actually seen it do so before, filed #84 for this.

KaptenJansson:
LGTM

PR #75, opened 2015-07-13
make sure getUserMedia var is set in Edge

fippo:
getUserMedia was never set in Edge. 
While I think we should get rid of that variable and make everyone (especially the samples, PR incoming) use navigator.getUserMedia instead, this solves the issue without requiring a major version bump.

### DIFF
--- a/adapter.js
+++ b/adapter.js
@@ -368,6 +368,8 @@ if (typeof window === 'undefined' || !window.navigator) {
   // the minimum version still supported by adapter.
   webrtcMinimumVersion = 12;
 
+  getUserMedia = navigator.getUserMedia;
+
   attachMediaStream = function(element, stream) {
     element.srcObject = stream;
   };


jan-ivar:
> use navigator.getUserMedia instead

or ideally navigator.mediaDevices.getUserMedia.

KaptenJansson:
> use navigator.getUserMedia instead
> 
> > or ideally navigator.mediaDevices.getUserMedia.

Then you need to shim Chrome and Opera. Seems like FF is the only one so far.

alvestrand:
Agree on navigator.mediaDevices.getUserMedia as the target. We should shim everything else with a function that does console.log('This function is deprecated. Please use navigator.mediaDevices.getUserMedia') in addition to calling the function.

But that's a different PR. I'm approving this one.
(still no thoughts about testing Edge? Can we run the test page manually on Edge?)

KaptenJansson:
> use navigator.getUserMedia instead
> 
> > or ideally navigator.mediaDevices.getUserMedia.
> > 
> > > Then you need to shim Chrome and Opera. Seems like FF is the only one so far.

I totally missed that we actually shim this in adapter.js. Maybe @fippo could change PR https://github.com/webrtc/samples/pull/548 to accommodate this for the samples repo?

Issue #41, opened 2015-06-08
Need to support static methods on RTCPeerConnection


alvestrand:
In implementing generateCertificate, which is a static method on RTCPeerConnection, we discovered that defining RTCPeerConnection as a function (and doing no more) hid the static method in Chrome (if the RTCPeerConnection class declaration was made visible).

Is it enough / correct to add "RTCPeerConnection.prototype = webkitRTCPeerConnection.prototype" to get static functions to work, or do we need to do something more complex?
We don't have any static functions so far, so I have nothing to test on.

Asking @fippo @jan-ivar for advice.

henbos:
So far while testing on a forked version of samples, I changed this in adapter.js to make it work locally:
  // old code (line 182)
  RTCPeerConnection = function(pcConfig, pcConstraints) {
    return new webkitRTCPeerConnection(pcConfig, pcConstraints);
  };
  // new code
  RTCPeerConnection.generateCertificate = webkitRTCPeerConnection.generateCertificate;

jan-ivar:
@alvestrand - There's [Object.setPrototypeOf](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/setPrototypeOf) but see performance warnings.

Why did we need a wrapper again (rather than modifying the prototype directly?)

fippo:
I tried fiddling with the constructor in Firefox -- http://jsfiddle.net/zx7a95w1/1/ -- but that didn't work (doesn't for chrome either). Since currently we only need to do that for Firefox <38 to fiddle with the RTCIceServers let's just remove the wrapper for Chrome and hope that by the time it is necessary for Firefox, pre-38 versions will be gone?

jan-ivar:
Or you could restore the `createIceServer` function. ;)

What's the version range that adapter.js tried to cover?

FWIW got a question from someone using Firefox 22 [just the other day](http://stackoverflow.com/questions/30581417/ns-error-unexpected-in-firefox-on-mozrtcpeerconnection/30600209#30600209).

jan-ivar:
@fippo - re-reading your response I think what you suggest is ok (removing it for Chrome only). You could remove it on Firefox 38+ as well.

fippo:
I think the answer is to use Object.defineProperty, i.e. 

```
Object.defineProperty(RTCPeerConnection, 'generateCertificate', {
  get: function() { return webkitRTCPeerConnection.generateCertificate(arguments); }
})
```

[...]

PR #77, opened 2015-07-14
add promise support for getStats

fippo:
taking a stab a promise support for getStats... see https://github.com/w3c/webrtc-pc/pull/235 for the spec.

Overloading arguments in JS is slightly tricky but I think this will also remove the problem with #41 which was reintroduced with the awesome getStats shim.

Firefox added this in FF37 according to @jan-ivar and technically the minimum version of FF supported by adapter is currently 31... I don't really feel like supporting that though so let's bump the min version here?

### DIFF
--- a/adapter.js
+++ b/adapter.js
@@ -223,7 +223,7 @@ if (typeof window === 'undefined' || !window.navigator) {
       }
 
       // promise-support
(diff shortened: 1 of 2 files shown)


fippo:
looks like I can't fix the problem from #41 like this... reverting to just adding promise support

PR #78, opened 2015-07-14
Fix add/removeEventListener for Chrome unstable

### LINKED ISSUE #73: No op add/removeEventListener methods
Currently Chrome unstable (45) has the `navigator.medaDevices` object but without `addEventListener` and `removeEventListener` methods.

This causes two [tests](https://travis-ci.org/webrtc/adapter/jobs/70575016) to fail because we conditionally add `addEventListener` and `removeEventListener`functions to `navigator.medaDevices` only if `!navigator.mediaDevices`.

@jan-ivar, @alvestrand, @fippo this was added in #45 to prevent some crash. Should we add something like this:

```
if (typeof navigator.mediaDevices.removeEventListener === 'undefined' ||
      typeof navigator.mediaDevices.addEventListener === 'undefined') {
    // in case someone wants to listen for the devicechange event.
    navigator.mediaDevices.addEventListener = function() { };
    navigator.mediaDevices.removeEventListener = function() { };
```

  }

to make the tests pass on Chrome unstable?

KaptenJansson:
This fixes the Chrome unstable navigator.mediaDevices.addEventListener tests even though they only check that the function exists.

@alvestrand, @fippo Not sure about the log text, any ideas?

NOTE: I will update console.log to use whatever is decided in #74.

Fixes #73

### DIFF
--- a/adapter.js
+++ b/adapter.js
@@ -353,9 +353,18 @@ if (typeof window === 'undefined' || !window.navigator) {
         });
       });
     }};
-    // in case someone wants to listen for the devicechange event.
-    navigator.mediaDevices.addEventListener = function() { };
-    navigator.mediaDevices.removeEventListener = function() { };
+  }
+  // Dummy devicechange event methods.
+  // TODO(KaptenJansson) remove once implemented in Chrome stable.
+  if (typeof navigator.mediaDevices.addEventListener === 'undefined') {
+    navigator.mediaDevices.addEventListener = function() {
+      console.log('Dummy mediaDevices.addEventListener OK');
+    };
+  }
+  if (typeof navigator.mediaDevices.removeEventListener === 'undefined') {
+    navigator.mediaDevices.removeEventListener = function() {
+      console.log('Dummy mediaDevices.removeEventListener OK');
+    };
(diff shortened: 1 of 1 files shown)


KaptenJansson:
The Firefox enumerateDevice test hang is annoying and it also confirms that the test timeout is 10 something minutes ;).

adapter.js
```diff
@@ -353,9 +353,18 @@ if (typeof window === 'undefined' || !window.navigator) {
...
     }};
-    // in case someone wants to listen for the devicechange event.
-    navigator.mediaDevices.addEventListener = function() { };
-    navigator.mediaDevices.removeEventListener = function() { };
+  }
+  // Dummy devicechange event methods.
+  // TODO(KaptenJansson) remove once implemented in Chrome stable.
+  if (typeof navigator.mediaDevices.addEventListener === 'undefined') {
+    navigator.mediaDevices.addEventListener = function() {
+      console.log('Dummy mediaDevices.addEventListener OK');
```
alvestrand: Suggest using "called" rather than "OK". Makes the log easier to read.

PR #79, opened 2015-07-15
Add shim for navigator.mediaDevices.getUserMedia when mediaDevices exist

KaptenJansson:
Fixes `Uncaught TypeError: navigator.mediaDevices.getUserMedia is not a function` currently seen in latest samples push, e.g. http://webrtc.github.io/samples/src/content/peerconnection/webaudio-input/ when using `--enable-blink-features=EnumerateDevices,AudioOutputDevices` flag on 45.0.2441.3 or later.

@alvestrand PTAL

PR #81, opened 2015-07-15
use window.RTCPeerConnection

fippo:
I figured we should use window.RTCPeerConnection instead of m.RTCPeerConnection in the tests.
And then i noticed we don't test this so now we are...

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -24,8 +24,10 @@ test('Browser supported by adapter.js', function(t) {
 test('getUserMedia shim', function(t) {
   t.ok(typeof navigator.getUserMedia !== 'undefined',
        'navigator.getUserMedia is defined');
-  t.ok(navigator.mediaDevices && typeof navigator.mediaDevices !== 'undefined',
+  t.ok(navigator.mediaDevices &&
+       typeof navigator.mediaDevices.getUserMedia !== 'undefined',
        'navigator.mediaDevices.getUserMedia is defined');
+  t.end();
 });
 
 test('RTCPeerConnection shim', function(t) {
@@ -35,6 +37,7 @@ test('RTCPeerConnection shim', function(t) {
        'RTCSessionDescription is defined');
   t.ok(typeof window.RTCIceCandidate !== 'undefined',
        'RTCIceCandidate is defined');
+  t.end();
 });
 
 test('create RTCPeerConnection', function(t) {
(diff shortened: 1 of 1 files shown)


test/test.js
```diff
@@ -21,6 +21,22 @@ test('Browser supported by adapter.js', function(t) {
       'Browser version supported by adapter.js');
 });
 
+test('getUserMedia shim', function(t) {
+  t.ok(typeof navigator.getUserMedia !== 'undefined',
+       'navigator.getUserMedia is defined');
+  t.ok(navigator.mediaDevices && typeof navigator.mediaDevices !== 'undefined',
+       'navigator.mediaDevices.getUserMedia is defined');
```
KaptenJansson, line 29: That's not true for Chrome, it has navigator.mediaDevices but not navigator.mediaDevices.getUserMedia hence you cannot assume this. Check my recent PR #79.

fippo, line 29: not natively, but it has after we require adapter, right?

fippo, line 29: ah... bug in the current shim. So we need to merge this after #79 :-)

KaptenJansson, line 29: Yes, after my PR #79 which I landed 10 mins ago but maybe you saw that? ;)

KaptenJansson, line 29: But I guess what I'm getting at is that we should not write that an explicit method exists unless we check for them explicitly.

fippo, line 29: oh... I actually intended to have

```
navigator.mediaDevices.getUserMedia !== undefined
```

there!

test/test.js
```diff
@@ -21,6 +21,25 @@ test('Browser supported by adapter.js', function(t) {
       'Browser version supported by adapter.js');
 });
 
+test('getUserMedia shim', function(t) {
+  t.ok(typeof navigator.getUserMedia !== 'undefined',
+       'navigator.getUserMedia is defined');
+  t.ok(navigator.mediaDevices &&
+       typeof navigator.mediaDevices.getUserMedia !== 'undefined',
+       'navigator.mediaDevices.getUserMedia is defined');
```
KaptenJansson, line 29: Do you think we should still keep this test after the test I added in https://github.com/webrtc/adapter/pull/79/files#diff-c1129c8b045390789fa8ff62f2c6b4a9R233?

KaptenJansson, line 29: @fippo ping ;)

fippo, line 29: yeah. This makes sure the variable is defined, slightly different from making sure it actually works. So this will pass even if getusermedia is broken internally, making it easier to pinpoint the failure.

KaptenJansson:
LGTM

PR #82, opened 2015-07-15
add getStats promises

fippo:
basically #77 but trying to do one thing at a time.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -288,3 +288,14 @@ test('originalChromeGetStats', function(t) {
     t.end();
   }
 });
+
+test('getStats promise', function(t) {
+  t.plan(2);
+  var pc1 = new m.RTCPeerConnection(null);
+
+  var p = pc1.getStats();
+  t.ok(typeof p === 'object', 'getStats with no arguments returns a Promise');
+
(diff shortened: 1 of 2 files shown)


test/test.js
```diff
@@ -288,3 +288,14 @@ test('originalChromeGetStats', function(t) {
...
+test('getStats promise', function(t) {
+  t.plan(2);
+  var pc1 = new m.RTCPeerConnection(null);
+
+  var p = pc1.getStats();
+  t.ok(typeof p === 'object', 'getStats with no arguments returns a Promise');
+
+  var q = pc1.getStats(null);
+  t.ok(typeof q === 'object', 'getStats with a selector returns a Promise');
+});
```
KaptenJansson, line 301: Add another test that test the old style getStats() (i.e. 2 arguments)?

KaptenJansson, line 301: Or Chrome style rather...

fippo, line 301: that's covered by the existing test I think

KaptenJansson, line 301: Yes you are right, https://github.com/fippo/adapter/blob/getstats-promise2/test/test.js#L238.

LGTM.

PR #83, opened 2015-07-17
fix constraints for navigator.mediaDevices.getUserMedia in canary

fippo:
As @selavazhagan pointed out in https://github.com/webrtc/samples/issues/564 the constraints sample is currently broken in Chrome 45.

This is because Chrome 45 adds a navigator.mediaDevices.getUserMedia which does not understand the spec constraints. (IMO it is bad to do so in an unprefixed version but that's why we have canaries)

Had to make a variable in order to avoid code duplication. Mostly because instead of relying on requestUserMedia I wanted to use the new native thing so it gets tested now.
@KaptenJansson: maybe add this into webrtcUtil instead after merging #74? Decide while I'm asleep :-)

Unfortunately, JS lambdas don't bind "properly" so doing

```
window.constraintsToChrome = function() { ... };
navigator.mediaDevices.getUserMedia = function(...) {
    constraintsToChrome(...)
})
delete window.constraintsToChrome;
```

doesn't work to avoid the global variable.

### DIFF
--- a/adapter.js
+++ b/adapter.js
@@ -348,14 +348,15 @@ if (typeof window === 'undefined' || !window.navigator) {
   }
 
   // A shim for getUserMedia method on the mediaDevices object.
-  // TODO(KaptenJansson) remove once in implemented in Chrome stable.
+  // TODO(KaptenJansson) remove once implemented in Chrome stable.
   if (!navigator.mediaDevices.getUserMedia) {
     navigator.mediaDevices.getUserMedia = function(constraints) {
       return requestUserMedia(constraints);
     };
   } else {
-    // even though Chrome 45 has navigator.mediaDevices and a getUserMedia
-    // function which returns a Promise, it does not accept spec-style constraints.
+    // Even though Chrome 45 has navigator.mediaDevices and a getUserMedia
+    // function which returns a Promise, it does not accept spec-style
+    // constraints.
     var origGetUserMedia = navigator.mediaDevices.getUserMedia.
         bind(navigator.mediaDevices);
     navigator.mediaDevices.getUserMedia = function(c) {
--- a/adapter.js
+++ b/adapter.js
@@ -263,57 +263,58 @@ if (typeof window === 'undefined' || !window.navigator) {
   });
 
   // getUserMedia constraints shim.
-  getUserMedia = function(c, onSuccess, onError) {
-    var constraintsToChrome = function(c) {
-      if (typeof c !== 'object' || c.mandatory || c.optional) {
-        return c;
+  var constraintsToChrome = function(c) {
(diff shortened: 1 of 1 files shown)


KaptenJansson:
This is strange, navigator.mediaDevices is no longer behind a flag which it should be (Enable experimental Web Platform features or --enable-blink-features=EnumerateDevices.). When I tested this on 45.0.2441.3 (yes it's old by now;)) it worked fine with or without the flag.

navigator.mediaDevices.getUserMedia() promise implementation was done here: https://chromium.googlesource.com/chromium/blink/+/eddec454b6021559cd7a31f3a13c9b3e8a184642

adapter.js
```diff
@@ -355,11 +341,23 @@ if (typeof window === 'undefined' || !window.navigator) {
   }
 
   // A shim for getUserMedia method on the mediaDevices object.
-  // TODO(KaptenJansson) remove once in implented in Chrome stable.
+  // TODO(KaptenJansson) remove once in implemented in Chrome stable.
```
KaptenJansson: Double type from me;) 

Should be `TODO(KaptenJansson) remove once implemented in Chrome stable.`

adapter.js
```diff
@@ -355,11 +341,23 @@ if (typeof window === 'undefined' || !window.navigator) {
...
   if (!navigator.mediaDevices.getUserMedia) {
     navigator.mediaDevices.getUserMedia = function(constraints) {
       return requestUserMedia(constraints);
     };
+  } else {
+    // even though Chrome 45 has navigator.mediaDevices and a getUserMedia
```
KaptenJansson: Start with capital.

adapter.js
```diff
@@ -355,11 +341,23 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+    // function which returns a Promise, it does not accept spec-style constraints.
```
KaptenJansson: Strange that jshint/jscs does not complain about it being longer than 80 chars.

KaptenJansson:
> This is because Chrome 45 adds a navigator.mediaDevices.getUserMedia which does not understand the spec constraints. (IMO it is bad to do so in an unprefixed version but that's why we have canaries)

hta@ Do you know why this is? Should we file a bug about this since it's no longer behind a flag (which it should be?)?

fippo:
affected version is 45.0.2453.0 btw

KaptenJansson:
@fippo Thanks for doing this, I've tested it and it works great. Just wondering if there are more tests that can be done so we cover all bases.

KaptenJansson:
FTR filed a bug about why it's no longer behind a flag: https://code.google.com/p/chromium/issues/detail?id=511226

KaptenJansson:
LGTM.

PR #86, opened 2015-07-19
add firefox extended support release to travis

fippo:
This adds Firefox ESR (https://www.mozilla.org/en-US/firefox/organizations/faq/) to the test matrix.

Local node_modules might have to be wiped, npm ls should show djo-shell as 1.4.0
I wonder if we should bump the minimum supported version of Firefox to that... it's better to test things instead of just intending them.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ env:
     - BROWSER=firefox BVER=stable
     - BROWSER=firefox BVER=beta
     - BROWSER=firefox BVER=unstable
+    - BROWSER=firefox BVER=esr
 
 matrix:
   fast_finish: true


fippo:
btw, I _think_ I have seen a way to get older chromium versions from @braveyao once. Is that public/supported enough we can use it to test stable-2 versions?

KaptenJansson:
LGTM

KaptenJansson:
> btw, I think I have seen a way to get older chromium versions from @braveyao once. Is that public/supported enough we can use it to test stable-2 versions?

Maybe he referred to the snapshots you can download [here](https://www.chromium.org/getting-involved/download-chromium)? Then I guess we have to add that to testling/travis-multirunner in order to effectively download a new browser.

fippo:
i think it was a more obscure URL... if I only remembered....

KaptenJansson:
I created #87 to track the investigation into the chromium stable-2 possibility.

PR #89, opened 2015-07-21
make npm add run-tests to node_modules/bin

fippo:
this should add run-tests to node_modules/.bin/ so we don't have to duplicate it in the samples (and I don't have to duplicate it in a number of my own repos)

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,6 +8,9 @@
     "type": "git",
     "url": "https://github.com/webrtc/adapter.git"
   },
+  "bin": {
+    "run-tests": "./test/run-test"
+  },
   "scripts": {
     "test": "test/run-tests"
   },


fippo:
hrm... test failure in FF (see https://travis-ci.org/webrtc/adapter/jobs/71991976) seems unrelated to this commit. Let's try again dear travis...

KaptenJansson:
> hrm... test failure in FF (see https://travis-ci.org/webrtc/adapter/jobs/71991976) seems unrelated to this commit. Let's try again dear travis...

Issue (test will be skipped rather on particular firefox versions) will be fixed in #90

KaptenJansson:
By duplicate you mean that you run-test when you install adapter via NPM? Also FYI run-test is different in adapter/ and samples/ due to using selenium in the latter.

fippo:
ah... that makes it a selfish request because I want to use an unmodified version in some repositories :-)

KaptenJansson:
LGTM

KaptenJansson:
That went a bit fast, oops ;). Well as mentioned this script is different for the repos hence it cannot be reused as it seemed to be intention of this PR. Closing.

PR #90, opened 2015-07-21
Skip constraint test on firefox <42 (fixes travis regression)

jan-ivar:
Fixes failures on https://travis-ci.org/webrtc/adapter/jobs/71994516#L414 which were introduced by https://github.com/DamonOehlman/travis-multirunner/pull/7

The problem is that firefox's fake devices don't ([yet](https://bugzilla.mozilla.org/show_bug.cgi?id=1182938)) support constraints, and there's no way to turn off fake devices Firefox <42.

The test still runs on 42+ and on Chrome.


KaptenJansson:
LGTM after you fix the errors (lines 97 and 109 exceed 80 char in test/tests.js).

PR #93, opened 2015-07-26
translate iceTransportPolicy to iceTransports

fippo:
until https://code.google.com/p/webrtc/issues/detail?id=4869 is fixed I don't want to write code that works around the wrong name if I can just get it fixed in adapter.
The two things using this are the candidate gathering sample and apprtc (?it=relay)

The test is indirect. If the policy is set and we try to gather candidates without a TURN server, we will not get any candidates.
Firefox is not supported yet so the test passes. @jan-ivar I failed to find a bugzilla id for this?

### DIFF
--- a/adapter.js
+++ b/adapter.js
@@ -210,6 +210,12 @@ if (typeof window === 'undefined' || !window.navigator) {
 
   // The RTCPeerConnection object.
   window.RTCPeerConnection = function(pcConfig, pcConstraints) {
+    // Translate iceTransportPolicy to iceTransports,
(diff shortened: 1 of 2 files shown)


fippo:
the test failure is rather interesting. It seems that ICE gathering never completes on travis? Works for me locally...

boo... such a nice test! But I think we can go without it

jan-ivar:
https://bugzil.la/1187775

KaptenJansson:
LGTM.

Filed #96 to investigate why the test fails.

PR #98, opened 2015-07-28
Revert "make npm add run-tests to node_modules/bin"

KaptenJansson:
Reverts webrtc/adapter#89


KaptenJansson:
Turns out I did not test that enough.

PR #102, opened 2015-07-30
shim srcObject

fippo:
@alvestrand -- you mentioned "experiment with setting src on video" a while back.
I'm not sure how I found https://developers.google.com/web/updates/2015/04/DOM-attributes-now-on-the-prototype but it was incredibly helpful...

@jan-ivar take your time with renaming mozSrcObject :-)

This isn't done yet. Most importantly, reattachMediaStream in Chrome is incomplete yet. I think that setting srcObject should set a googSrcObject (?) property which the get accessor can return.

Also, should we remove attachMediaStream and reattachMediaStream and bump the major version? Probably some other cleanup to do if we break backward compat anyway...

This probably also requires Chrome 43+, probably Paul Kinlan can confirm. Heck, this "move stuff to the prototype chain" in 43 has had positive effects!

samples-tests still pass, merge with confidence

### DIFF
--- a/adapter.js
+++ b/adapter.js
@@ -442,10 +442,20 @@ if (typeof window === 'undefined' || !window.navigator) {
 
   // Attach a media stream to an element.
   attachMediaStream = function(element, stream) {
-    element.srcObject = stream;
+    if (webrtcDetectedVersion >= 43) {
+      element.srcObject = stream;
+    } else if (typeof element.src !== 'undefined') {
+      element.src = URL.createObjectURL(stream);
+    } else {
+      webrtcUtils.log('Error attaching stream to element.');
+    }
   };
   reattachMediaStream = function(to, from) {
-    to.srcObject = from.srcObject;
+    if (webrtcDetectedVersion >= 43) {
+      to.srcObject = from.srcObject;
+    } else {
+      to.src = from.src;
+    }
   };
 
 } else if (navigator.mediaDevices && navigator.userAgent.match(
--- a/adapter.js
+++ b/adapter.js
@@ -187,13 +187,22 @@ if (typeof window === 'undefined' || !window.navigator) {
       });
     };
   }
+
+  Object.defineProperty(HTMLVideoElement.prototype, 'srcObject', {
+    get: function() {
+      return this.mozSrcObject;
+    },
+    set: function(stream) {
+      this.mozSrcObject = stream;
+    }
+  });
   // Attach a media stream to an element.
   attachMediaStream = function(element, stream) {
-    element.mozSrcObject = stream;
+    element.srcObject = stream;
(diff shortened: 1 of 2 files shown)


adapter.js
```diff
@@ -420,19 +429,19 @@ if (typeof window === 'undefined' || !window.navigator) {
     };
   }
 
+  Object.defineProperty(HTMLVideoElement.prototype, 'srcObject', {
+    set: function(stream) {
+      this.src = URL.createObjectURL(stream);
```
alvestrand, line 439: Could you just add "this._srcObject = stream" here? (with the obvious getter)

alvestrand:
This makes me happy :-)

BUT: You need the tests...... with all the pseudo-flags, including --allow-file-access-from-files, added to the latest travis-multirunner, it should be possible to run the tests even when capturing media (which is what stymied me when I experimented with it).

adapter.js
fippo, line 439: yeah, was just pondering the name.

fippo, line 439: hrm. Doesn't work. Or rather: in the getter the _srcObject I set is undefined again. Will try to use a closure approach.

Pesky tests, uncovering errors in my thinking!

fippo:
I added tests. They work in Firefox but fail in Chrome locally so the travis failure was kind of expected.

What I don't understand: removing the first attachMediaStream/reattachMediaStreams tests makes things pass for me. So this depends on the order?!

There is no issue with calling things many times manually in the samples so i'm lost :-(

(note: we should probably wait until 45 before merging this or at least do a conservative approach which retains things for <43. Until we decide to do a big major version bump cleanup)

fippo:
the failure seems unrelated. I can repro this hanging with the old code as well. See https://github.com/fippo/adapter/tree/metadatabug -- this is related to either where onloadedmetadata is hooked up or the stream.stop. Working and non-working commit. @KaptenJansson probably worth filing a chrome bug?

fippo:
ok, tests pass but I think I just added move enough loadedmetadata+stop to make things pass...

KaptenJansson:
I know there has been issues in Chrome regarding the onloadedetadata event not firing for video elements a while back, not sure if it was fixed, we had to work around it last time. Actually found the bug  https://code.google.com/p/chromium/issues/detail?id=110938. It's closed but I'm willing to reopen it with fresh data and sample code. WDYT @fippo ?

fippo:
decided not to break compat yet -- we can still do that after M45 is stable, i've heard that up to 5% of users are still on < M43. @samdutton unless Paul knows a way to achieve the same thing in < 43?

What we can do however is merge and update the samples.

KaptenJansson:
LGTM

PR #103, opened 2015-07-30
Bump version to 0.1.9

KaptenJansson:


KaptenJansson:
Bumping version to get the last fixes and also to fix my screw up by not creating Releases in github for the last few versions which makes it impossible (github only allows you to create release for 20 something commits back in time) to push a new version to bower.

KaptenJansson:
The error for the last commit happened because I deleted the branch to fast....

PR #104, opened 2015-07-30
Fix ReferenceError in FF39

ghost:
```
> window.RTCSessionDescription = mozRTCSessionDescription; console.debug("ping");
ReferenceError: mozRTCSessionDescription is not defined

> window.RTCSessionDescription = window.mozRTCSessionDescription; console.debug("ping");
undefined
"ping"
```

Since the ReferenceError breaks my dependency file I cannot ignore the error. Thoughts on this one?
Cheers Lukas

related to sualko/strophe.jinglejs#1

### DIFF
--- a/adapter.js
+++ b/adapter.js
@@ -91,10 +91,10 @@ if (typeof window === 'undefined' || !window.navigator) {
   };
 
   // The RTCSessionDescription object.
-  window.RTCSessionDescription = mozRTCSessionDescription;
+  window.RTCSessionDescription = window.mozRTCSessionDescription;
 
   // The RTCIceCandidate object.
-  window.RTCIceCandidate = mozRTCIceCandidate;
+  window.RTCIceCandidate = window.mozRTCIceCandidate;
 
   // getUserMedia constraints shim.
   getUserMedia = function(constraints, onSuccess, onError) {


ghost:
I signed it!

fippo:
@Zauberstuhl are you using the latest version? I recall something like that breaking a while ago. Probably worth doing just to get rid of the jshint hints

ghost:
@fippo that is also a problem for me in the newest version.

I mean if you check a undefined var in JS you should do that in the window object as well, no?

```
> if (someVar === undefined) { console.debug("ping"); }
ReferenceError: someVar is not defined
> if (window.someVar === undefined) { console.debug("ping"); }
undefined
"ping"
```

fippo:
@Zauberstuhl yes, but what worries me is that this code should never be executed on chrome since it's in an if (mozilla)

ghost:
thats the thing I am using firefox v39

fippo:
@Zauberstuhl plain firefox or a variant with webrtc disabled? (moonsomething)

jan-ivar:
Any add-ons? @Zauberstuhl can you try launching `firefox -P` and test this in a brand new profile?

jan-ivar:
You might also want to check that `media.peerconnection.enabled` is `true` in `about:config`.

ghost:
> You might also want to check that media.peerconnection.enabled is true in about:config.

... ok well black magic and now it works. Thanks guys!
But still if someone disabled it shouldn't there be some kind of error handling? Or what happens with older FF versions?

jan-ivar:
Good point. I don't care so much about (very) old versions, but webrtc being disabled seems like something we should catch and tell people about, since they'll likely want to re-enable given that they came here.

fippo:
I think alot of other stuff would break if we detected a firefox with this disabled as webrtc-capable so i'd go the #106 way

jan-ivar:
Agree. My comment on error handling probably belong in https://webrtc.github.io/samples since adapter.js shouldn't talk to the user.

ghost:
Since you already merged #106 I am closing this ID! Thanks for the fast help

PR #106, opened 2015-07-30
check for mozGetUserMedia and mozRTCPeerConnection

fippo:
I think there is an old issue mentioning this somewhere... but this should be the proper fix for #104


KaptenJansson:
LGTM

fippo:
@KaptenJansson can you bump on npm please?

PR #108, opened 2015-08-01
Fix promise getstats

fippo:
```
pc.getStats(null).then(...)
```

didn't fix up the stats. Not sure how to test that...

(plus removed the TODO since it just works like charm)


KaptenJansson:
LGTM

PR #111, opened 2015-08-03
unbreak logging, fixes #110

### LINKED ISSUE #110: Logging is broken
This was found by @tednakamura.

The following currently does not work when loaded as a module in nodejs hence the [test](https://github.com/webrtc/adapter/blob/master/test/test.js#L21) does not work. Also it only tests the "module" case, filed #109 to cover the "non-module" case.

```
var webrtcUtils = {
  log: function() {
    // suppress console.log output when being included as a module.
    if (!(typeof module !== 'undefined' ||
        typeof require === 'function') && (typeof define === 'function')) {
      console.log.apply(console, arguments);
    }
  }
};
```

Also do note, this breaks logging when not loaded as a module, as seen in the basic datachannel [demo](https://github.com/webrtc/samples/blob/master/src/content/datachannel/basic/js/main.js#L40).

@fippo I took a quick stab at it but could not get it work and since I did not implement this and do not know to much about modules, could you take a look? However if you do not have time in the nearest future, I will take another stab at it.

fippo:
the old expression did... unexpected things.

### DIFF
--- a/adapter.js
+++ b/adapter.js
@@ -25,10 +25,11 @@ var webrtcMinimumVersion = null;
 var webrtcUtils = {
   log: function() {
     // suppress console.log output when being included as a module.
-    if (!(typeof module !== 'undefined' ||
-        typeof require === 'function') && (typeof define === 'function')) {
-      console.log.apply(console, arguments);
+    if (typeof module !== 'undefined' ||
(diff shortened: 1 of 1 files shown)


KaptenJansson:
LGTM

PR #114, opened 2015-08-03
Bump version to 0.2.1

KaptenJansson:
Turned out postinstall grunt githooks prevented it from being installed


KaptenJansson:
Travis is having serious delay issue but several are green, gonna let this pass.

PR #117, opened 2015-08-06
Add editorconfig for helping enforce code style

xdumaine:
I find myself having to change my editor settings from 4 to 2 spaces - having an editorconfig in the project makes this happen automatically.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,14 @@
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
+end_of_line = lf
+insert_final_newline = true
+
+# Matches multiple files with brace expansion notation
+# Set default charset
+[*.{js, json}]
+charset = utf-8
+indent_style = space
+indent_size = 2


KaptenJansson:
Well it's only true if you have an editor with native support, lots of editors requires plugins.

KaptenJansson:
@alvestrand what's your take on this?

xdumaine:
Yeah, hence _helps_ - otherwise it takes manual code review to catch, which is already the case, so it's not a _guarantee_ but it helps. Auto configuring editor & catching/failing them in tests would require this as well as linter.

alvestrand:
As it stands, this seems to be an automagic file with no information about what it will affect, and when.
I tend to distrust magic dot-files that try to reconfigure things for the user - they will surprise anyone (like me) who has never heard of .editorconfig.

Please add to the file:
- What editors are affected by .editorconfig
- What files are affected by .editorconfig (this dir, or all subdirs)
- Where the spec for .editorconfig is.

Issue #109, opened 2015-08-03
Add console.log test case for when not loaded as a module


KaptenJansson:
Current test case only validates the case when added as a module.

fippo:
I just tried the following as the last test

```
test('Non-module logging still works', function(t) {
  var logCount = 0;
  var saveConsole = console.log.bind(console);
  console.log = function() {
    logCount++;
    saveConsole.apply(saveConsole, arguments);
  };
  var script = document.createElement('script');
  script.src = 'adapter.js';
  (document.head||document.documentElement).appendChild(script);
  script.parentNode.removeChild(script);

  console.log = saveConsole;
  console.log(logCount);
});
```

While that doesn't work because the script is loaded async, it runs into an exception:
Uncaught TypeError: Cannot redefine property: srcObject

So that doesn't work.

fippo:
will make a PR once #119 is in.

PR #119, opened 2015-08-06
Make srcObject polyfill more generic

xdumaine:
This will help with firefox unprefixing srcObject h/t @fippo

### DIFF
--- a/adapter.js
+++ b/adapter.js
@@ -46,7 +46,9 @@ function trace(text) {
   }
 }
 
-if (typeof window !== 'undefined' && window.HTMLVideoElement &&
+// In a browser with support for, HTMLMediaElement define
+// srcObject on the prototype for attaching media streams to elements
+if (typeof window !== 'undefined' && window.HTMLMediaElement &&
     !window.HTMLMediaElement.prototype.srcObject) {
   webrtcUtils.log('Creating HTMLMediaElement.prototype.srcObject');
   Object.defineProperty(window.HTMLMediaElement.prototype, 'srcObject', {
--- a/adapter.js
+++ b/adapter.js
@@ -46,11 +46,8 @@ function trace(text) {
   }
 }
 
-if (typeof window === 'undefined' || !window.HTMLVideoElement) {
-  webrtcUtils.log('HTMLVideoElement is not supported');
-} else if (window.HTMLVideoElement.prototype.srcObject) {
-  webrtcUtils.log('srcObject is supported natively');
-} else {
+if (typeof window !== 'undefined' && window.HTMLVideoElement &&
+    !window.HTMLVideoElement.prototype.srcObject) {
   webrtcUtils.log('Creating HTMLVideoElement.prototype.srcObject');
   Object.defineProperty(HTMLVideoElement.prototype, 'srcObject', {
     get: function() {
--- a/adapter.js
+++ b/adapter.js
@@ -47,9 +47,9 @@ function trace(text) {
 }
 
 if (typeof window !== 'undefined' && window.HTMLVideoElement &&
-    !window.HTMLVideoElement.prototype.srcObject) {
-  webrtcUtils.log('Creating HTMLVideoElement.prototype.srcObject');
-  Object.defineProperty(HTMLVideoElement.prototype, 'srcObject', {
+    !window.HTMLMediaElement.prototype.srcObject) {
+  webrtcUtils.log('Creating HTMLMediaElement.prototype.srcObject');
+  Object.defineProperty(window.HTMLMediaElement.prototype, 'srcObject', {
     get: function() {
       return this.mozSrcObject || this._srcObject;
     },
--- a/adapter.js
+++ b/adapter.js
@@ -49,7 +49,7 @@ function trace(text) {
 // In a browser with support for HTMLMediaElement, define srcObject
 // on the prototype (only once!) for attaching media streams to elements.
 if (typeof window === 'object' && window.HTMLMediaElement &&
-    !window.HTMLMediaElement.prototype.srcObject) {
+    !window.HTMLMediaElement.prototype.hasOwnProperty('srcObject')) {
   Object.defineProperty(window.HTMLMediaElement.prototype, 'srcObject', {
     get: function() {
       return this.mozSrcObject || this._srcObject;
--- a/adapter.js
+++ b/adapter.js
@@ -48,7 +48,7 @@ function trace(text) {
 
 // In a browser with support for HTMLMediaElement, define srcObject
 // on the prototype (only once!) for attaching media streams to elements
-if (typeof window !== 'undefined' && window.HTMLMediaElement &&
+if (typeof window === 'object' && window.HTMLMediaElement &&
     !window.HTMLMediaElement.prototype.srcObject) {
   Object.defineProperty(window.HTMLMediaElement.prototype, 'srcObject', {
     get: function() {
@@ -65,7 +65,7 @@ if (typeof window !== 'undefined' && window.HTMLMediaElement &&
   });
 }
 
-if (typeof window === 'undefined' || !window.navigator) {
+if (typeof window !== 'object' || !window.navigator) {
   webrtcUtils.log('This does not appear to be a browser');
   webrtcDetectedBrowser = 'not a browser';
 } else if (navigator.mozGetUserMedia && window.mozRTCPeerConnection) {
@@ -505,7 +505,7 @@ Object.defineProperty(webrtcTesting, 'version', {
 
 if (typeof module !== 'undefined') {
   var RTCPeerConnection;
-  if (typeof window !== 'undefined') {
+  if (typeof window === 'object') {
     RTCPeerConnection = window.RTCPeerConnection;
   }
   module.exports = {
--- a/adapter.js
+++ b/adapter.js
@@ -46,6 +46,27 @@ function trace(text) {
   }
 }
 
+if (typeof window === 'undefined' || !window.HTMLVideoElement) {
+  webrtcUtils.log('HTMLVideoElement is not supported');
+} else if (window.HTMLVideoElement.prototype.srcObject) {
+  webrtcUtils.log('srcObject is supported natively');
+} else {
+  webrtcUtils.log('Creating HTMLVideoElement.prototype.srcObject');
+  Object.defineProperty(HTMLVideoElement.prototype, 'srcObject', {
+    get: function() {
+      return this.mozSrcObject || this._srcObject;
+    },
+    set: function(stream) {
+      if (this.mozSrcObject) {
+        this.mozSrcObject = stream;
+      } else {
+        this._srcObject = stream;
+        this.src = URL.createObjectURL(stream);
+      }
+    }
+  });
+}
+
 if (typeof window === 'undefined' || !window.navigator) {
   webrtcUtils.log('This does not appear to be a browser');
   webrtcDetectedBrowser = 'not a browser';
@@ -189,14 +210,6 @@ if (typeof window === 'undefined' || !window.navigator) {
     };
   }
 
-  Object.defineProperty(HTMLVideoElement.prototype, 'srcObject', {
-    get: function() {
-      return this.mozSrcObject;
-    },
-    set: function(stream) {
-      this.mozSrcObject = stream;
-    }
-  });
   // Attach a media stream to an element.
   attachMediaStream = function(element, stream) {
     element.srcObject = stream;
@@ -436,16 +449,6 @@ if (typeof window === 'undefined' || !window.navigator) {
       webrtcUtils.log('Dummy mediaDevices.removeEventListener called.');
     };
   }
-
-  Object.defineProperty(HTMLVideoElement.prototype, 'srcObject', {
-    get: function() {
-      return this._srcObject;
-    },
-    set: function(stream) {
-      this._srcObject = stream;
-      this.src = URL.createObjectURL(stream);
-    }
-  });
 
   // Attach a media stream to an element.
   attachMediaStream = function(element, stream) {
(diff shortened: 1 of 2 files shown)


fippo:
+1 for detangling things

xdumaine:
@fippo sure. wasn't sure about how much logging was preferred. Updated.

adapter.js
```diff
@@ -46,6 +46,24 @@ function trace(text) {
   }
 }
 
+if (typeof window !== 'undefined' && window.HTMLVideoElement &&
+    !window.HTMLVideoElement.prototype.srcObject) {
+  webrtcUtils.log('Creating HTMLVideoElement.prototype.srcObject');
+  Object.defineProperty(HTMLVideoElement.prototype, 'srcObject', {
```
jan-ivar: Sorry for not noticing this sooner, but note that srcObject is on `HTMLMediaElement`, not just `HTMLVideoElement`. e.g. it [works for audio as well](http://jsfiddle.net/dy944rag).

fippo: oh. Does inheritance work like one would expect? We should have an audio element test to cover this!

xdumaine: Changed to use `HTMLMediaElement` and added a test for setting `srcObject` on an `HTMLAudioElement`

jan-ivar: Yes inheritance should work right.

fippo:
actually this solves the redefine issue I ran into with #109 as well (and I just had to downgrade the adapter ref in a bunch of dependent modules)

adapter.js
```diff
@@ -46,6 +46,24 @@ function trace(text) {
   }
 }
```
KaptenJansson, line 48: Could you add a short comment about the purpose of this? 

The reason I pick on this particular PR is that I've just realized we could improve on the comments a little, at least add a reason for why it's added.

KaptenJansson:
LGTM with nit.

adapter.js
```diff
@@ -46,6 +46,24 @@ function trace(text) {
...
 
+if (typeof window !== 'undefined' && window.HTMLVideoElement &&
```
jan-ivar: Nit: I'd prefer `if (window && window.HTML...` here over the typedef !== stuff, as we don't want to de-reference window on _any_ falsy values. If we want to be narrow, we could use `if (typeof window === 'object' && window.HTML...`, but that's a bit overkill imho.

jan-ivar: Also you missed a s/HTMLVideoElement/HTMLMediaElement/ here.

adapter.js
```diff
@@ -46,6 +46,24 @@ function trace(text) {
...
+    !window.HTMLMediaElement.prototype.srcObject) {
+  webrtcUtils.log('Creating HTMLMediaElement.prototype.srcObject');
```
jan-ivar: Also missed a log.

adapter.js
```diff
@@ -46,6 +46,24 @@ function trace(text) {
   }
 }
 
+if (typeof window !== 'undefined' && window.HTMLVideoElement &&
+    !window.HTMLMediaElement.prototype.srcObject) {
```
fippo: I think this needs to use prototype.hasOwnProperty('srcObject')
-- otherwise you run into trouble when requiring adapter multiple times along the way because you can't redefine properties.

```
HTMLVideoElement.prototype.hasOwnProperty('srcObject'); // -> false
Object.defineProperty(HTMLVideoElement.prototype, 'srcObject', {get: function() { return this._srcObject; } });
HTMLVideoElement.prototype.hasOwnProperty('srcObject'); // -> true
```

adapter.js
xdumaine: @jan-ivar `if (window && window.HTMLMediaElement)` will throw a reference error when required in a node context. Detection of window with `typeof window !== undefined` adds safety there.

xdumaine:
@KaptenJansson comment added. @jan-ivar removed last log, and corrected `HTMLVideoElement` to `HTMLMediaElement`. See my other comment regarding why `if(window)` isn't safe.

adapter.js
jan-ivar: Ok makes sense. How about `typeof window === 'object'` then?

xdumaine: Sure, should I also change it on [L49](https://github.com/webrtc/adapter/blob/master/adapter.js#L49) where I'm using `typeof window === 'undefined'`?

jan-ivar: Sure why not.

adapter.js
```diff
@@ -46,7 +46,26 @@ function trace(text) {
   }
 }
 
-if (typeof window === 'undefined' || !window.navigator) {
+// In a browser with support for HTMLMediaElement, define srcObject
+// on the prototype (only once!) for attaching media streams to elements
```
KaptenJansson: Sorry to be picky, could you add punctuation ?

(should really have linting rules to enforce stuff like this, will look into it.)

xdumaine:
Good to go @jan-ivar @KaptenJansson?

xdumaine:
@jan-ivar because it's not a static property; it's on prototype, right? `srcObject` defined on the prototype of `HTMLMediaElement` won't exist at `HTMLMediaElement.srcObject`.

And when defining on the prototype, you'd need to use `hasOwnProperty` because a property check on the prototype won't catch it. To expand on @fippo's example:

``` javascript
HTMLMediaElement.prototype.hasOwnProperty('srcObject'); // -> false
HTMLMediaElement.prototype.srcObject; // falsy
Object.defineProperty(HTMLMediaElement.prototype, 'srcObject', {get: function() { return this._srcObject; } });
HTMLMediaElement.prototype.hasOwnProperty('srcObject'); // true
HTMLMediaElement.prototype.srcObject; // -> falsy 
```

So if we don't check `.prototype.hasOwnProperty('srcObject')` an error could thrown when we attempt to define the property because it might already be defined. Your example without `.prototype` in the check won't work at all.

jan-ivar:
Thanks for explaining. TIL. LGTM!

fippo:
Just manually checked in Edge btw. HTMLMediaElement.prototype.hasOwnProperty('srcObject') return true there so. phew :-)

adapter.js
```diff
@@ -46,7 +46,26 @@ function trace(text) {
...
+        this.mozSrcObject = stream;
+      } else {
+        this._srcObject = stream;
+        this.src = URL.createObjectURL(stream);
+      }
+    }
+  });
+}
+
+if (typeof window !== 'object' || !window.navigator) {
```
alvestrand, line 68: What's the reason for putting the test for srcObject in front of the test for "is this a browser at all"? This looks backwards to me.

adapter.js
```diff
@@ -46,7 +46,26 @@ function trace(text) {
...
-if (typeof window === 'undefined' || !window.navigator) {
+// In a browser with support for HTMLMediaElement, define srcObject
+// on the prototype (only once!) for attaching media streams to elements.
+if (typeof window === 'object' && window.HTMLMediaElement &&
+    !window.HTMLMediaElement.prototype.hasOwnProperty('srcObject')) {
+  Object.defineProperty(window.HTMLMediaElement.prototype, 'srcObject', {
+    get: function() {
+      return this.mozSrcObject || this._srcObject;
+    },
+    set: function(stream) {
```
alvestrand, line 57: I actually don't like this disentanglement style. It means that code will be executed taht checks for a Mozilla-specific property on non-Mozilla browsers, which is not elegant.

Please explain why this is a Good Thing....

adapter.js
```diff
@@ -504,7 +505,7 @@ Object.defineProperty(webrtcTesting, 'version', {
 
 if (typeof module !== 'undefined') {
   var RTCPeerConnection;
-  if (typeof window !== 'undefined') {
+  if (typeof window === 'object') {
```
alvestrand, line 508: What's the case that warranted this change?

test/test.js
```diff
@@ -164,6 +164,25 @@ test('video srcObject getter/setter test', function(t) {
...
+    t.pass('srcObject set');
+    t.ok(audio.srcObject.id === stream.id,
+        'srcObject getter returns stream object');
+    stream.getTracks().forEach(function(track) { track.stop(); });
+  })
+  .catch(function(err) {
+    t.fail(err.toString());
+  });
+});
+
```
alvestrand, line 185: I like the test!

adapter.js
xdumaine, line 508: @jan-ivar asked that I change this. See https://github.com/webrtc/adapter/pull/119#discussion_r36480439

adapter.js
xdumaine, line 68: It's moved outside what is effectively a switch. It still has safety against non browser context.

adapter.js
fippo, line 57: this keeps things in one place, otherwise we have minor variants of this all over the place and need to make sure that any changes to one variant (like moving to HTMLMediaElement) is also made in the other.

adapter.js
```diff
@@ -46,7 +46,26 @@ function trace(text) {
...
+// In a browser with support for HTMLMediaElement, define srcObject
+// on the prototype (only once!) for attaching media streams to elements.
+if (typeof window === 'object' && window.HTMLMediaElement &&
+    !window.HTMLMediaElement.prototype.hasOwnProperty('srcObject')) {
+  Object.defineProperty(window.HTMLMediaElement.prototype, 'srcObject', {
+    get: function() {
+      return this.mozSrcObject || this._srcObject;
+    },
+    set: function(stream) {
+      if (this.mozSrcObject) {
```
fippo, line 58: should this be a hasOwnProperty? By default, mozSrcObject is null

adapter.js
xdumaine, line 57: It's not "not elegant," it's just a different approach. Your current approach is just as "not elegant" because it's terribly repetitive and subject to breakage without stringent code review (which you're pretty good at though) as well as tests. The current code design is "recreate the expected behavior for each browser individually" instead of "define the behavior once and adapt to each browser."

xdumaine:
This PR solves two significant bugs, passes existing tests, adds another test, and is being picked apart for style. Might I suggest that if you don't like the style, you open a PR that has one you prefer? I don't have the time or energy go to back and forth with you on this or argue elegance. This review process is hostile and not welcoming to contributors. It's easier for contributors and yourselves to merge working code and make minor adjustments as you see fit. If that process doesn't work for you, that's your perogitive, but don't expect to see more PRs from the community.

adapter.js
alvestrand, line 57: The hope is of course that all new browsers (and all existing browsers, once they get their bugs fixed) will not need any code, so that the polyfill for the two existing non-conformant browsers is all the polyfill that will ever be needed.
This hope may be false or true....

KaptenJansson:
Sorry to hear that, we are currently looking into making sure our tools catch more nits and simple style issues (still lacking in some respects, suggestions are welcome of course) which should reduce some of the "nitpicky" comments. 

But in regards to the overall code approach/style you have to expect it to be scrutinized and analyzed thoroughly, different people will have different set of questions based on their angle/experience. This is true in general when doing code reviews, it takes time but in the end, everyone will have learnt something and the outcome will generally be higher quality code.

Also, the first PR is obviously going to be the hardest as you learn the procedures and guidelines for the repository. If you have any suggestions on improving this, please feel free to file issues/PR's (as you have in #117).

xdumaine:
A strict or stringent code review process is not what makes a hostile environment. It's the fact that during that process, not one person made any effort make it apparent that my changes were welcome or valued. 

The review process was simply: please make this code amalgamate into our existing code with the least disruption. One team member even emailed me that I should be grateful already because you took time after hours or on weekends to do this. 

Thanking contributors or providing encouragement isn't easy, I get that. It's also not natural - especially if the project is a part of your day job. For others, it's not. There is no inherent value to me in getting code merged into your repo, and I'm not getting paid for my work in doing so. For much of the open source community "giving back" is all the reward they need, and that's how I felt too, except it quickly changed to feet that I was just an annoyance. Because of that, I would rather spend my efforts "giving back" to places where I feel valued, not like a nuisance. 

It shouldn't be simply up to the reviewee to "get over" the difficulty of the process. Reviewers should be open to changing (not the strictness of review) but the language used, so that the reviewee feels like it's an effort worth pursuing.

KaptenJansson:
> A strict or stringent code review process is not what makes a hostile environment. It's the fact that during that process, not one person made any effort make it apparent that my changes were welcome or valued.

We will do our best in the future to make sure contributors feel welcome and appreciated in the WebRTC repos. Thank you for the detailed and actionable criticism.

> It shouldn't be simply up to the reviewee to "get over" the difficulty of the process. Reviewers should be open to changing (not the strictness of review) but the language used, so that the reviewee feels like it's an effort worth pursuing.

Off course it's a two way street, but one have to be able to have a discussion about the style without the risk of having a fallout after a few comments. But in this case I do understand that the initial welcoming might have been detrimental here.

austinkelleher:
I think this change is good and it LGTM. I agree that there is a difference between style and nitpicking. Asking for punctuation in a comment is nitpicking and is quite unrelated to the change itself. I definitely agree that style is important, but there is a point where it is overkill and it distracts from the original change. It seems like a majority of these comments were more focused on style. It would be a shame if this didn't get merged in.

fippo:
IMHO the javascript magic in the srcObject functionality is so arcane that it should be done in _one_ place and it should be documented (I volunteer to write a paragraph or two)

alvestrand:
@fippo I can see that argument. I understand that the arcane-ness of this stuff has already bitten you.
I'm willing to hold my misgivings and accept this approach.

KaptenJansson:
@xdumaine Sorry for how this turned out, but thanks for the effort you put in nonetheless. Some of it is now merged via @calebboyd's PR. 

(And FYI I will be less picky on the style now in the comments and rather make sure the linting tools catch punctuation errors etc instead.).

Issue #71, opened 2015-07-12
Include support for iOS cordova-plugin-iosrtc?


calebboyd:
The implementation: [iosrtc](https://github.com/eface2face/cordova-plugin-iosrtc) is (afaik) following spec. the adapter is able to work by altering the UA check [here](https://github.com/webrtc/adapter/blob/master/adapter.js#L205) but that is less than ideal when the actual browser is indeed a WebView.

alvestrand:
What platforms does the iosrtc plugin run on? (apart from IOS, which I assume..)
We should support it to the same level as we support EDGE - but we need a PR with code that detects the plugin and the browser, and it's likely not possible to run that under our unit tests....

calebboyd:
Just iOS > 7. I'll see about submitting a PR

However, There are several issues that @ibc points out [here](https://groups.google.com/forum/?#!topic/cordova-plugin-iosrtc/_QS5oT0pJjU)...

ibc:
adapter.js seems to be JS loaders' friendly (at the bottom of the script it does some checks for "module" and "require") but the fact is that, anyhow, it still exposes `window.RTCPeerConnection` and so on regardless the script is loaded when a JS loader is present. That is not the polite way at all but the old/legacy one in which for every JS library a separate `<script>` tag is required and a global exported to the `window` namespace (not cool in 2015).

Said that, adapter.js may still be loaded within a Cordova app + cordova-plugin-iosrtc if the whole JavaScript code uses a module loader (browserify, requireJS, etc) and the adapter.js module is loaded once Cordova's "deviceready" event is fired. But yes, it needs some work in adapter.js to detect the environment. It should be easy: just check whether `window.cordova.plugins.iosrtc` exists and, if so, call to `cordova.plugins.iosrtc.registerGlobals()` so the code running after that will find the WebRTC globals in the `window` namespace.

The problem (in Cordova) is when a library try to detect WebRTC global classes/functions at the time the JS script itself is loaded.

ibc:
BTW this throws an error when adapter.js is loaded in Cordova + cordova-plugins-iosrtc:

``` js
 // the detected chrome version.
  webrtcDetectedVersion =
    parseInt(navigator.userAgent.match(/Chrom(e|ium)\/([0-9]+)\./)[2], 10);
```

[...]

PR #120, opened 2015-08-11
shim srcObject and set export defaults

### LINKED ISSUE #94: webkitGetUserMedia adapter fails in WebKit browsers other than Chrome
The demo at https://webrtc.github.io/samples/src/content/getusermedia/gum/ fails right now on BlackBerry 10 devices because the JavaScript assumes that Chrome is the only WebKit-based browser that supports navigator.webkitGetUserMedia:

``` javascript
} else if (navigator.webkitGetUserMedia) {
  webrtcUtils.log('This appears to be Chrome');

  webrtcDetectedBrowser = 'chrome';

  // the detected chrome version.
  webrtcDetectedVersion =
    parseInt(navigator.userAgent.match(/Chrom(e|ium)\/([0-9]+)\./)[2], 10);
```

A TypeError is thrown because "Chrom(e|ium)" does not appear in the BB10 User-Agent.

The sample works on BlackBerry 10.3.2 when I use Web Inspector to override the User-Agent and masquerade as Chrome.

Could we make webrtcDetectedVersion a bit more robust? There are other WebKit-based browsers out there that would probably trip over this too. It would be nice if the code could recover gracefully from navigator.userAgent not containing "Chrom(e|ium)", even if it ends up treating the browser the same as Chrome.

(The BlackBerry 10 browser is based on WebKit and supports webkitGetUserMedia in all releases. In BlackBerry 10.3.2, the BlackBerry Browser additionally supports WebRTC, but it doesn't currently support Promises or WebAudio.)

calebboyd:
This PR provides a similar shim from #119, Except that the property detection will not return the non-prefixed value incorrectly, when the prefixed value is falsy.

Its main purpose though is to initialize the exports (getUserMedia, attachMediaStream, reattachMediaStream) so that a browser that doesn't need "detection" (one that supports spec or has been shimmed to do so prior to running adapter.js) can still have those exports provided. (imho the adapter shouldn't take the responsibility of providing detected browser to its dependents)

It also addresses chromium "detection" which fixes issues in #71 (removes the need to detect a plugin that registers globally available spec compliant methods) and should fix #94 (can't confirm).

### DIFF
--- a/adapter.js
+++ b/adapter.js
@@ -46,6 +46,41 @@ function trace(text) {
   }
 }
 
+if (typeof window === 'object') {
+  if (window.HTMLMediaElement &&
+    !('srcObject' in window.HTMLMediaElement.prototype)) {
+    // Shim the srcObject property, once, when HTMLMediaElement is found.
+    Object.defineProperty(window.HTMLMediaElement.prototype, 'srcObject', {
+      get: function() {
+        // If prefixed srcObject property exists, return it.
+        // Otherwise use the shimmed property, _srcObject
+        return 'mozSrcObject' in this ? this.mozSrcObject : this._srcObject;
+      },
+      set: function(stream) {
+        if ('mozSrcObject' in this) {
+          this.mozSrcObject = stream;
+        } else {
+          // Use _srcObject as a private property for this shim
+          this._srcObject = stream;
+          // TODO: revokeObjectUrl(this.src) when !stream to release resources?
+          this.src = URL.createObjectURL(stream);
+        }
+      }
+    });
+  }
+  // Proxy existing globals
+  getUserMedia = window.navigator && window.navigator.getUserMedia;
+}
+
+// Attach a media stream to an element.
+attachMediaStream = function(element, stream) {
+  element.srcObject = stream;
+};
+
+reattachMediaStream = function(to, from) {
+  to.srcObject = from.srcObject;
+};
+
 if (typeof window === 'undefined' || !window.navigator) {
   webrtcUtils.log('This does not appear to be a browser');
   webrtcDetectedBrowser = 'not a browser';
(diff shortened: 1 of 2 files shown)


adapter.js
```diff
@@ -46,6 +46,41 @@ function trace(text) {
...
+      }
+    });
+  }
+  // Proxy existing globals
+  getUserMedia = window.navigator && window.navigator.getUserMedia;
+}
+
+// Attach a media stream to an element.
+attachMediaStream = function(element, stream) {
+  element.srcObject = stream;
```
fippo, line 77: not yet, this breaks chrome < 43 which is still 5% of users according to numbers I have access to.
This requires a major version bump too.

adapter.js
```diff
@@ -46,6 +46,41 @@ function trace(text) {
...
+        // Otherwise use the shimmed property, _srcObject
+        return 'mozSrcObject' in this ? this.mozSrcObject : this._srcObject;
+      },
+      set: function(stream) {
+        if ('mozSrcObject' in this) {
+          this.mozSrcObject = stream;
+        } else {
+          // Use _srcObject as a private property for this shim
+          this._srcObject = stream;
+          // TODO: revokeObjectUrl(this.src) when !stream to release resources?
```
fippo, line 65: that comment is obsolete (removed in c9850d5bea8f1a713b0a8fd997a09343fad7551f)

adapter.js
calebboyd, line 77: This is just for default behavior. It is still reset for chrome < 43

adapter.js
calebboyd, line 65: Yes, the null pass through is no longer a todo. But I raised this todo as question for a potential memory leak of lingering objectUrls. (unless they are gc'd when dereferenced, my understanding from mdn is that it stays around until the document is reloaded)

adapter.js
fippo, line 77: ah, didn't see that in the diff. fab.

fippo:
@alvestrand I think avoiding yet another browser (kind-of) case is an even better argument for moving this!

adapter.js
alvestrand, line 65: This is a known problem with the objectURL mechanism in general - I think I remember that people tried to alleviate it by passing "invalidate when used" arguments to the createObjectURL function - but that created even more problems than it solved.
It's one reason why createObjectURL is a pattern we shouldn't emulate. But it's OK to not solve that problem in adapter.js.

alvestrand:
For testing: Can we wrap the code block at line 49 in a function? (call it installShimProperties or something like that)?
That would make it easy to call it twice in a test without loading adapter.js twice, so that we can verify that the detection of already-installed shims works.

KaptenJansson:
@calebboyd Thanks for PR and the detailed description!

I agree for testing wrapping the code at line 49 in a function would be preferable however then we leak another function name in the global namespace. Not an issue when loading adapter.js as a module as you defined what functions to expose.

alvestrand:
On 08/12/2015 10:05 AM, Christoffer Jansson wrote:

> @calebboyd https://github.com/calebboyd Thanks for PR and the
> detailed description!
> 
> I agree for testing wrapping the code at line 49 in a function would
> be preferable however then we leak another function name in the global
> namespace. Not an issue when loading adapter.js as a module as you
> defined what functions to expose.

Good point.

We should introduce a "function {}()" scope that we can put local
variables inside, but I think this is beyond the scope of the current PR.

This LGTM.

Issue #47, opened 2015-06-09
Add tests for all public functions in adapter.js to detect interop issues


KaptenJansson:
https://github.com/webrtc/samples/issues/523 was caused by webrtc/adapter@04af828 where window.createIceServer was removed.

We should add tests for this, just a rudimentary check like "typeof foo !== undefined" for all expected public api functions/methods is better than nothing.

Also its easier then to make sure that when we do break something we properly update the version number to indicate it's incompatible.

kashifshamaz21:
+1, this would be very helpful as webrtc apis are still under changes and lot of interop issues are expected for now.

KaptenJansson:
Believe this has been solved in the refactoring of adapter.

PR #121, opened 2015-08-12
add test for loading non-module, fixes #109

### LINKED ISSUE #109: Add console.log test case for when not loaded as a module

fippo:
@KaptenJansson -- I wonder if we should also test the intentionally globalized variables so we can modularize things without breaking stuff.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -739,6 +739,7 @@ test('iceTransportPolicy is translated to iceTransports', function(t) {
     // not implemented yet.
     t.pass('iceTransportPolicy is not implemented by Firefox yet.');
     t.end();
+    return;
   }
   var pc1 = new RTCPeerConnection({iceTransportPolicy: 'relay',
       iceServers: []});
--- a/test/test.js
+++ b/test/test.js
@@ -775,7 +775,6 @@ test('Non-module logging to console still works', function(t) {
   var saveConsole = console.log.bind(console);
   console.log = function() {
     logCount++;
-    //saveConsole.apply(saveConsole, arguments);
   };
   var script = document.createElement('script');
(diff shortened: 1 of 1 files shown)


fippo:
so presumably the missing return in the iceTransportPolicy test went unnoticed because it was the last test... very odd. And very hard to spot.

test/test.js
```diff
@@ -766,3 +767,27 @@ test('iceTransportPolicy is translated to iceTransports', function(t) {
...
 });
+
+// This MUST to be the last test since it loads adapter
+// again which may result in unintended behaviour.
+test('Non-module logging to console still works', function(t) {
+  var logCount = 0;
+  var saveConsole = console.log.bind(console);
+  console.log = function() {
+    logCount++;
+    //saveConsole.apply(saveConsole, arguments);
```
KaptenJansson: Remove the commented code.

KaptenJansson:
Nice test and catch of the missing return (I wish github web ui could display the diff in the whole file so one can see the stuff around + the new changes).

I think it makes sense to test everything that we expose globally, filed #47 a while back that touches on this.

fippo:
checking for "public api" as well. Note that I did not consider webrtcUtils part of that.

fippo:
not sure why FF stable didn't return @ https://travis-ci.org/webrtc/adapter/jobs/75443265 -- all tests pass?

KaptenJansson:
That looks really odd, does not repro locally off course? ;)

fippo:
let's try again and see if it persists...

KaptenJansson:
Thanks for this! LGTM.

PR #126, opened 2015-08-18
make adapter available on gh-pages

### LINKED ISSUE #116: make adapter.js versions available on the gh-pages branch
coming from here: https://github.com/MicrosoftEdge/Demos/issues/90#issuecomment-127363723

the "easy" way to do this is probably to put adapter.js onto the gh-pages branch so we'd have them available as https://webrtc.github.io/adapter/adapter-0.2.1.js etc

The samples can then include that (or a -latest if we dare).
@samdutton @KaptenJansson?

fippo:
fixes #116 -- symlink support was a little tricky, but https://fippo.github.io/adapter/adapter-latest.js works so...

I don't think we need old versions so I just added the two 0.2.x ones.


KaptenJansson:
LGTM

PR #127, opened 2015-08-19
Firefox: check if RTCSessionDescription and RTCIceCandidate exist before overwriting them

fippo:
since i'm anticipating @jan-ivar dealing with [this](https://bugzilla.mozilla.org/show_bug.cgi?id=917092) soon :-)


KaptenJansson:
Yeah this makes sense, LGTM.

PR #131, opened 2015-08-24
dont use datachannels in tests

fippo:
removes createDataChannels from tests in favor of fake streams. I don't really want to comment on the 'why' right now.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -626,6 +626,65 @@ test('basic connection establishment with promise', function(t) {
   });
 });
 
+test('basic connection establishment with datachannel', function(t) {
+  var pc1 = new RTCPeerConnection(null);
+  var pc2 = new RTCPeerConnection(null);
+  var ended = false;
+  if (typeof pc1.createDataChannel !== 'function') {
+    t.pass('datachannel is not supported.');
+    t.end();
+    return;
+  }
+
+  pc1.oniceconnectionstatechange = function() {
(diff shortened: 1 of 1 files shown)


test/test.js
```diff
@@ -735,9 +745,9 @@ test('getStats promise', function(t) {
 });
 
 test('iceTransportPolicy is translated to iceTransports', function(t) {
-  if (m.webrtcDetectedBrowser === 'firefox') {
-    // not implemented yet.
-    t.pass('iceTransportPolicy is not implemented by Firefox yet.');
+  if (m.webrtcDetectedBrowser !== 'chrome') {
+    // Only chrome requires this.
```
alvestrand: This comment is wrong. The test was written so that it will pass on a) compliant browsers and b) browsers where the shim does what it needs to do. I appreciate wanting to not special-case all the browsers, but we still can't state it in this way.

alvestrand:
Can you abstract it out to a function "addDummyChannel", which can use createDataChannel where it works, and fake channel where it doesn't?
I don't like to not use datachannel for things where datachannel should work if spec is followed.

test/test.js
fippo: The test name is wrong -- it should test that setting iceTransportPolicy to relay has the desired effect. Which in Chrome implies that this is mapped to iceTransport in the constructor. Easy to fix.

fippo:
How about adding another basic connectivity check using datachannels?

Issue #133, opened 2015-09-10
IE8 does not like catch function


sualko:
IE8 throws "identifier expected" on [L222](https://github.com/webrtc/adapter/blob/master/adapter.js#L222).

Not sure, why it is evaluating this part and how to fix this.

4lejandrito:
Same problem here. I am currently trying to conditionally load the script only if webrtc is supported by the browser (for which I was using this script :D)

4lejandrito:
My guess is that catch is a reserved word (try...catch) and IE8 is not that permisive. I had this problem when one of they keys of an inline object was called "with" IE8 wouldn't like it.

4lejandrito:
would this be a hacky workaround?

``` js
orgEnumerateDevices()['catch'](function(e) {
```

4lejandrito:
I managed to support IE8 using my fork: https://github.com/4lejandrito/adapter

If someone maintaining the main repo is happy with it I could submit a pull request. For the time being you could switch to mine if you want. By the way I have only tested it including es5-shim/sham before my main bundle.

jan-ivar:
A better workaround might be `.then(null, function(e) { });` which is the same.

sualko:
In my opinion, both solutions have pros and cons, so it doesn't matter, which one is merged. But thanks for your work anyway. I think this is important if you concatenate all your js files.

4lejandrito:
Yeah @sualko that was my use case, I am using browserify. I am happy with my fork it has been live for a few days now in my personal project but will be happy to switch to the mainstream if this issue gets fixed.

By the way, I have seen the fix #134 only covers this specific problem for IE8. However I had to make one more fix to make it work which is in this commit https://github.com/4lejandrito/adapter/commit/2da3d0fffeb94240e6bb147e436a53cf23fd1fdd.

Cheers!

alvestrand:
.catch is part of the ES6 specification of promises. If IE8 doesn't like it, IE8 has a bug.
(Can someone verify that this is indeed true, and not just that it parses "return functionReturningPromise().catch()" differently from all other browsers? If it's as simple as saying "return (functionReturningPrimise().catch()) that might be a better patch.)

that said, adapter.js is all about papering over browsers' bugs, so there's an argument to be made for making this fix.

jan-ivar:
[...]

PR #134, opened 2015-09-15
s/.catch(f)/.then(null, f)/ to load on IE8

jan-ivar:
Fix https://github.com/webrtc/adapter/issues/133


sualko:
Could you also include https://github.com/4lejandrito/adapter/commit/2da3d0fffeb94240e6bb147e436a53cf23fd1fdd ?

PR #136, opened 2015-09-16
shim rtcpeerconnection ontop of ortc

fippo:
@juberti @pthatcherg I think you might be interested in this. A couple of rough edges but it passes most tests here (pending #131) and makes many of the samples work.

Many thanks to @shijuns and @aboba for answering all of my questions.

### DIFF
--- a/adapter.js
+++ b/adapter.js
@@ -545,7 +545,7 @@ if (typeof window === 'undefined' || !window.navigator) {
       }
 
       // per-track iceGathers etc
-      this.tracks = [];
+      this.mLines = [];
 
       this._iceCandidates = [];
 
@@ -902,7 +902,7 @@ if (typeof window === 'undefined' || !window.navigator) {
         if (!description.ortc) {
           // FIXME: throw?
         } else {
-          this.tracks = description.ortc;
+          this.mLines = description.ortc;
         }
       } else if (description.type === 'answer') {
         var sections = self.remoteDescription.sdp.split('\r\nm=');
@@ -910,16 +910,16 @@ if (typeof window === 'undefined' || !window.navigator) {
         sections.forEach(function(section, sdpMLineIndex) {
           section = 'm=' + section;
 
-          var iceGatherer = self.tracks[sdpMLineIndex].iceGatherer;
-          var iceTransport = self.tracks[sdpMLineIndex].iceTransport;
-          var dtlsTransport = self.tracks[sdpMLineIndex].dtlsTransport;
-          var rtpSender = self.tracks[sdpMLineIndex].rtpSender;
+          var iceGatherer = self.mLines[sdpMLineIndex].iceGatherer;
+          var iceTransport = self.mLines[sdpMLineIndex].iceTransport;
+          var dtlsTransport = self.mLines[sdpMLineIndex].dtlsTransport;
+          var rtpSender = self.mLines[sdpMLineIndex].rtpSender;
           var localCapabilities =
-              self.tracks[sdpMLineIndex].localCapabilities;
+              self.mLines[sdpMLineIndex].localCapabilities;
           var remoteCapabilities =
-              self.tracks[sdpMLineIndex].remoteCapabilities;
-          var sendSSRC = self.tracks[sdpMLineIndex].sendSSRC;
-          var recvSSRC = self.tracks[sdpMLineIndex].recvSSRC;
+              self.mLines[sdpMLineIndex].remoteCapabilities;
+          var sendSSRC = self.mLines[sdpMLineIndex].sendSSRC;
+          var recvSSRC = self.mLines[sdpMLineIndex].recvSSRC;
 
           var remoteIceParameters = self._getIceParameters(section,
               sessionpart);
@@ -1050,7 +1050,7 @@ if (typeof window === 'undefined' || !window.navigator) {
             rtpSender = new RTCRtpSender(localtrack, transports.dtlsTransport);
           }
 
-          self.tracks[sdpMLineIndex] = {
+          self.mLines[sdpMLineIndex] = {
             iceGatherer: transports.iceGatherer,
             iceTransport: transports.iceTransport,
             dtlsTransport: transports.dtlsTransport,
@@ -1064,13 +1064,13 @@ if (typeof window === 'undefined' || !window.navigator) {
             recvSSRC: recvSSRC
           };
         } else {
-          iceGatherer = self.tracks[sdpMLineIndex].iceGatherer;
-          iceTransport = self.tracks[sdpMLineIndex].iceTransport;
-          dtlsTransport = self.tracks[sdpMLineIndex].dtlsTransport;
-          rtpSender = self.tracks[sdpMLineIndex].rtpSender;
-          rtpReceiver = self.tracks[sdpMLineIndex].rtpReceiver;
-          sendSSRC = self.tracks[sdpMLineIndex].sendSSRC;
-          recvSSRC = self.tracks[sdpMLineIndex].recvSSRC;
+          iceGatherer = self.mLines[sdpMLineIndex].iceGatherer;
+          iceTransport = self.mLines[sdpMLineIndex].iceTransport;
+          dtlsTransport = self.mLines[sdpMLineIndex].dtlsTransport;
+          rtpSender = self.mLines[sdpMLineIndex].rtpSender;
+          rtpReceiver = self.mLines[sdpMLineIndex].rtpReceiver;
+          sendSSRC = self.mLines[sdpMLineIndex].sendSSRC;
+          recvSSRC = self.mLines[sdpMLineIndex].recvSSRC;
         }
 
         var remoteIceParameters = self._getIceParameters(section, sessionpart);
@@ -1123,7 +1123,7 @@ if (typeof window === 'undefined' || !window.navigator) {
             };
             rtpReceiver.receive(params, kind);
             stream.addTrack(rtpReceiver.track);
-            self.tracks[sdpMLineIndex].recvSSRC = recvSSRC;
+            self.mLines[sdpMLineIndex].recvSSRC = recvSSRC;
           }
         }
       });
@@ -1154,7 +1154,7 @@ if (typeof window === 'undefined' || !window.navigator) {
     };
 
     window.RTCPeerConnection.prototype.close = function() {
-      this.tracks.forEach(function(track) {
+      this.mLines.forEach(function(track) {
         /* not yet
         if (track.iceGatherer) {
           track.iceGatherer.close();
@@ -1192,7 +1192,7 @@ if (typeof window === 'undefined' || !window.navigator) {
         function(newState) {
       var self = this;
       if (this.iceConnectionState !== newState) {
-        var agreement = self.tracks.every(function(track) {
+        var agreement = self.mLines.every(function(track) {
           return track.iceTransport.state === newState;
         });
         if (agreement) {
@@ -1388,7 +1388,7 @@ if (typeof window === 'undefined' || !window.navigator) {
           'o=thisisadapterortc 8169639915646943137 2 IN IP4 127.0.0.1\r\n' +
           's=-\r\n' +
           't=0 0\r\n';
-      this.tracks.forEach(function(track/*, sdpMLineIndex*/) {
+      this.mLines.forEach(function(track/*, sdpMLineIndex*/) {
         var iceGatherer = track.iceGatherer;
         //var iceTransport = track.iceTransport;
         var dtlsTransport = track.dtlsTransport;
@@ -1461,7 +1461,7 @@ if (typeof window === 'undefined' || !window.navigator) {
     };
 
     window.RTCPeerConnection.prototype.addIceCandidate = function(candidate) {
-      var track = this.tracks[candidate.sdpMLineIndex];
+      var track = this.mLines[candidate.sdpMLineIndex];
       if (track) {
         var cand = Object.keys(candidate.candidate).length > 0 ?
             this._toCandidateJSON(candidate.candidate) : {};
@@ -1486,7 +1486,7 @@ if (typeof window === 'undefined' || !window.navigator) {
 
     window.RTCPeerConnection.prototype.getStats = function() {
       var promises = [];
-      this.tracks.forEach(function(track) {
+      this.mLines.forEach(function(track) {
         ['rtpSender', 'rtpReceiver', 'iceGatherer', 'iceTransport',
             'dtlsTransport'].forEach(function(thing) {
           if (track[thing]) {
--- a/adapter.js
+++ b/adapter.js
@@ -491,6 +491,15 @@ if (typeof window === 'undefined' || !window.navigator) {
   webrtcMinimumVersion = 10547;
 
   if (RTCIceGatherer) {
+    // Generate an alphanumeric identifier for cname or mids.
+    // TODO: use UUIDs instead? https://gist.github.com/jed/982883
+    var generateIdentifier = function() {
+      return Math.random().toString(36).substr(2, 10);
+    };
+
+    // The RTCP CNAME used by all peerconnections from the same JS.
+    var canonicalName = generateIdentifier();
+
     // SDP helpers from https://github.com/otalk/sdp-jingle-json with modifications.
     // Will be moved into separate module.
     var SDPUtils = {};
@@ -700,12 +709,6 @@ if (typeof window === 'undefined' || !window.navigator) {
       // must not emit candidates until after setLocalDescription we buffer
       // them in this array.
       this._localIceCandidatesBuffer = [];
-
-      this._peerConnectionId = 'PC_' + Math.floor(Math.random() * 65536);
-
-      // FIXME: Should be generated according to spec (guid?)
-      // and be the same for all PCs from the same JS
-      this._cname = Math.random().toString(36).substr(2, 10);
     };
 
     window.RTCPeerConnection.prototype.addStream = function(stream) {
@@ -960,7 +963,7 @@ if (typeof window === 'undefined' || !window.navigator) {
             params.muxId = sendSsrc;
             params.encodings = [self._getEncodingParameters(sendSsrc)];
             params.rtcp = {
-              cname: self._cname,
+              cname: canonicalName,
               reducedSize: false,
               ssrc: recvSsrc,
               mux: true
@@ -971,7 +974,7 @@ if (typeof window === 'undefined' || !window.navigator) {
             params.muxId = recvSsrc;
             params.encodings = [self._getEncodingParameters(recvSsrc)];
             params.rtcp = {
-              cname: self._cname,
+              cname: canonicalName,
               reducedSize: false,
               ssrc: sendSsrc,
               mux: true
@@ -1040,7 +1043,6 @@ if (typeof window === 'undefined' || !window.navigator) {
         })[0].substr(6);
 
         var cname;
-
         var remoteCapabilities;
         var params;
 
@@ -1116,7 +1118,7 @@ if (typeof window === 'undefined' || !window.navigator) {
             params.muxId = sendSsrc;
             params.encodings = [self._getEncodingParameters(sendSsrc)];
             params.rtcp = {
-              cname: self._cname,
+              cname: canonicalName,
               reducedSize: false,
               ssrc: recvSsrc,
               mux: true
@@ -1339,7 +1341,7 @@ if (typeof window === 'undefined' || !window.navigator) {
         // potentially rtpsender and rtpreceiver.
         var track = mline.track;
         var kind = mline.kind;
-        var mid = Math.random().toString(36).substr(2, 10);
+        var mid = generateIdentifier();
 
         var transports = self._createIceAndDtlsTransports(mid, sdpMLineIndex);
 
@@ -1411,7 +1413,7 @@ if (typeof window === 'undefined' || !window.navigator) {
           sdp += 'a=ssrc:' + sendSsrc + ' ' + 'msid:' +
               self.localStreams[0].id + ' ' + track.id + '\r\n';
         }
-        sdp += 'a=ssrc:' + sendSsrc + ' cname:' + self._cname + '\r\n';
+        sdp += 'a=ssrc:' + sendSsrc + ' cname:' + canonicalName + '\r\n';
       });
 
       var desc = new RTCSessionDescription({
@@ -1497,7 +1499,7 @@ if (typeof window === 'undefined' || !window.navigator) {
           sdp += 'a=ssrc:' + sendSsrc + ' ' + 'msid:' +
               self.localStreams[0].id + ' ' + rtpSender.track.id + '\r\n';
         }
-        sdp += 'a=ssrc:' + sendSsrc + ' cname:' + self._cname + '\r\n';
+        sdp += 'a=ssrc:' + sendSsrc + ' cname:' + canonicalName + '\r\n';
       });
 
       var desc = new RTCSessionDescription({
--- a/adapter.js
+++ b/adapter.js
@@ -1029,6 +1029,35 @@ if (typeof window === 'undefined' || !window.navigator) {
       };
     };
 
+    // Start the RTP Sender and Receiver for a transceiver.
+    window.RTCPeerConnection.prototype._transceive = function(transceiver,
+        send, recv) {
+      var params = this._getCommonCapabilities(transceiver.localCapabilities,
+          transceiver.remoteCapabilities);
+      if (send && transceiver.rtpSender) {
+        params.encodings = [{
+          ssrc: transceiver.sendSsrc,
+          active: true
+        }];
+        params.rtcp = {
+          cname: localCName,
+          ssrc: transceiver.recvSsrc
+        };
+        transceiver.rtpSender.send(params);
+      }
+      if (recv && transceiver.rtpReceiver) {
+        params.encodings = [{
+          ssrc: transceiver.recvSsrc,
+          active: true
+        }];
+        params.rtcp = {
+          cname: transceiver.cname,
+          ssrc: transceiver.sendSsrc
+        };
+        transceiver.rtpReceiver.receive(params);
+      }
+    };
+
     window.RTCPeerConnection.prototype.setLocalDescription =
         function(description) {
       var self = this;
@@ -1046,13 +1075,8 @@ if (typeof window === 'undefined' || !window.navigator) {
           var iceGatherer = transceiver.iceGatherer;
           var iceTransport = transceiver.iceTransport;
           var dtlsTransport = transceiver.dtlsTransport;
-          var rtpSender = transceiver.rtpSender;
-          var rtpReceiver = transceiver.rtpReceiver;
           var localCapabilities = transceiver.localCapabilities;
           var remoteCapabilities = transceiver.remoteCapabilities;
-          var cname = transceiver.cname;
-          var sendSsrc = transceiver.sendSsrc;
-          var recvSsrc = transceiver.recvSsrc;
 
           var remoteIceParameters = SDPUtils.getIceParameters(mediaSection,
               sessionpart);
@@ -1062,31 +1086,12 @@ if (typeof window === 'undefined' || !window.navigator) {
               sessionpart);
           dtlsTransport.start(remoteDtlsParameters);
 
-          // calculate intersection of capabilities
+          // Calculate intersection of capabilities.
           var params = self._getCommonCapabilities(localCapabilities,
               remoteCapabilities);
-          if (rtpSender && params.codecs.length) {
-            params.encodings = [{
-              ssrc: sendSsrc,
-              active: true
-            }];
-            params.rtcp = {
-              cname: localCName,
-              ssrc: recvSsrc
-            };
-            rtpSender.send(params);
-          }
-          if (rtpReceiver && params.codecs.length) {
-            params.encodings = [{
-              ssrc: recvSsrc,
-              active: true
-            }];
-            params.rtcp = {
-              cname: cname,
-              ssrc: sendSsrc
-            };
-            rtpReceiver.receive(params);
-          }
+          self._transceive(transceiver,
+              params.codecs.length > 0,
+              params.codecs.length > 0);
         });
       }
 
@@ -1157,7 +1162,7 @@ if (typeof window === 'undefined' || !window.navigator) {
               return obj.attribute === 'cname';
             })[0];
         if (remoteSsrc) {
-          recvSsrc = remoteSsrc.ssrc;
+          recvSsrc = parseInt(remoteSsrc.ssrc, 10);
           cname = remoteSsrc.value;
         }
 
@@ -1214,33 +1219,12 @@ if (typeof window === 'undefined' || !window.navigator) {
           iceTransport.start(iceGatherer, remoteIceParameters, 'controlling');
           dtlsTransport.start(remoteDtlsParameters);
 
-          // calculate intersection of capabilities
-          var params = self._getCommonCapabilities(localCapabilities,
-              remoteCapabilities);
-          if (rtpSender &&
-              (direction === 'sendrecv' || direction === 'recvonly')) {
-            params.encodings = [{
-              ssrc: sendSsrc,
-              active: true
-            }];
-            params.rtcp = {
-              cname: localCName,
-              ssrc: recvSsrc
-            };
-            rtpSender.send(params);
-          }
+          self._transceive(transceiver,
+              direction === 'sendrecv' || direction === 'recvonly',
+              direction === 'sendrecv' || direction === 'sendonly');
 
           if (rtpReceiver &&
               (direction === 'sendrecv' || direction === 'sendonly')) {
-            params.encodings = [{
-              ssrc: recvSsrc,
-              active: true
-            }];
-            params.rtcp = {
-              cname: cname,
-              ssrc: sendSsrc
-            };
-            rtpReceiver.receive(params);
             stream.addTrack(rtpReceiver.track);
           }
         }
@@ -1442,7 +1426,6 @@ if (typeof window === 'undefined' || !window.navigator) {
 
         // generate an ssrc now, to be used later in rtpSender.send
         var sendSsrc = (2 * sdpMLineIndex + 1) * 1001;
-        var recvSsrc; // don't know yet
         if (track) {
           rtpSender = new RTCRtpSender(track, transports.dtlsTransport);
         }
@@ -1462,7 +1445,7 @@ if (typeof window === 'undefined' || !window.navigator) {
           kind: kind,
           mid: mid,
           sendSsrc: sendSsrc,
-          recvSsrc: recvSsrc
+          recvSsrc: null
         };
 
         sdp += SDPUtils.writeRtpDescription(kind, localCapabilities);
@@ -1532,7 +1515,6 @@ if (typeof window === 'undefined' || !window.navigator) {
         var rtpReceiver = transceiver.rtpReceiver;
         var kind = transceiver.kind;
         var sendSsrc = transceiver.sendSsrc;
-        //var recvSsrc = transceiver.recvSsrc;
 
         // Calculate intersection of capabilities.
         var commonCapabilities = self._getCommonCapabilities(localCapabilities,
--- a/adapter.js
+++ b/adapter.js
@@ -1012,16 +1012,19 @@ if (typeof window === 'undefined' || !window.navigator) {
         }
       };
       iceTransport.onicestatechange = function() {
-        self._updateIceConnectionState();
+        self._updateConnectionState();
       };
 
       var dtlsTransport = new RTCDtlsTransport(iceTransport);
       dtlsTransport.ondtlsstatechange = function() {
-        //self._updateConnectionState();
+        self._updateConnectionState();
       };
-      dtlsTransport.onerror = function(error) {
-        console.error('dtls error', error);
+      dtlsTransport.onerror = function() {
+        // onerror does not set state to failed by itself.
+        dtlsTransport.state = 'failed';
+        self._updateConnectionState();
       };
+
       return {
         iceGatherer: iceGatherer,
         iceTransport: iceTransport,
@@ -1277,7 +1280,6 @@ if (typeof window === 'undefined' || !window.navigator) {
       });
       // FIXME: clean up tracks, local streams, remote streams, etc
       this._updateSignalingState('closed');
-      this._updateIceConnectionState('closed');
     };
 
     // Update the signaling state.
@@ -1295,8 +1297,8 @@ if (typeof window === 'undefined' || !window.navigator) {
       // TODO
     };
 
-    // Update the ICE connection state.
-    window.RTCPeerConnection.prototype._updateIceConnectionState =
+    // Update the connection state.
+    window.RTCPeerConnection.prototype._updateConnectionState =
         function() {
       var self = this;
       var newState;
@@ -1319,16 +1321,14 @@ if (typeof window === 'undefined' || !window.navigator) {
       newState = 'new';
       if (states.failed > 0) {
         newState = 'failed';
-      } else if (states.new + states.closed === 2 * this.transceivers.length) {
-        newState = 'new';
-      } else if (states.connecting + states.checking > 0) {
+      } else if (states.connecting > 0 || states.checking > 0) {
         newState = 'connecting';
-      } else if (states.connected > 0 &&
-          states.connected + states.closed === 2 * this.transceivers.length) {
-        newState = 'connected';
-      } else if (states.disconnected > 0 &&
-          (states.connecting === 0 && states.checking === 0)) {
+      } else if (states.disconnected > 0) {
         newState = 'disconnected';
+      } else if (states.new > 0) {
+        newState = 'new';
+      } else if (states.connecting > 0 || states.completed > 0) {
+        newState = 'connected';
       }
 
       if (newState !== self.iceConnectionState) {
--- a/adapter.js
+++ b/adapter.js
@@ -871,7 +871,7 @@ if (typeof window === 'undefined' || !window.navigator) {
         }
       };
       iceTransport.onicestatechange = function() {
-        self._updateIceConnectionState(iceTransport.state);
+        self._updateIceConnectionState();
       };
 
       var dtlsTransport = new RTCDtlsTransport(iceTransport);
--- a/adapter.js
+++ b/adapter.js
@@ -557,7 +557,10 @@ if (typeof window === 'undefined' || !window.navigator) {
       // everything that is needed to describe a SDP m-line.
       this.mLines = [];
 
-      this._iceCandidates = [];
+      // since the iceGatherer is currently created in createOffer but we
+      // must not emit candidates until after setLocalDescription we buffer
+      // them in this array.
+      this._localIceCandidatesBuffer = [];
 
       this._peerConnectionId = 'PC_' + Math.floor(Math.random() * 65536);
 
@@ -870,9 +873,9 @@ if (typeof window === 'undefined' || !window.navigator) {
             return mLine.iceGatherer && mLine.iceGatherer.state === 'completed';
           });
           if (self.localDescription && self.localDescription.type === '') {
-            self._iceCandidates.push(event);
+            self._localIceCandidatesBuffer.push(event);
             if (complete) {
-              self._iceCandidates.push({});
+              self._localIceCandidatesBuffer.push({});
             }
           } else {
             self.onicecandidate(event);
@@ -978,12 +981,12 @@ if (typeof window === 'undefined' || !window.navigator) {
       // FIXME: need to _reliably_ execute after args[1] or promise
       window.setTimeout(function() {
         // FIXME: need to apply ice candidates in a way which is async but in-order
-        self._iceCandidates.forEach(function(event) {
+        self._localIceCandidatesBuffer.forEach(function(event) {
           if (self.onicecandidate !== null) {
             self.onicecandidate(event);
           }
         });
-        self._iceCandidates = [];
+        self._localIceCandidatesBuffer = [];
       }, 50);
       if (arguments.length > 1 && typeof arguments[1] === 'function') {
         window.setTimeout(arguments[1], 0);
--- a/adapter.js
+++ b/adapter.js
@@ -807,7 +807,7 @@ if (typeof window === 'undefined' || !window.navigator) {
           self.onicecandidate(event);
         }
       });
-      self._localIceCandidatesBuffer = [];
+      this._localIceCandidatesBuffer = [];
     };
 
     window.RTCPeerConnection.prototype.addStream = function(stream) {
@@ -1054,16 +1054,17 @@ if (typeof window === 'undefined' || !window.navigator) {
           throw new TypeError('unsupported type "' + description.type + '"');
       }
 
+      window.setTimeout(this._emitBufferedCandidates.bind(this), 50);
       if (arguments.length > 1 && typeof arguments[1] === 'function') {
         var cb = arguments[1];
         window.setTimeout(function() {
           cb();
-          self._emitBufferedCandidates();
+          //self._emitBufferedCandidates();
         }, 0);
       }
       return new Promise(function(resolve) {
         resolve();
-        self._emitBufferedCandidates();
+        //self._emitBufferedCandidates();
       });
     };
 
--- a/adapter.js
+++ b/adapter.js
@@ -744,6 +744,46 @@ if (typeof window === 'undefined' || !window.navigator) {
           'a=ice-pwd:' + params.password + '\r\n';
     };
 
+    // Parses the SDP media section and return RTCRtpParameters.
+    SDPUtils.parseRtpDescription = function(mediaSection) {
+      var description = {
+        codecs: [],
+        headerExtensions: [],
+        fecMechanisms: [],
+        rtcp: []
+      };
+      var lines = SDPUtils.splitLines(mediaSection);
+      var mline = lines[0].substr(2).split(' ');
+      for (var i = 3; i < mline.length; i++) { // find all codecs from mline[3..]
+        var line = SDPUtils.matchPrefix(
+            mediaSection, 'a=rtpmap:' + mline[i] + ' ')[0];
+        if (line) {
+          var codec = SDPUtils.parseRtpMap(line);
+          var fmtp = SDPUtils.matchPrefix(
+              mediaSection, 'a=fmtp:' + mline[i] + ' ');
+          codec.parameters = fmtp.length ? SDPUtils.parseFmtp(fmtp[0]) : {};
+          codec.rtcpFeedback = SDPUtils.matchPrefix(
+              mediaSection, 'a=rtcp-fb:' + mline[i] + ' ')
+            .map(SDPUtils.parseRtcpFb);
+          description.codecs.push(codec);
+        }
+      }
+      // FIXME: parse headerExtensions, fecMechanisms and rtcp.
+      return description;
+    };
+
+    // Generates parts of the SDP media section describing the capabilities / parameters.
+    SDPUtils.writeRtpDescription = function(caps) {
+      var sdp = '';
+      caps.codecs.forEach(function(codec) {
+        sdp += SDPUtils.writeRtpMap(codec);
+        sdp += SDPUtils.writeFtmp(codec);
+        sdp += SDPUtils.writeRtcpFb(codec);
+      });
+      // FIXME: add headerExtensions, fecMechanismş and rtcp.
+      return sdp;
+    };
+
     // ORTC defines an RTCIceCandidate object but no constructor.
     // Not implemented in Edge.
     if (!window.RTCIceCandidate) {
@@ -840,44 +880,6 @@ if (typeof window === 'undefined' || !window.navigator) {
         this.localStreams.splice(idx, 1);
         this._maybeFireNegotiationNeeded();
       }
-    };
-
-    // This function parses an SDP media section to determine capabilities.
-    window.RTCPeerConnection.prototype._getRemoteCapabilities =
-        function(mediaSection) {
-      var remoteCapabilities = {
-        codecs: [],
-        headerExtensions: [],
-        fecMechanisms: []
-      };
-      var lines = SDPUtils.splitLines(mediaSection);
-      var mline = lines[0].substr(2).split(' ');
-      for (var i = 3; i < mline.length; i++) { // find all codecs from mline[3..]
-        var line = SDPUtils.matchPrefix(
-            mediaSection, 'a=rtpmap:' + mline[i] + ' ')[0];
-        if (line) {
-          var codec = SDPUtils.parseRtpMap(line);
-          var fmtp = SDPUtils.matchPrefix(
-              mediaSection, 'a=fmtp:' + mline[i] + ' ');
-          codec.parameters = fmtp.length ? SDPUtils.parseFmtp(fmtp[0]) : {};
-          codec.rtcpFeedback = SDPUtils.matchPrefix(
-              mediaSection, 'a=rtcp-fb:' + mline[i] + ' ')
-            .map(SDPUtils.parseRtcpFb);
-          remoteCapabilities.codecs.push(codec);
-        }
-      }
-      return remoteCapabilities;
-    };
-
-    // Serializes capabilities to SDP.
-    window.RTCPeerConnection.prototype._capabilitiesToSDP = function(caps) {
-      var sdp = '';
-      caps.codecs.forEach(function(codec) {
-        sdp += SDPUtils.writeRtpMap(codec);
-        sdp += SDPUtils.writeFtmp(codec);
-        sdp += SDPUtils.writeRtcpFb(codec);
-      });
-      return sdp;
     };
 
     // Calculates the intersection of local and remote capabilities.
@@ -1108,6 +1110,12 @@ if (typeof window === 'undefined' || !window.navigator) {
         var sendSsrc;
         var recvSsrc;
 
+        var remoteCapabilities = SDPUtils.parseRtpDescription(mediaSection);
+        var remoteIceParameters = SDPUtils.getIceParameters(mediaSection,
+            sessionpart);
+        var remoteDtlsParameters = SDPUtils.getDtlsParameters(mediaSection,
+            sessionpart);
+
         var mid = lines.filter(function(line) {
           return line.indexOf('a=mid:') === 0;
         })[0].substr(6);
@@ -1125,9 +1133,6 @@ if (typeof window === 'undefined' || !window.navigator) {
           recvSsrc = remoteSsrc.ssrc;
           cname = remoteSsrc.value;
         }
-
-        // determine remote caps from SDP
-        var remoteCapabilities = self._getRemoteCapabilities(mediaSection);
 
         if (description.type === 'offer') {
           var transports = self._createIceAndDtlsTransports(mid, sdpMLineIndex);
@@ -1172,11 +1177,6 @@ if (typeof window === 'undefined' || !window.navigator) {
           sendSsrc = transceiver.sendSsrc;
           //recvSsrc = transceiver.recvSsrc;
         }
-
-        var remoteIceParameters = SDPUtils.getIceParameters(mediaSection,
-            sessionpart);
-        var remoteDtlsParameters = SDPUtils.getDtlsParameters(mediaSection,
-            sessionpart);
 
         // for answers we start ice and dtls here, otherwise this is done in SLD
         if (description.type === 'answer') {
@@ -1461,6 +1461,9 @@ if (typeof window === 'undefined' || !window.navigator) {
         sdp += 'c=IN IP4 0.0.0.0\r\n';
         sdp += 'a=rtcp:9 IN IP4 0.0.0.0\r\n';
 
+        // Add a=rtpmap lines for each codec. Also fmtp and rtcp-fb.
+        sdp += SDPUtils.writeRtpDescription(localCapabilities);
+
         // Map ICE parameters (ufrag, pwd) to SDP.
         sdp += SDPUtils.writeIceParameters(
             transports.iceGatherer.getLocalParameters());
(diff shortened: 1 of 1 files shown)


calebboyd:
That looks pretty slick... Very nice

jan-ivar:
Sweet!

adapter.js
```diff
@@ -486,6 +488,1030 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+        iceServers: []
+      };
+      if (config && config.iceTransportPolicy) {
+        switch (config.iceTransportPolicy) {
+        case 'all':
+        case 'relay':
+          this.iceOptions.gatherPolicy = config.iceTransportPolicy;
+          break;
+        case 'none':
+          console.warn('can not map iceTransportPolicy none,' +
```
pthatcherg: That seems a little dangerous.  Why not just throw an exception?

adapter.js
```diff
@@ -486,6 +488,1030 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+              'falling back to "all"');
+          break;
+        }
+      }
+      if (config && config.iceServers) {
+        this.iceOptions.iceServers = config.iceServers;
+      }
+
+      // per-track iceGathers etc
+      this.tracks = [];
```
pthatcherg: That is a this a list of?  It seems like a list of (RtpSender, RtpReceiver), or in other words an m-line.  But the name "tracks" doesn't make that very clear.  A more full comment would be useful.

adapter.js
```diff
@@ -486,6 +488,1030 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+        priority: 1.0,
+        maxBitrate: 2000000.0,
+        minQuality: 0,
+        framerateBias: 0.5,
+        resolutionScale: 1.0,
+        framerateScale: 1.0,
+        active: true,
+        dependencyEncodingId: undefined,
+        encodingId: undefined
+      };
```
pthatcherg: These are almost all the defaults.  Are they just here for documentation, to make it easy to change them?

adapter.js
```diff
@@ -486,6 +488,1030 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+
+          var remoteDtlsParameters = self._getDtlsParameters(section,
+              sessionpart);
+          dtlsTransport.start(remoteDtlsParameters);
+
+          if (rtpSender) {
+            // calculate intersection of capabilities
+            var params = self._getCommonCapabilities(localCapabilities,
+                remoteCapabilities);
+            params.muxId = sendSSRC;
```
pthatcherg: muxId is supposed to be a string.  Is JS just magically converting here?

adapter.js
```diff
@@ -486,6 +488,1030 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+
+          var localCapabilities = RTCRtpReceiver.getCapabilities(kind);
+          // determine remote caps from SDP
+          remoteCapabilities = self._getRemoteCapabilities(section);
+
+          line = lines.filter(function(line) {
+            return line.indexOf('a=ssrc:') === 0 &&
+                line.split(' ')[1].indexOf('cname:') === 0;
+          });
+          sendSSRC = (2 * sdpMLineIndex + 2) * 1001;
```
pthatcherg: There's an interesting way to choose SSRCs.  Not sure it's really to spec, though.

adapter.js
```diff
@@ -486,6 +488,1030 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+        }
+      });
+
+      this.remoteDescription = description;
+      switch (description.type) {
+      case 'offer':
+        this._updateSignalingState('have-remote-offer');
+        break;
+      case 'answer':
+        this._updateSignalingState('stable');
```
pthatcherg: What, no pranswer :)?

adapter.js
```diff
@@ -486,6 +488,1030 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+    };
+
+    // Update the ICE connection state.
+    window.RTCPeerConnection.prototype._updateIceConnectionState =
+        function(newState) {
+      var self = this;
+      if (this.iceConnectionState !== newState) {
+        var agreement = self.tracks.every(function(track) {
+          return track.iceTransport.state === newState;
+        });
```
pthatcherg: I think the algorithm is a bit more complex than this, especially since we decided to make it 'connectionState' which includes both ICE and DTLS.

adapter.js
pthatcherg: Woh, sorry about the bad grammar there.  I meant "What is this a list of?".

adapter.js
```diff
@@ -486,6 +488,1030 @@ if (typeof window === 'undefined' || !window.navigator) {
 
   // the minimum version still supported by adapter.
   webrtcMinimumVersion = 12;
+
```
pthatcherg, line 496: Despite all my negative-sounding comments I left, I must say: this thing is great.  I'm surprised at how small it ended up being.  Well done.

fippo, line 496: I know how code-review with you folks works. In the end, it results in better code :-)

adapter.js
fippo: Edge has some restrictions about a minimum distance between ssrcs currently.

adapter.js
fippo: well, I can add that and a test (for https://github.com/webrtc/samples/issues/178) if you fix the chrome implementation :-)

adapter.js
fippo: yes, that code predates the connection state discussion. Need to update it, the algorithm you proposed makes more sense.

adapter.js
fippo: I pondered quite a bit here, m-line is such an sdp-ish term. But in the end, I am using sdpMLineIndex to index this list  so... will change.

adapter.js
fippo: mostly because I didn't know what was required what and not. Turns out that apart from ssrc and active most things are ignored for now. See blogs.windows.com/msedgedev/2015/09/18/ortc-api-is-now-available-in-microsoft-edge/

adapter.js
fippo: the demo i know (https://ortcdemo.azurewebsites.net/scripts/ortc-peer.js) is using an empty string. But I suspect that webidl magic happens here.

fippo:
see https://github.com/webrtc/samples/pull/646#issuecomment-142157106 for notes about making this a standalone file. I can do that when we're done here and decide the solution there is good.
That would require a major version bump since it removes global variables but solves the scoping problem once and for all. And adds a major feature :-)

adapter.js
```diff
@@ -486,6 +488,1032 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+          });
+          sendSSRC = (2 * sdpMLineIndex + 2) * 1001;
+          if (line) { // FIXME: alot of assumptions here
+            recvSSRC = line[0].split(' ')[0].split(':')[1];
+            cname = line[0].split(' ')[1].split(':')[1];
+          }
+          rtpReceiver = new RTCRtpReceiver(transports.dtlsTransport, kind);
+
+          // calculate intersection so no unknown caps get passed into the RTPReciver
+          params = self._getCommonCapabilities(localCapabilities,
```
foobarflu: note : this assumes that the intersection is not empty. i.e. it is prone to a similar error as https://code.google.com/p/webrtc/issues/detail?id=4957 -- this probably also means that calling rtpReceiver.receive _here_ might be wrong and this should only happen in SLD of the answer. But we need the track from the receiver so it needs to be created here...

Am I the only person who finds this confusing?

fippo: note : this assumes that the intersection is not empty. i.e. it is prone to a similar error as https://code.google.com/p/webrtc/issues/detail?id=4957 -- this probably also means that calling rtpReceiver.receive here might be wrong and this should only happen in SLD of the answer. But we need the track from the receiver so it needs to be created here...

Am I the only person (other than my sock puppet) who finds this confusing?

adapter.js
```diff
@@ -486,6 +488,1032 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+
+    window.RTCPeerConnection.prototype.removeStream = function(stream) {
+      var idx = this.localStreams.indexOf(stream);
+      if (idx > -1) {
+        this.localStreams.splice(idx, 1);
+      }
+      // FIXME: maybe trigger negotiationneeded?
+    };
+
+    // SDP helper from sdp-jingle-json with modifications.
```
alvestrand: Just wearing my amateur lawyer hat: Is this a copy from code you wrote yourself, or is it a copy from someone else's code?
If the latter, you should probably put in an ACK somewhere.
I'd also want either to have a complete link to sdp-jingle-json somewhere or to delete the comment, since others may ask the same question.

adapter.js
```diff
@@ -486,6 +488,1032 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+      };
+      var parseFmtp = function(line) {
+        var parsed = {};
+        var kv;
+        var parts = line.substr(('a=fmtp:' + mline[i]).length + 1).split(';');
+        for (var j = 0; j < parts.length; j++) {
+          kv = parts[j].split('=');
+          parsed[kv[0].trim()] = kv[1];
+        }
+        console.log('fmtp', mline[i], parsed);
```
alvestrand: This code has a total of 3 console.log statements, and the 2 others are commented out. Is this a leftover from debugging?

alvestrand:
My test engineer genes get terribly itchy on this one .... I wonder if it's reasonable to take the time to fake out an environment that looks like Edge to this code in node.js and run unittests on this code without a browser at all. There are some paths (like createOffer().then(setLocalDescription())) that should be easy to test that way.
But we could also just accept this PR and file an issue that such a test harness is needed...

adapter.js
fippo: https://github.com/otalk/sdp-jingle-json -- written by @legastero and me. Actually, if I can use modules I can require() full sdp-jingle-json (once its format has been updated) which is going to remove all the sdp parsing.

fippo:
@alvestrand the test situation is not that bad.
Using
`browserify test/test.js | testling --html > some.html`
one can generate a standalone version (currently uploaded [here](http://hancke.name/tmp/adaptertest.html)) and run it in Edge. One has to accept GUM a couple of times. Shall I update the "Run tests" section in the README.md with this information?
The tests currently fails in the iceTransportPolicy test, i'm looking into that. But that is one of the last tests so this is mostly looking good.

I would like to have more automated tests and am still looking into https://github.com/webrtc/samples/issues/592 :-)

adapter.js
fippo: moved to SLD in 29a14b71904a6679a4685b7d6cf96b1389aa8dc7

fippo:
other than merging https://github.com/fippo/adapter/tree/connectionstatechange-algorithm there is one major issue remaining: generating and process of end-of-candidates.
Edge requires an addIceCandidate with {} to start checking.
For the samples, I am relying on a candidate with a typ 'endOfCandidates' field and some bogus properties.

In simplewebrtc, I modified the signaling protocol, getting rid of the hack that looked for the ice-tcp candidate. Which made things work in Firefox but then Edge got confused about those invalid ice-tcp candidates.
I currently don't see a way to solve this without modifying the signaling :-/

adapter.js
```diff
@@ -484,8 +486,1049 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+          return line.indexOf('a=mid:') === 0;
+        })[0].substr(6);
+
+        var cname;
+
+        var remoteCapabilities;
+        var params;
+
+        if (description.type === 'offer') {
+          var transports = self._createIceAndDtlsTransports(mid, sdpMLineIndex);
```
fippo, line 1258: while the ice transport and dtls transport need to be created here (since the rtpreceiver needs to be created with a dtlsTransport, the dtlsTransport needs to be created with an iceTransport), at least the creation of the ice gatherer could be postponed until createAnswer. Which would remedy some of my confusion about not creating any state in createAnswer.

I suspect this is a rabbit hole wrt to ice restarts.

adapter.js
```diff
@@ -484,8 +486,1049 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+
+      parsed.name = parts[0];
+      parsed.clockRate = parseInt(parts[1], 10); // was: clockrate
+      parsed.numChannels = parts.length === 3 ? parseInt(parts[2], 10) : 1; // was: channels
+      return parsed;
+    };
+
+    // Parses SDP to determine capabilities.
+    window.RTCPeerConnection.prototype._getRemoteCapabilities =
+        function(section) {
```
fippo: section should be mLine probably. Should also be indicated in comment.

alvestrand: actually mediaSection is the better name. We've been dinged on using the term "m-line" both for the m= line itself and for the section it introduces.
Writing "This function parses an SDP media section to determine capabilities" in the comment would be good.

adapter.js
alvestrand: Would suggest an error case - switching on a string without a default is prone to interesting errors.

adapter.js
```diff
@@ -484,8 +486,1049 @@ if (typeof window === 'undefined' || !window.navigator) {
   webrtcDetectedVersion =
     parseInt(navigator.userAgent.match(/Edge\/(\d+).(\d+)$/)[2], 10);
 
-  // the minimum version still supported by adapter.
-  webrtcMinimumVersion = 12;
+  // The minimum version still supported by adapter.
+  // Build number since the major version is 12.
```
KaptenJansson: What does this mean in this context? Current Edge is version 12?

adapter.js
KaptenJansson: I agree, this applies to the [switch case](https://github.com/webrtc/adapter/pull/136/files#diff-07fae4471ea2a21eb6b4dd5eae5de84cR975) in the set local description as well.

adapter.js
```diff
@@ -484,8 +486,1049 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+
+      this.localDescription = description;
+      switch (description.type) {
+      case 'offer':
+        this._updateSignalingState('have-local-offer');
+        break;
+      case 'answer':
+        this._updateSignalingState('stable');
+        break;
+      }
```
KaptenJansson, line 1186: Maybe add a default/error case here in case the type is out of whack?

KaptenJansson:
Awesome work @fippo ;)

adapter.js
fippo: removed unused ones in 9a7fd07ecbd01d29a18473e5cfa70091f1837b82

adapter.js
fippo: throwing typeError (is that correct? done for the wrong iceTransportPolicy value) in b4db595d2b68e79ecd412aacbc2f4f57f6e4b5e0
(and missing a 'done' button)

adapter.js
fippo, line 1186: done in b4db595d2b68e79ecd412aacbc2f4f57f6e4b5e0

adapter.js
fippo: it's 13 for me (insider, 10547). The important part is the build number. This is described in http://blogs.windows.com/msedgedev/2015/09/21/understanding-versions-in-an-evergreen-browser/
Removed the wrong part of the comment in 7790eef41105d5bc3badf0c8ded37c29d807a0f1

adapter.js
fippo: Done in 335e63b2cf154e43265450008bcf289907fe68a1

Shall I also rename this.mLines to this.mediaSections? I don't think so since mediaSection here is the sdp term whereas the mLines are... something indexed by the sdpMLineIndex :-)

adapter.js
```diff
@@ -484,8 +486,1061 @@ if (typeof window === 'undefined' || !window.navigator) {
...
     parseInt(navigator.userAgent.match(/Edge\/(\d+).(\d+)$/)[2], 10);
 
-  // the minimum version still supported by adapter.
-  webrtcMinimumVersion = 12;
+  // The minimum version still supported by adapter.
+  // This is the build number for Edge.
+  webrtcMinimumVersion = 10547;
+
+  if (RTCIceGatherer) {
+    window.RTCIceCandidate = function(args) {
```
juberti: This seems like it should already exist in ORTC.

juberti: @pthatcherg says that it does have the same name.

adapter.js
```diff
@@ -484,8 +486,1061 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+      this.oniceconnectionstatechange = null;
+      this.onnegotiationneeded = null;
+      this.ondatachannel = null;
+
+      this.localStreams = [];
+      this.remoteStreams = [];
+      this.getLocalStreams = function() { return self.localStreams; };
+      this.getRemoteStreams = function() { return self.remoteStreams; };
+
+      this.localDescription = new RTCSessionDescription({
```
juberti: I think these are null initially?

adapter.js
```diff
@@ -484,8 +486,1061 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+      this.signalingState = 'stable';
+      this.iceConnectionState = 'new';
+
+      this.iceOptions = {
+        gatherPolicy: 'all',
+        iceServers: []
+      };
+      if (config && config.iceTransportPolicy) {
+        switch (config.iceTransportPolicy) {
+        case 'all':
```
juberti: google style is to indent |case|... but just match what we have in adapter

adapter.js
```diff
@@ -484,8 +486,1061 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+      // and be the same for all PCs from the same JS
+      this._cname = Math.random().toString(36).substr(2, 10);
+    };
+
+    window.RTCPeerConnection.prototype.addStream = function(stream) {
+      // clone just in case we're working in a local demo
+      // FIXME: seems to be fixed
+      this.localStreams.push(stream.clone());
+
+      // FIXME: maybe trigger negotiationneeded?
```
juberti: Yes, you probably want to call maybeFireNegotiationNeeded, which can run the spec algorithm internally.

adapter.js
```diff
@@ -484,8 +486,1061 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+      }
+      if (config && config.iceServers) {
+        this.iceOptions.iceServers = config.iceServers;
+      }
+
+      // per-track iceGathers, iceTransports, dtlsTransports, rtpSenders, ...
+      // everything that is needed to describe a SDP m-line.
+      this.mLines = [];
+
+      this._iceCandidates = [];
```
juberti: Could use a comment

adapter.js
```diff
@@ -484,8 +486,1061 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+      var idx = this.localStreams.indexOf(stream);
+      if (idx > -1) {
+        this.localStreams.splice(idx, 1);
+      }
+      // FIXME: maybe trigger negotiationneeded?
+    };
+
+    // SDP helper from sdp-jingle-json with modifications.
+    // https://github.com/otalk/sdp-jingle-json
+    window.RTCPeerConnection.prototype._toCandidateJSON = function(line) {
```
juberti: this could be better named parseCandidate

adapter.js
```diff
@@ -484,8 +486,1061 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+      }
+
+      var candidate = {
+        foundation: parts[0],
+        component: parts[1],
+        protocol: parts[2].toLowerCase(),
+        priority: parseInt(parts[3], 10),
+        ip: parts[4],
+        port: parseInt(parts[5], 10),
+        // skip parts[6] == 'typ'
```
juberti, line 550: for the ultimate in pedantry, you should do this in the key-value parsing below.

adapter.js
```diff
@@ -484,8 +486,1061 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+        type: parts[7]
+        //generation: '0'
+      };
+
+      for (var i = 8; i < parts.length; i += 2) {
+        if (parts[i] === 'raddr') {
```
juberti: switch/case?

adapter.js
```diff
@@ -484,8 +486,1061 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+          candidate.relatedAddress = parts[i + 1]; // was: relAddr
+        } else if (parts[i] === 'rport') {
+          candidate.relatedPort = parseInt(parts[i + 1], 10); // was: relPort
+        } else if (parts[i] === 'generation') {
```
juberti: @_@

adapter.js
```diff
@@ -484,8 +486,1061 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+          candidate.generation = parts[i + 1];
+        } else if (parts[i] === 'tcptype') {
+          candidate.tcpType = parts[i + 1];
+        }
+      }
+      return candidate;
+    };
+
+    // SDP helper from sdp-jingle-json with modifications.
+    window.RTCPeerConnection.prototype._toCandidateSDP = function(candidate) {
```
juberti: writeCandidate

adapter.js
```diff
@@ -484,8 +486,1061 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+      sdp.push(candidate.component);
+      sdp.push(candidate.protocol.toUpperCase());
+      sdp.push(candidate.priority);
+      sdp.push(candidate.ip);
+      sdp.push(candidate.port);
+
+      var type = candidate.type;
+      sdp.push('typ');
+      sdp.push(type);
+      if (type === 'srflx' || type === 'prflx' || type === 'relay') {
```
juberti: simpler: if (type !== 'host' && .relatedAddress && .relatedPort)

adapter.js
```diff
@@ -484,8 +486,1061 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+          sdp.push(candidate.relatedAddress); // was: relAddr
+          sdp.push('rport');
+          sdp.push(candidate.relatedPort); // was: relPort
+        }
+      }
+      if (candidate.tcpType && candidate.protocol.toUpperCase() === 'TCP') {
+        sdp.push('tcptype');
+        sdp.push(candidate.tcpType);
+      }
+      return 'a=candidate:' + sdp.join(' ');
```
juberti: don't use a=

adapter.js
```diff
@@ -484,8 +486,1061 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+      sdp.push(type);
+      if (type === 'srflx' || type === 'prflx' || type === 'relay') {
+        if (candidate.relatedAddress && candidate.relatedPort) {
+          sdp.push('raddr');
+          sdp.push(candidate.relatedAddress); // was: relAddr
+          sdp.push('rport');
+          sdp.push(candidate.relatedPort); // was: relPort
+        }
+      }
+      if (candidate.tcpType && candidate.protocol.toUpperCase() === 'TCP') {
```
juberti: suggest we use .toLowerCase and 'tcp', for consistency throughout this file

adapter.js
```diff
@@ -484,8 +486,1061 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+        sdp.push('tcptype');
+        sdp.push(candidate.tcpType);
+      }
+      return 'a=candidate:' + sdp.join(' ');
+    };
+
+    // SDP helper from sdp-jingle-json with modifications.
```
juberti: suggest adding sample input for each of these functions

adapter.js
```diff
@@ -484,8 +486,1061 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+        function(mediaSection) {
+      var remoteCapabilities = {
+        codecs: [],
+        headerExtensions: [],
+        fecMechanisms: []
+      };
+      var i;
+      var lines = mediaSection.split('\r\n');
+      var mline = lines[0].substr(2).split(' ');
+      var rtpmapFilter = function(line) {
```
juberti: This is going to need to be reworked. Perhaps write your own filter that takes args.

adapter.js
```diff
@@ -484,8 +486,1061 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+      parts = parts[0].split('/');
+
+      parsed.name = parts[0];
+      parsed.clockRate = parseInt(parts[1], 10); // was: clockrate
+      parsed.numChannels = parts.length === 3 ? parseInt(parts[2], 10) : 1; // was: channels
+      return parsed;
+    };
+
+    // This function parses an SDP media section to determine capabilities.
+    window.RTCPeerConnection.prototype._getRemoteCapabilities =
```
juberti: Peter suggests calling this parseMediaSection, and making it return a dictionary with media, ICE, and DTLS parameters.

adapter.js
```diff
@@ -484,8 +486,1061 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+          codec.rtcpFeedback = lines.filter(rtcpFbFilter).map(parseRtcpFb);
+
+          remoteCapabilities.codecs.push(codec);
+        }
+      }
+      return remoteCapabilities;
+    };
+
+    // Serializes capabilities to SDP.
+    window.RTCPeerConnection.prototype._capabilitiesToSDP = function(caps) {
```
juberti: This would be writeMediaSection, but take a dictionary with media, ICE, and DTLS parameters, resulting in a complete m= block.

adapter.js
```diff
@@ -484,8 +486,1061 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+      localCapabilities.codecs.forEach(function(lCodec) {
+        for (var i = 0; i < remoteCapabilities.codecs.length; i++) {
+          var rCodec = remoteCapabilities.codecs[i];
+          if (lCodec.name === rCodec.name &&
+              lCodec.clockRate === rCodec.clockRate &&
+              lCodec.numChannels === rCodec.numChannels) {
+            // push rCodec so we reply with offerer payload type
+            commonCapabilities.codecs.push(rCodec);
+
+            // FIXME: also need to calculate intersection between
```
juberti: sadly, this probably needs to have everything from webrtc's media(?)sessiondescriptionfactory.cc -> createAnswer

adapter.js
```diff
@@ -484,8 +486,1061 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+          }
+        }
+      });
+
+      // FIXME: fecMechanisms
+      return commonCapabilities;
+    };
+
+    // Parses DTLS parameters from SDP section or sessionpart.
+    window.RTCPeerConnection.prototype._getDtlsParameters =
```
juberti: combine into SDP uber-parsing function above

adapter.js
```diff
@@ -484,8 +486,1061 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+        fingerprints: [{
+          algorithm: fpLine.split(' ')[0],
+          value: fpLine.split(' ')[1]
+        }]
+      };
+      return dtlsParameters;
+    };
+
+    // Serializes DTLS parameters to SDP.
+    window.RTCPeerConnection.prototype._dtlsParametersToSDP =
```
juberti: combine into writeMediaSection

adapter.js
```diff
@@ -484,8 +486,1061 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+        function(params, setupType) {
+      var sdp = 'a=setup:' + setupType + '\r\n';
+      params.fingerprints.forEach(function(fp) {
+        sdp += 'a=fingerprint:' + fp.algorithm + ' ' + fp.value + '\r\n';
+      });
+      return sdp;
+    };
+
+    // Parses ICE information from SDP section or sessionpart.
+    window.RTCPeerConnection.prototype._getIceParameters =
```
juberti: combine into SDP parsing above

adapter.js
```diff
@@ -484,8 +486,1061 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+        })[0].substr(12),
+        password: lines.filter(function(line) {
+          return line.indexOf('a=ice-pwd:') === 0;
+        })[0].substr(10),
+      };
+      return iceParameters;
+    };
+
+    // Serializes ICE parameters to SDP.
+    window.RTCPeerConnection.prototype._iceParametersToSDP = function(params) {
```
juberti: same same

adapter.js
```diff
@@ -484,8 +486,1061 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+      return 'a=ice-ufrag:' + params.usernameFragment + '\r\n' +
+          'a=ice-pwd:' + params.password + '\r\n';
+    };
+
+    window.RTCPeerConnection.prototype._getEncodingParameters = function(ssrc) {
```
juberti: get->create

adapter.js
```diff
@@ -484,8 +486,1061 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+      return {
+        ssrc: ssrc,
+        codecPayloadType: 0,
```
juberti: Peter has many questions. Should work if just left as undefined. active should also be able to be undefined.

adapter.js
```diff
@@ -484,8 +486,1061 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+        function(mid, sdpMLineIndex) {
+      var self = this;
+      var iceGatherer = new RTCIceGatherer(self.iceOptions);
+      var iceTransport = new RTCIceTransport(iceGatherer);
+      iceGatherer.onlocalcandidate = function(evt) {
+        var event = {};
+        event.candidate = {sdpMid: mid, sdpMLineIndex: sdpMLineIndex};
+
+        var cand = evt.candidate;
+        var isEndOfCandidates = !(cand && Object.keys(cand).length > 0);
```
juberti: would be more obvious as !cand || length == 0

adapter.js
```diff
@@ -484,8 +486,1061 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+        if (isEndOfCandidates) {
+          event.candidate.candidate =
```
juberti: Need to comment what is going on here, and when we can remove it.

Makes sense to me now that you have explained it.

adapter.js
```diff
@@ -484,8 +486,1061 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+              'candidate:1 1 udp 1 0.0.0.0 9 typ endOfCandidates';
+          if (iceGatherer.state === undefined) { // polyfill
```
juberti: also would be good to have a note on when this can be removed, so that it doesn't exist for eternity

adapter.js
```diff
@@ -484,8 +486,1061 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+            iceGatherer.state = 'completed';
+          }
+        } else {
+          // RTCIceCandidate doesn't have a component, needs to be added
+          cand.component = iceTransport.component === 'RTCP' ? 2 : 1;
```
juberti, line 1055: try enum

adapter.js
```diff
@@ -484,8 +486,1061 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+          event.candidate.candidate = self._toCandidateSDP(cand);
+        }
+        if (self.onicecandidate !== null) {
+          // Emit null candidate when all gatherers are complete.
+          var complete = self.mLines.every(function(mLine) {
+            return mLine.iceGatherer && mLine.iceGatherer.state === 'completed';
+          });
+          if (self.localDescription && self.localDescription.type === '') {
```
juberti, line 1071: I think we need to find a way such that createOffer doesn't start gathering. createOffer shouldn't have side-effects.

adapter.js
```diff
@@ -484,8 +486,1061 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+          }
+        }
+      };
+      iceTransport.onicestatechange = function() {
+        self._updateIceConnectionState();
+      };
+
+      var dtlsTransport = new RTCDtlsTransport(iceTransport);
+      dtlsTransport.ondtlsstatechange = function() {
+        //self._updateConnectionState();
```
juberti: should be uncommented, obvs

adapter.js
```diff
@@ -484,8 +486,1061 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+        iceTransport: iceTransport,
+        dtlsTransport: dtlsTransport
+      };
+    };
+
+    window.RTCPeerConnection.prototype.setLocalDescription =
+        function(description) {
+      var self = this;
+      if (description.type === 'offer') {
+        if (!description.ortc) {
```
juberti: This should probably get refactored to have a table rather than pass attached to the session description like this.

adapter.js
```diff
@@ -484,8 +486,1061 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+          // FIXME: throw?
+        } else {
+          this.mLines = description.ortc;
+        }
+      } else if (description.type === 'answer') {
+        var sections = self.remoteDescription.sdp.split('\r\nm=');
```
juberti: I would be happier if we had an explicit splitSessionDescription function that returned an array of m= sections as text.

adapter.js
```diff
@@ -484,8 +486,1061 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+        var sessionpart = sections.shift();
+        sections.forEach(function(section, sdpMLineIndex) {
+          section = 'm=' + section;
+
+          var iceGatherer = self.mLines[sdpMLineIndex].iceGatherer;
```
juberti: let's have a local var to hold self.mLines[sdpMLineIndex]

adapter.js
```diff
@@ -484,8 +486,1061 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+          var iceTransport = self.mLines[sdpMLineIndex].iceTransport;
+          var dtlsTransport = self.mLines[sdpMLineIndex].dtlsTransport;
+          var rtpSender = self.mLines[sdpMLineIndex].rtpSender;
+          var rtpReceiver = self.mLines[sdpMLineIndex].rtpReceiver;
+          var localCapabilities =
+              self.mLines[sdpMLineIndex].localCapabilities;
+          var remoteCapabilities =
+              self.mLines[sdpMLineIndex].remoteCapabilities;
+          var sendSSRC = self.mLines[sdpMLineIndex].sendSSRC;
```
juberti: sendSsrc (here and elsewhere)

adapter.js
```diff
@@ -484,8 +486,1061 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+          iceTransport.start(iceGatherer, remoteIceParameters, 'controlled');
+
+          var remoteDtlsParameters = self._getDtlsParameters(section,
+              sessionpart);
+          dtlsTransport.start(remoteDtlsParameters);
+
+          // calculate intersection of capabilities
+          var params = self._getCommonCapabilities(localCapabilities,
+              remoteCapabilities);
+          if (rtpSender && params.codecs.length) {
```
juberti: This logic seems to be duplicated below and probably could be easily extracted into a function

adapter.js
```diff
@@ -484,8 +486,1061 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+          var localCapabilities =
+              self.mLines[sdpMLineIndex].localCapabilities;
+          var remoteCapabilities =
+              self.mLines[sdpMLineIndex].remoteCapabilities;
+          var sendSSRC = self.mLines[sdpMLineIndex].sendSSRC;
+          var recvSSRC = self.mLines[sdpMLineIndex].recvSSRC;
+
+          var remoteIceParameters = self._getIceParameters(section,
+              sessionpart);
+          iceTransport.start(iceGatherer, remoteIceParameters, 'controlled');
```
juberti: use enum for controlled?

adapter.js
```diff
@@ -484,8 +486,1061 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+          var params = self._getCommonCapabilities(localCapabilities,
+              remoteCapabilities);
+          if (rtpSender && params.codecs.length) {
+            params.muxId = sendSSRC;
+            params.encodings = [self._getEncodingParameters(sendSSRC)];
+            params.rtcp = {
+              cname: self._cname,
+              reducedSize: false,
+              ssrc: recvSSRC,
+              mux: true
```
juberti: check to see if this is still needed

adapter.js
```diff
@@ -484,8 +486,1061 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+      window.setTimeout(function() {
+        // FIXME: need to apply ice candidates in a way which is async but in-order
+        self._iceCandidates.forEach(function(event) {
+          if (self.onicecandidate !== null) {
+            self.onicecandidate(event);
+          }
+        });
+        self._iceCandidates = [];
+      }, 50);
+      if (arguments.length > 1 && typeof arguments[1] === 'function') {
```
juberti: This needs a comment to explain that this is explicitly firing the success callback

adapter.js
```diff
@@ -484,8 +486,1061 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+        this._updateSignalingState('stable');
+        break;
+      default:
+        throw new TypeError('unsupported type "' + description.type + '"');
+      }
+
+      // FIXME: need to _reliably_ execute after args[1] or promise
+      window.setTimeout(function() {
+        // FIXME: need to apply ice candidates in a way which is async but in-order
+        self._iceCandidates.forEach(function(event) {
```
juberti: Move this into some emitCandidates function that you can call either after calling the success callback, or after the promise is resolved (instead of using a gross timeout)

adapter.js
```diff
@@ -484,8 +486,1061 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+      if (mLine) {
+        var cand = Object.keys(candidate.candidate).length > 0 ?
+            this._toCandidateJSON(candidate.candidate) : {};
+        // dirty hack to make simplewebrtc work.
+        // FIXME: need another dirty hack to avoid adding candidates after this
+        if (cand.type === 'endOfCandidates') {
+          cand = {};
+        }
+        // dirty hack to make chrome work.
+        if (cand.protocol === 'tcp' && cand.port === 0) {
```
juberti, line 1604: Should just eat this candidate (as there won't ever be a browser-browser TCP connection).

adapter.js
```diff
@@ -484,8 +486,1061 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+        if (numVideoTracks > 0) {
+          tracks.push({
+            kind: 'video',
+            wantReceive: true
+          });
+          numVideoTracks--;
+        }
+      }
+
+      var sdp = 'v=0\r\n' +
```
juberti: a lot of this boilerplate looks like it could be shared between createOffer and createAnswer.

adapter.js
```diff
@@ -484,8 +486,1061 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+          'o=thisisadapterortc 8169639915646943137 2 IN IP4 127.0.0.1\r\n' +
+          's=-\r\n' +
+          't=0 0\r\n';
+      var mLines = [];
+      tracks.forEach(function(mline, sdpMLineIndex) {
+        // For each track, create an ice gatherer, ice transport, dtls transport,
+        // potentially rtpsender and rtpreceiver.
+        var track = mline.track;
+        var kind = mline.kind;
+        var mid = Math.random().toString(36).substr(2, 10);
```
juberti: function, please

adapter.js
```diff
@@ -484,8 +486,1061 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+    window.RTCPeerConnection.prototype._updateSignalingState =
+        function(newState) {
+      this.signalingState = newState;
+      if (this.onsignalingstatechange !== null) {
+        this.onsignalingstatechange();
+      }
+    };
+
+    // Update the ICE connection state.
+    window.RTCPeerConnection.prototype._updateIceConnectionState =
```
juberti: @pthatcherg to review with Taylor

adapter.js
```diff
@@ -484,8 +486,1061 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+      var numVideoTracks = 0;
+      // Default to sendrecv.
+      if (this.localStreams.length) {
+        numAudioTracks = this.localStreams[0].getAudioTracks().length;
+        numVideoTracks = this.localStreams[0].getAudioTracks().length;
+      }
+      // Determine number of audio and video tracks we need to send/recv.
+      if (offerOptions) {
+        // Deal with Chrome legacy constraints...
+        if (offerOptions.mandatory) {
```
juberti: Does anyone still use this? I would strongly prefer to avoid hacks atop hacks.

juberti:
Reviewed with @pthatcherg. Next step - apply the major comments regarding parse/write - and start adding tests, possibly ported from peerconnection_unittest.cc.

@pthatcherg has the task to take a few spec tweaks back to ORTC.

adapter.js
fippo: Edge doesn't have an object for it. And in the ORTC spec there is no constructor, unlike in the PC spec. I now remember noticing this (and in other places) and being happy with the current behaviour.

adapter.js
fippo: done

adapter.js
fippo: added an "does it already exist" check around both. Now I don't expect Edge to implement RTCSessionDescription but it would be fun to run this code in Chrome one day :-)

adapter.js
fippo: https://github.com/webrtc/apprtc/blob/master/src/web_app/js/peerconnectionclient.js#L61-L69
and the samples until https://github.com/webrtc/samples/pull/632/files last month.

Throwing a TypeError.

adapter.js
fippo: done.

adapter.js
fippo: done. also renamed to _localIceCandidatesBuffer.

adapter.js
fippo: three lines optional attribute instead of two. but... I gave up on the 999 goal long ago. done.

adapter.js
fippo, line 550: that would allow it to be elsewhere which is not allowed by the ABNF. Also, that would allow it not to be present at all.

adapter.js
fippo: done. As SDPUtils.parseCandidate for now to make it easier to require() SDPUtils later.

adapter.js
fippo: done.

adapter.js
fippo: done. ta

adapter.js
fippo: done

adapter.js
fippo: done

adapter.js
fippo: what about using https://gist.github.com/jed/982883 ? :-)
moved to a generateIdentifier until I understand that gist.
done.

adapter.js
fippo, line 1604: done.

adapter.js
fippo: done

adapter.js
fippo: done

adapter.js
fippo: done.
SDPUtils.splitSections returns an array containing the session part followed by the mediaSections -- one needs to shift the session part before iterating over the media sections.

adapter.js
fippo: done

adapter.js
fippo: done

adapter.js
fippo: done -- or rather: removed the function.

adapter.js
fippo: done

adapter.js
fippo: done. Also the timeout was easier to remove than I thought
update: not as easy. still looking into that :-/

adapter.js
fippo: done by adding a matchPrefix in SDPUtils (I think this is the third time i'm writing JS for this)

adapter.js
fippo: done. reducedSze and mux are not needed -- cname is set by JS so it can be be signaled in the a=ssrc line as there is no way to query the generated cname.

adapter.js
fippo, line 1055: no such enum. will investigate why

adapter.js
fippo: no such enum, as above

adapter.js
fippo: done

adapter.js
fippo: done

adapter.js
fippo: done -- reduced to just ssrc.

adapter.js
fippo: implemented per handwritten note. Still emitting iceconnectionstate, assuming we will shim the connectionState property + change event based on that.

adapter.js
fippo: done

adapter.js
fippo: for now I'm keeping a _pendingOffer and throwing an error if createOffer is called again without calling setLocalDescription.

Might return the identical offer and/or one with a new iceGatherer (to match chrome's behaviour)

fippo:
Phew. Thanks @juberti @pthatcherg, reviewing this in person was worth it for me.

Only one major point of refactoring remaining: SDPUtils.writeRtpDescription et al is repeated. There are slight differences which boil down mostly to the dtls role and the cname to use.
I am somewhat unhappy but not sure how to solve this.

For this, I'm in favor of squashing commits. Possibly by making a new PR that also makes adapter.js the result of browserify....
@alvestrand instead of bumping the major version I'd just go for webrtc@2.0.0 on npm -- webrtc-adapter-test is so... :-)

KaptenJansson:
> @alvestrand instead of bumping the major version I'd just go for webrtc@2.0.0 on npm -- webrtc-adapter-test is so... :-)

The webrtc NPM package points to a different repo: https://www.npmjs.com/package/webrtc. 
Also adapter.js is currently 0.2.3, do we need to jump to 2.0?

alvestrand:
@KaptenJansson there are webrtc, webrtc-adapter and webrtc-adapter-test on NPM. All 3 point to different repos and I have admin rights on all of them - whatever we overwrite, someone might get a nasty surprise.

fippo:
I think this is ready for another round @juberti @pthatcherg (everyone else: consider filtering github email or make sure are not watching :-))

Still tracking down a minor regression in the munge-sdp and stats samples.

fippo:
I fixed the getStats regression (calling getStats on an rtpreceiver without calling .receive() is a bad idea apparently). I think that indicates I am currently creating the receivers too early.

The munge regression is more tricky. It errors in the "Set Offer" step. Not sure why it worked before...

In all the other demos, attachMediaStream is (due to a zero timeout) called after the SLD.
In the munge demo, attachMediaStream is called in a state where .receive has not yet been called on the RTP receiver. I am hesitant to call receive at the answerer before SLD... feels wrong. But then, SRD at the answerer creates too much state?

alvestrand:
OK, I think we're coming up on decision time.

Things to consider:
- This code triples the size of adapter.js. I don't think this is a big enough problem to not merge.
- This code isn't tested by regression tests. This means that when we break it (and I'm sure we will), we won't detect the breakage until someone with Windows 10 and Edge runs the tests by hand.
- This code is a huge boost to the Edge/Chrome interoperability story, and we want it out there.

Suggested action:
- Integrate this.
- Bump the version number to 1.0.0 after integration, to indicate that there's a major change afoot.
- Add to the README a section saying that the Edge support is experimental, and anyone depending on Edge specifically is well advised to pin themselves to a specific version rather than going with "latest".
- Keep on looking for an open CI solution that supports Windows 10 and Edge...

Please comment!

KaptenJansson:
I agree with @alvestrand.

FYI I've been looking into CI running windows, I've tested appveyor and it works well however it does not support Windows 10 yet. If anyone has any suggestions for other CI's supporting Win 10, let me know.

fippo:
let's also break things by making adapter the result of browserify to as explained in https://github.com/webrtc/samples/pull/646#issuecomment-142157106
-- this also allows keeping _this_ code in a separate file. I'll make a new PR over the weekend.

adapter.js
```diff
@@ -484,8 +486,1119 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+            candidate.tcpType = parts[i + 1];
+            break;
+          default: // Unknown extensions are silently ignored.
+            break;
+        }
+      }
+      return candidate;
+    };
+
+    // Translates candidate object into SDP candidate attribute.
```
fippo: translates A candidate...

adapter.js
```diff
@@ -484,8 +486,1119 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+        sdp.push(candidate.relatedPort); // was: relPort
+      }
+      if (candidate.tcpType && candidate.protocol.toLowerCase() === 'tcp') {
+        sdp.push('tcptype');
+        sdp.push(candidate.tcpType);
+      }
+      return 'candidate:' + sdp.join(' ');
+    };
+
+    // Parses a rtpmap line, returns RTCRtpCoddecParameters. Sample input:
```
fippo: parses an

adapter.js
```diff
@@ -484,8 +486,1119 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+    SDPUtils.writeRtpMap = function(codec) {
+      var pt = codec.payloadType;
+      if (codec.preferredPayloadType !== undefined) {
+        pt = codec.preferredPayloadType;
+      }
+      return 'a=rtpmap:' + pt + ' ' + codec.name + '/' + codec.clockRate +
+          (codec.numChannels !== 1 ? '/' + codec.numChannels : '') + '\r\n';
+    };
+
+    // Parses a ftmp line, returns dictionary. Sample input:
```
fippo: an fmtp

adapter.js
```diff
@@ -484,8 +486,1119 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+        var params = [];
+        Object.keys(codec.parameters).forEach(function(param) {
+          params.push(param + '=' + codec.parameters[param]);
+        });
+        line += 'a=fmtp:' + pt + ' ' + params.join(';') + '\r\n';
+      }
+      return line;
+    };
+
+    // Parses a rtcp-fb line, returns RTCPRtcpFeedback object. Sample input:
```
fippo: same same

adapter.js
```diff
@@ -484,8 +486,1119 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+        // FIXME: special handling for trr-int?
+        codec.rtcpFeedback.forEach(function(fb) {
+          lines += 'a=rtcp-fb:' + pt + ' ' + fb.type + ' ' + fb.parameter +
+              '\r\n';
+        });
+      }
+      return lines;
+    };
+
+    // parses a RFC 5576 ssrc media attribute. Sample input:
```
fippo: Parses an..

adapter.js
```diff
@@ -484,8 +486,1119 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+    // a=ssrc:3735928559 cname:something
+    SDPUtils.parseSsrcMedia = function(line) {
+      var sp = line.indexOf(' ');
+      var parts = {
+        ssrc: line.substr(7, sp - 7),
```
fippo, line 687: make it more clear, no magic 7

adapter.js
```diff
@@ -484,8 +486,1119 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+
+    // Extracts DTLS parameters from SDP media section or sessionpart.
+    // FIXME: for consistency with other functions this should only
+    //   get the fingerprint line as input. See also getIceParameters.
+    SDPUtils.getDtlsParameters = function(mediaSection, sessionpart) {
+      var lines = SDPUtils.splitLines(mediaSection);
+      lines = lines.concat(SDPUtils.splitLines(sessionpart)); // Search in session part, too.
+      var fpLine = lines.filter(function(line) {
+        return line.indexOf('a=fingerprint:') === 0;
+      });
```
fippo: put [0] here

fippo: blow up more friendly, a sdp with all mlines rejected _might_ have no fingerprints

adapter.js
```diff
@@ -484,8 +486,1119 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+    // Parses ICE information from SDP media section or sessionpart.
+    // FIXME: for consistency with other functions this should only
+    //   get the ice-ufrag and ice-pwd lines as input.
+    SDPUtils.getIceParameters = function(mediaSection, sessionpart) {
+      var lines = SDPUtils.splitLines(mediaSection);
+      lines = lines.concat(SDPUtils.splitLines(sessionpart)); // Search in session part, too.
+      var iceParameters = {
+        usernameFragment: lines.filter(function(line) {
+          return line.indexOf('a=ice-ufrag:') === 0;
+        })[0].substr(12),
```
fippo, line 736: blow up nicely if there is no ufrag/pwd

adapter.js
```diff
@@ -484,8 +486,1119 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+      return iceParameters;
+    };
+
+    // Serializes ICE parameters to SDP.
+    SDPUtils.writeIceParameters = function(params) {
+      return 'a=ice-ufrag:' + params.usernameFragment + '\r\n' +
+          'a=ice-pwd:' + params.password + '\r\n';
+    };
+
+    // Parses the SDP media section and return RTCRtpParameters.
```
fippo: returns

fippo: parseRtpParameters

adapter.js
```diff
@@ -484,8 +486,1119 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+    SDPUtils.parseRtpDescription = function(mediaSection) {
+      var description = {
+        codecs: [],
+        headerExtensions: [],
+        fecMechanisms: [],
+        rtcp: []
+      };
+      var lines = SDPUtils.splitLines(mediaSection);
+      var mline = lines[0].substr(2).split(' ');
```
fippo: not accessing mline[0] so substr is not required

adapter.js
```diff
@@ -484,8 +486,1119 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+      for (var i = 3; i < mline.length; i++) { // find all codecs from mline[3..]
+        var line = SDPUtils.matchPrefix(
+            mediaSection, 'a=rtpmap:' + mline[i] + ' ')[0];
```
fippo: use var pt = mline[i]

adapter.js
```diff
@@ -484,8 +486,1119 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+      var description = {
+        codecs: [],
+        headerExtensions: [],
+        fecMechanisms: [],
+        rtcp: []
+      };
+      var lines = SDPUtils.splitLines(mediaSection);
+      var mline = lines[0].substr(2).split(' ');
+      for (var i = 3; i < mline.length; i++) { // find all codecs from mline[3..]
+        var line = SDPUtils.matchPrefix(
```
fippo: rtpmapline

adapter.js
```diff
@@ -484,8 +486,1119 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+            mediaSection, 'a=rtpmap:' + mline[i] + ' ')[0];
+        if (line) {
+          var codec = SDPUtils.parseRtpMap(line);
+          var fmtp = SDPUtils.matchPrefix(
```
fippo: fmtps

adapter.js
```diff
@@ -484,8 +486,1119 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+      }
+      // FIXME: parse headerExtensions, fecMechanisms and rtcp.
+      return description;
+    };
+
+    // Generates parts of the SDP media section describing the capabilities / parameters.
+    SDPUtils.writeRtpDescription = function(kind, caps) {
+      var sdp = '';
+
+      // Build the mline.
```
fippo, line 784: push to array, join instead of concat

adapter.js
```diff
@@ -484,8 +486,1119 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+      sdp += 'm=' + kind + ' ';
+      sdp += caps.codecs.length > 0 ? '9' : '0'; // reject if no codecs.
```
fippo, line 786: are there other reasons why you would reject?

adapter.js
```diff
@@ -484,8 +486,1119 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+      sdp += ' UDP/TLS/RTP/SAVPF ';
+      sdp += caps.codecs.map(function(codec) {
+        if (codec.preferredPayloadType !== undefined) {
+          return codec.preferredPayloadType;
+        }
+        return codec.payloadType;
+      }).join(' ') + '\r\n';
+
+      // FIXME: those should be IPv6 ::
```
fippo: supposed to be ipv4 now, double-check with spec

adapter.js
```diff
@@ -484,8 +486,1119 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+      sdp += 'a=rtcp:9 IN IP4 0.0.0.0\r\n';
+
+      // Add a=rtpmap lines for each codec. Also fmtp and rtcp-fb.
+      caps.codecs.forEach(function(codec) {
+        sdp += SDPUtils.writeRtpMap(codec);
+        sdp += SDPUtils.writeFtmp(codec);
+        sdp += SDPUtils.writeRtcpFb(codec);
+      });
+      // FIXME: add headerExtensions, fecMechanismş and rtcp.
+      sdp += 'a=rtcp-mux\r\n';
```
fippo, line 805: fixme: check while parsing

adapter.js
```diff
@@ -484,8 +486,1119 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+        sdp += 'a=sendonly\r\n';
+      } else if (transceiver.rtpReceiver) {
+        sdp += 'a=recvonly\r\n';
+      } else {
+        sdp += 'a=inactive\r\n';
+      }
+
+      if (transceiver.rtpSender) {
+        sdp += 'a=msid:' + stream.id + ' ' +
+            transceiver.rtpSender.track.id + '\r\n';
```
fippo: make a helper var to save a line

adapter.js
```diff
@@ -484,8 +486,1119 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+        sdp += 'a=ssrc:' + transceiver.sendSsrc + ' ' + 'msid:' +
```
fippo: FIXME: for rtx there might be multiple ssrc

adapter.js
```diff
@@ -484,8 +486,1119 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+      this.iceOptions = {
+        gatherPolicy: 'all',
+        iceServers: []
+      };
+      if (config && config.iceTransportPolicy) {
+        switch (config.iceTransportPolicy) {
+          case 'all':
+          case 'relay':
+            this.iceOptions.gatherPolicy = config.iceTransportPolicy;
+            break;
```
fippo, line 925: add nohost (or whatever its called)

adapter.js
```diff
@@ -484,8 +486,1119 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+      dtlsTransport.onerror = function() {
+        // onerror does not set state to failed by itself.
+        dtlsTransport.state = 'failed';
+        self._updateConnectionState();
+      };
+
+      return {
+        iceGatherer: iceGatherer,
+        iceTransport: iceTransport,
+        dtlsTransport: dtlsTransport
```
fippo: null/undefined sender/receiver

fippo, line 1101: check with latest spec updates about creating transceiver

adapter.js
```diff
@@ -484,8 +486,1119 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+    };
+
+    // Start the RTP Sender and Receiver for a transceiver.
+    window.RTCPeerConnection.prototype._transceive = function(transceiver,
+        send, recv) {
+      var params = this._getCommonCapabilities(transceiver.localCapabilities,
+          transceiver.remoteCapabilities);
+      if (send && transceiver.rtpSender) {
+        params.encodings = [{
+          ssrc: transceiver.sendSsrc
```
fippo, line 1112: fixme: multiple ssrcs (rtx)

adapter.js
```diff
@@ -484,8 +486,1119 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+          var remoteDtlsParameters = SDPUtils.getDtlsParameters(mediaSection,
+              sessionpart);
+          dtlsTransport.start(remoteDtlsParameters);
+
+          // Calculate intersection of capabilities.
+          var params = self._getCommonCapabilities(localCapabilities,
+              remoteCapabilities);
+          self._transceive(transceiver,
+              params.codecs.length > 0,
+              false);
```
fippo: receive happens elsewhere; comment where

adapter.js
```diff
@@ -484,8 +486,1119 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+          this._updateSignalingState('have-local-offer');
+          break;
+        case 'answer':
+          this._updateSignalingState('stable');
+          break;
+        default:
+          throw new TypeError('unsupported type "' + description.type + '"');
+      }
+
+      window.setTimeout(this._emitBufferedCandidates.bind(this), 50);
```
fippo: FIXME: find a better way (or add a comment)

adapter.js
```diff
@@ -484,8 +486,1119 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+      this.signalingState = newState;
+      if (this.onsignalingstatechange !== null) {
+        this.onsignalingstatechange();
+      }
+    };
+
+    // Determine whether to fire the negotiationneeded event.
+    window.RTCPeerConnection.prototype._maybeFireNegotiationNeeded =
+        function() {
+      // TODO
```
fippo: just fire it

adapter.js
fippo, line 925: nohost seems to be only in ORTC, not PC

KaptenJansson:
FYI just merge master in order to disable FF beta (or just change .travis.yml yourself in case of merge issues).

What's the next step here ? I've not followed to closely lately so I do not know the latest here.

adapter.js
fippo, line 784: Not sure if that would make things more clear actually.

adapter.js
fippo, line 736: I tried adding a check for rejected m-lines which should cover the reasonable cases where there is no ufrag/pwd or fingerprint. That however lead to the question what needs to be done with the transceiver in that case.

Also, error handling...

fippo:
it turned out that emitting the buffered candidates _after_ the resolve function is pretty hard since resolve() is executed async. This broke the promise-based test so i'm now calling emit with a 50ms delay for the promise-based version. :-/

adapter.js
```diff
@@ -488,8 +490,1143 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+            .filter(function(obj) {
+              return obj.attribute === 'cname';
+            })[0];
+        if (remoteSsrc) {
+          recvSsrc = parseInt(remoteSsrc.ssrc, 10);
+          cname = remoteSsrc.value;
+        }
+
+        if (description.type === 'offer') {
+          var transports = self._createIceAndDtlsTransports(mid, sdpMLineIndex);
```
pthatcherg, line 1258: I don't see that this parses and adds ICE candidates from the SDP anywhere.  Did you just forget to do that?

fippo, line 1258: no, non-trickle is one of the open issues -- noted in https://github.com/fippo/adapter/issues/2 ;-)

fippo, line 1258: i think the main issue I had here was the lack of end-of-candidates signaling in Chrome...  your issue 3772 ;-)

KaptenJansson:
Me and @fippo managed to get this to work in apprtc on Edge (currently served on https://10-dot-apprtc.appspot.com). Hence it's proof that this works. We found a couple of issues on Edge and in adapter but will fix the latter later after this has landed.

We will also after this has been merged create a way to generate adapter.js with and without ortc/edge support.

Issue #122, opened 2015-08-14
throws exception on chrome device emulator


yocontra:
Reproduce:
- Open chrome devtools
- Enable emulator
- Set device to "Apple iPhone 6"

Error:

```
Uncaught TypeError: Cannot read property '2' of null

/Users/chirp/node_modules/wildfire-client/node_modules/attachmediastream/node_modules/webrtc-adapter-test/adapter.js @ adapter.js:206

/Users/chirp/node_modules/wildfire-client/node_modules/getusermedia/node_modules/webrtc-adapter-test/adapter.js @ index-browser.js:80
```

Root of the problem: https://github.com/webrtc/adapter/blob/master/adapter.js#L232-L233

KaptenJansson:
Thanks for reporting.

This is due to assuming navigator.userAgent.match(/Chrom(e|ium)\/([0-9]+)./) will always return something when window.chrome exists.

KaptenJansson:
@calebboyd FYI

KaptenJansson:
Basically we do not cover the case where the UserAgent is Safari (or anything else than Chrome/Chromium) and window.chrome exists, e.g. Chrome on iOS.

calebboyd:
Perhaps the UA match should be used as the switch case. It would then most likely still break when emulating a device in a version of chrome that needed to be shimmed...

KaptenJansson:
@contra #142 has landed which should improve the version checking in general, could you give it a try again? Thanks!

KaptenJansson:
Ping?

yocontra:
Ended up using something else, can't comment on if this fixed it or not

calebboyd:
:hand: I'll confirm. This is fixed.

PR #142, opened 2015-10-06
make versioncheck more robust

### LINKED ISSUE #132: Doesn't recognize Opera for Android as a capable browser
I just reported this to Opera (bug DNA-43013), the problem is that `window.chrome` is not defined like in desktop version.

Don't know if we can do something here to be able to recognize this like user agent detection.

### LINKED ISSUE #135: Requests from Chrome WebView no longer detected as Chrome
5c5fff7b8972a0871a2f023fcc00c34eb2605834 changed the Chrome detection logic from:

``` javascript
else if (navigator.webkitGetUserMedia)
```

to:

``` javascript
else if (navigator.webkitGetUserMedia && !!window.chrome)
```

Unfortunately, window.chrome is not present for WebView. This is mentioned in [this Chromium bug report](https://code.google.com/p/chromium/issues/detail?id=527483#c19), but I think it's view restricted, so here's the relevant text:

> is it expected that !!window.chrome returns false for WebView?
> Checked webview m39 and trunk. Nope, window.chrome is not there for webview.

I'm hoping that reverting just that part of 5c5fff7b8972a0871a2f023fcc00c34eb2605834 is enough to make WebView requests be detected as Chrome again, but I'm not certain.

Instructions to build WebView, and in particular the WebView shell app, are available [here](https://www.chromium.org/developers/how-tos/android-build-instructions#TOC-Build-WebView-shell).

fippo:
This fixes #132 and #122
@tednakamura can you check http://hancke.name/tmp/opr.html shows a version instead of a blank page in webview? (it would then fix #135)

Not sure what the unsupported version should be... set it to 0.

### DIFF
--- a/adapter.js
+++ b/adapter.js
@@ -233,12 +233,10 @@ if (typeof window === 'undefined' || !window.navigator) {
   webrtcDetectedBrowser = 'chrome';
 
   // the detected chrome version.
-  webrtcDetectedVersion = navigator.userAgent.match(/Chrom(e|ium)\/([0-9]+)\./);
-  if (webrtcDetectedVersion && webrtcDetectedVersion.length >= 2) {
-    webrtcDetectedVersion = parseInt(webrtcDetectedVersion[2], 10);
-  } else {
-    webrtcDetectedVersion = 0;
-  }
+  webrtcDetectedVersion = (function() {
+    var match = navigator.userAgent.match(/Chrom(e|ium)\/([0-9]+)\./);
+    return match.length >= 2 && parseInt(match[2]);
+  })();
 
   // the minimum chrome version still supported by adapter.
   webrtcMinimumVersion = 38;
--- a/adapter.js
+++ b/adapter.js
@@ -227,14 +227,18 @@ if (typeof window === 'undefined' || !window.navigator) {
       });
     };
   }
-} else if (navigator.webkitGetUserMedia && !!window.chrome) {
+} else if (navigator.webkitGetUserMedia && window.webkitRTCPeerConnection) {
   webrtcUtils.log('This appears to be Chrome');
 
   webrtcDetectedBrowser = 'chrome';
 
   // the detected chrome version.
-  webrtcDetectedVersion =
-    parseInt(navigator.userAgent.match(/Chrom(e|ium)\/([0-9]+)\./)[2], 10);
+  webrtcDetectedVersion = navigator.userAgent.match(/Chrom(e|ium)\/([0-9]+)\./);
+  if (webrtcDetectedVersion && webrtcDetectedVersion.length >= 2) {
+    webrtcDetectedVersion = parseInt(webrtcDetectedVersion[2], 10);
+  } else {
+    webrtcDetectedVersion = 0;
+  }
 
   // the minimum chrome version still supported by adapter.
   webrtcMinimumVersion = 38;
--- a/adapter.js
+++ b/adapter.js
@@ -30,6 +30,10 @@ var webrtcUtils = {
       return;
     }
     console.log.apply(console, arguments);
+  },
+  extractVersion: function(uastring, expr, pos) {
+    var match = uastring.match(expr);
+    return match && match.length >= pos && parseInt(match[pos]);
   }
 };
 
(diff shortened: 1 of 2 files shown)


adapter.js
```diff
@@ -227,14 +227,18 @@ if (typeof window === 'undefined' || !window.navigator) {
       });
     };
   }
-} else if (navigator.webkitGetUserMedia && !!window.chrome) {
+} else if (navigator.webkitGetUserMedia && window.webkitRTCPeerConnection) {
   webrtcUtils.log('This appears to be Chrome');
```
alvestrand, line 235: Please continue checking window.chrome. There is at least one other Webkit-based browser out there with WebRTC functionality (Bowser).
(Note: opera has defined window.chrome, but does not define webkitRTCPeerConnection - it is instead defined on window.)

adapter.js
```diff
@@ -227,14 +227,18 @@ if (typeof window === 'undefined' || !window.navigator) {
...
 
   webrtcDetectedBrowser = 'chrome';
 
   // the detected chrome version.
-  webrtcDetectedVersion =
-    parseInt(navigator.userAgent.match(/Chrom(e|ium)\/([0-9]+)\./)[2], 10);
+  webrtcDetectedVersion = navigator.userAgent.match(/Chrom(e|ium)\/([0-9]+)\./);
```
alvestrand: Please use a different (local) temporary variable. Overwriting a string with an integer is just ugly.

alvestrand:
Can you write a test for this? (asking, not requesting - it would probably involve extending pretendVersion to take strings).

adapter.js
fippo: function scope hence global scope. The solution turned out not to be that bad... done.

I am looking forward to using browserify to generate adapter so we can declare all the variables we like :-)

adapter.js
fippo, line 235: but we want to work in a webview (see #135) where window.chrome is not defined.

Bowser does not use webkitGetUserMedia and webkitRTCPeerConnection hopefully. Checking for window.webkitRTCPeerConnection seems mandatory to me since we later fiddle with it.

fippo:
A test for webrtcUtils.extractionVersionFromUserAgent that throws some matching and non matching strings into it and makes sure things don't blow up like they did in the past?

tednakamura:
http://hancke.name/tmp/opr.html lgtm - see attached screenshot:
webview-good

And as a comparison, here's what happens when I visit https://webrtc.github.io/samples/src/content/getusermedia/gum/ with the same webview instance:
webview-bad

fippo:
@tednakamura: thanks, that is what I was hoping for.

testing turns out to be more complicated since
- navigator.userAgent can't be fiddled with
- I don't want add another global variable while we're in non-module module
- putting it onto webrtcUtils along with log() seemed ok but that is not exported to the module (which is good imo)

alvestrand:
@fippo would it work to put the parsing code (parseUserAgent?) into webrtcUtils, call webrtcUtils.parseUserAgent() from the tests, and call webrtcUtils.parseUserAgent(navigator.userAgent) in the main execution thread?

pavlospt:
Does anyone know when this will be merged?

fippo:
@alvestrand ugly but works. done.
Also made the two other version checks use this. The match position argument is somewhat odd since it relies on the user to know about the semantics of match but... that's safe to assume when passing a regexp before that argument.

fippo:
right, but that way leads to reinventing browser version detection modules -- which is why I'd rather keep this as a "safe regexp"

KaptenJansson:
Anything else needed here?

calebboyd:
It looks like the chrome specific patches will be applied in a shimmed safari environment (on iOS). I don't know if this would have any negative side effects but I should be able to test it soon.

fippo:
@calebboyd does that env have webkitRTCPeerConnection?

calebboyd:
Yeah, here is the specific shim I'm referring to [eface2face/iosrtc](https://github.com/eface2face/cordova-plugin-iosrtc/blob/master/docs/iosrtc.md#iosrtcregisterglobals)

I could also just delete it after it is written so it will fall through...

fippo:
i'd suggest that this code doesn't attempt to register chrome-specific prefixed objects (which webkitRTCPeerConnection is despite the name). Ideally adapter would check for !window.RTCPeerConnection also, but the code will need to be more modular for that.

alvestrand:
This part is done. I've merged it.

calebboyd:
Was just able to test this on iOS and it is no longer working. I'll see if I can come up with a patch...

For now I can just `delete window.webkitRTCPeerConnection` after `iosrtc.registerGlobals` is called

tgabi333:
What if you provide a way to force a behavior and skip detection?

leonklingele:
:+1:

PR #143, opened 2015-10-08
Remove trace function

KaptenJansson:
This mirrors what's done in https://github.com/webrtc/samples/pull/650.

@fippo @samdutton PTAL

### DIFF
--- a/adapter.js
+++ b/adapter.js
@@ -12,7 +12,7 @@
    quotmark: single, undef: true, unused: strict */
 /* global mozRTCIceCandidate, mozRTCPeerConnection, Promise,
 mozRTCSessionDescription, webkitRTCPeerConnection, MediaStreamTrack */
-/* exported trace,requestUserMedia */
+/* exported requestUserMedia */
 
 'use strict';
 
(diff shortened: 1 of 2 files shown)


samdutton:
LGTM.

Can't remember if we had discussion about this, but should we find a way for the samples to use adapter.js from the adapter repo — maybe even just by linking to https://raw.githubusercontent.com/webrtc/adapter/master/adapter.js rather than https://webrtc.github.io/samples/src/js/adapter.js?

Feels wrong to maintain two copies.

KaptenJansson:
We do merge the adapter.js version to [gh-pages](https://github.com/webrtc/adapter/tree/gh-pages) (for MS demos) hence it's available to use with a src tag if we want, can either pin it to a specific version or use the symlinked adapter-latest.js pointing to the latest version.

fippo:
i wonder if we should put this on hold until we do a major version bump because of ORTC anyway.

trace isn't exported but I wonder if anyone is using it from the 'raw js' version

KaptenJansson:
closing this as this will be removed as part of the 2.0 refactor in #164

PR #144, opened 2015-10-08
Initial refactoring of tests to use Webdriver

KaptenJansson:
First round of refactoring tests, I've not refactored all tests yet (they are commented out) but to make sure this does not get out of control (it's a pretty large change already) I will continue refactor and commit more to this PR after some reviews.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,7 @@
 browsers/
+firefox-*.tar.bz2
+.DS_Store
 node_modules/
+validation-report.json
+validation-status.json
 *~
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 sudo: false
 language: node_js
 node_js:
-- 0.12.7
+- 4.1
 
 env:
   matrix:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 sudo: false
 language: node_js
 node_js:
-- 0.10
+- 0.12.7
 
 env:
   matrix:
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ env:
     - BROWSER=chrome  BVER=unstable
     - BROWSER=firefox BVER=stable
     - BROWSER=firefox BVER=beta
-    - BROWSER=firefox BVER=unstable
+    - BROWSER=firefox BVER=nightly
     - BROWSER=firefox BVER=esr
 
 matrix:
@@ -18,7 +18,7 @@ matrix:
 
   allow_failures:
     - env: BROWSER=chrome  BVER=unstable
-    - env: BROWSER=firefox BVER=unstable
+    - env: BROWSER=firefox BVER=nightly
 
 before_script:
   - ./node_modules/travis-multirunner/setup.sh
--- a/adapter.js
+++ b/adapter.js
@@ -519,7 +519,8 @@ if (typeof module !== 'undefined') {
     webrtcDetectedBrowser: webrtcDetectedBrowser,
     webrtcDetectedVersion: webrtcDetectedVersion,
     webrtcMinimumVersion: webrtcMinimumVersion,
-    webrtcTesting: webrtcTesting
+    webrtcTesting: webrtcTesting,
+    webrtcUtils: webrtcUtils
     //requestUserMedia: not exposed on purpose.
     //trace: not exposed on purpose.
   };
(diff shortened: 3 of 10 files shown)


KaptenJansson:
This is interesting, Firefox beta and esr seem to be really slow at opening the fake device hence the local stream cannot be bound to window.stream in a reasonable amount of time.

fippo:
```
getTestPage
```

is interesting. Does that fetch the page once or every time?

KaptenJansson:
I've looked into it and it does not fetch the page everytime it seems, also setting variables to null at the end of each test does not bite either it seems. I've now tried to do `driver.navigate().refresh();` when `driver.getCurrentUrl !== ''` but I still see objects carried over to the next test.

KaptenJansson:
I do realize that getTestPage only calls driver.get() the first time and stores the value. I have to make a wrapper (I tried a conditional variable but that did not help either, i.e. it does not call upon the function/method every time it's accessed, only once....).

KaptenJansson:
Or maybe just scrap that and let each test call upon driver.get() directly.

fippo:
yeah, that was what I think you had intended :-)

KaptenJansson:
I was trying to make it easy to read but instead it will be rather complicated, keep it simple :)

KaptenJansson:
So what's your take on using the `window` object as a global namespace for webdriver to retrieve results/data?

fippo:
Can you point me to a line where you're using it? the diff is rather unhelpful, even hidden by github :-/

KaptenJansson:
Click on an individual commit, I've split it up into chunks just because of this.

fippo:
ah... things like the constraintsArray in https://github.com/KaptenJansson/adapter/blob/7dd8ae39cfd7fa069d9b7bbc2524209640e2cbea/test/test.js#L910 ?

https://github.com/KaptenJansson/adapter/blob/7dd8ae39cfd7fa069d9b7bbc2524209640e2cbea/test/test.js#L1123-L1126 looks like what I was trying to do for getStats initially. driver.executeAsyncScript might be a better approach, see https://github.com/webrtc/samples/blob/master/test/selenium-lib.js#L60-L64

fippo:
This is pretty rad actually. You are defining the test function in the node-javascript, e.g. [here](https://github.com/KaptenJansson/adapter/blob/7dd8ae39cfd7fa069d9b7bbc2524209640e2cbea/test/test.js#L25-L38). But the test is executed in the browser context.

I've tried that before but ended up with jshint yelling at me (probably you avoided this by using window. in several places), so I fell back to passing strings constructed in the node-js context.
Please document this pattern. Once you understand it, it is really cool but bad things can happen if you try to make changes without taking into account the restrictions imposed by this (e.g. you can't use binding)

KaptenJansson:
> Once you understand it, it is really cool but bad things can happen if you try to make changes without taking into account the restrictions imposed by this (e.g. you can't use binding)

AFAIK it just parses the defined function as a string, I did it this way because I could not stand writing code in strings (no syntax checking, linting etc).

I'm already hitting some restrictions, like getStats() reports in the node-js context does not inherit the full prototype chain, in this case [report.names()](https://github.com/webrtc/adapter/blob/master/test/test.js#L777) is returned as empty object rather than a function(method) when returned to webdriver and executed in the node.js context which means for the getStats tests I need execute all of  it in the browser context, but I can still verify it in the node.js context however.

KaptenJansson:
@alvestrand @samdutton @fippo @juberti PTAL when you get a chance. I want to do some more cleanups on some of the tests (after lessons learnt from later tests) but will do that after addressing your comment.

Note you have to click on individual commits to be able to get a proper diff due to the size of this PR.

fippo:
LGTM -- unifying the testing methods is certainly good.

I think https://github.com/KaptenJansson/adapter/blob/master/test/test.js#L100-L120 is better done with callbacks than using window for state. Let me try that... result is [here](https://github.com/KaptenJansson/adapter/pull/1)

nit: I wonder if we should replace the one-line executeScript calls with proper JS functions.

KaptenJansson:
@fippo is already part of this googlebot ;)

alvestrand:
@KaptenJansson one high level thing: Can you add some words to the README about how we do testing?

It might be as simple as saying "We use Selenium for testing. You can run tests with Grunt using the command ....., and they are also executed as continuous integration tests. You need ..... to be in place before you start testing."

KaptenJansson:
@alvestrand Absolutely! I will also add a description of how the tests are structured.

KaptenJansson:
FYI currently working on integrating @fippo's callback solution in all the tests, also trying to see if I can pass objects like that to webdriver but it seems like I'm hitting an inheritance problem when moving between contexts in a callback (between node.js where the test is defined and the browser where the test is executed).

KaptenJansson:
Any objections merging this?

alvestrand:
@fippo I think you need to say LGTM or something to keep googlebot happy.

samdutton:
No objections from me.

fippo:
Can I say LGTM for other reasons than keeping googlebot happy? ;-)

KaptenJansson:
Well the LGTM did not make it happy, manually overriding ;).

PR #146, opened 2015-10-17
make adapter.js the result of browserification

fippo:
(if browserification is a word... :-))

this adds a grunt task for browserify-ing. I think we don't want to include the resulting adapter.js in the main branch anymore, do we? I don't like committing autogenerated files, they make diffs hard to read.
But then, do we need to rename adapter.js to adapter-core or shall we keep adapter.js?

some issues remaining:
- [x] ensure this doesn't change the way the module is included (fairly sure but need to check)
- [x] fix some of the tests in the samples -- this makes variables like webrtcdetectedbrowser unavailable. Unless we want to explicitly make it a global variable again... or we make use of the [standalone](https://github.com/substack/browserify-handbook#standalone) option in browserify


fippo:
of course the conflict between browserify and grunt-browserify wasn't locally reproducible until I wiped node_modules and ~/.npm... squash it?

KaptenJansson:
Yes please squash. Also the documentation will need some updating as well.

WRT to the autogenerated adapter.js file, it would be nice if we could push it to gh-pages in a convenient and reliable way.

PR #152, opened 2015-11-12
Expose RTCIceCandidate and RTCSessionDescription

FelixMcFelix:
Very minor change mirroring the API of projects like [js-platform/node-webrtc](https://github.com/js-platform/node-webrtc), which explicity expose the above functions.

This change is very useful for developers targeting both client- and server-side deployment - presently the API forces users to check from their own modules whether or not ICE candidates and session descriptions are globally defined.

### DIFF
--- a/adapter.js
+++ b/adapter.js
@@ -517,6 +517,8 @@ if (typeof module !== 'undefined') {
   }
   module.exports = {
     RTCPeerConnection: RTCPeerConnection,
+    RTCIceCandidate: RTCIceCandidate,
+    RTCSessionDescription: RTCSessionDescription,
     getUserMedia: getUserMedia,
     attachMediaStream: attachMediaStream,
     reattachMediaStream: reattachMediaStream,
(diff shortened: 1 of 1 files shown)


KaptenJansson:
Thank you for your pull request!

@fippo, @alvestrand any objections to this?

@FelixMcFelix I've temporarily allowed FF beta to fail due to the download location is currently returning 404, you just need to merge master into your branch to fix it, no need to do this before @fippo and @alvestrand have answered  in case they have comments.

fippo:
no objections, but I don't see a big use-case either. 
I am not currently using importedAdapter.RTCSessionDescription anywhere.

FelixMcFelix:
@KaptenJansson Okay, I've merged master back into my branch. 3/4 of the Firefox downloads 404'd, which is odd, to say the least - any idea what's causing this? Cheers.

KaptenJansson:
Last week the Firefox download servers started returning 404 for some locations like latest-beta and latest-nightly but now it seems to affect stable as well. Will take care of that separately.

LGTM.

PR #154, opened 2015-11-17
add radix to parseInt

gasp:


KaptenJansson:
Hi,

Thanks for the PR!

Firefox download mechanism is broken due to Mozilla changing their download server, we can ignore the build error.

LGTM.

PR #158, opened 2015-12-15
add generateCertificate, fixes #41

### LINKED ISSUE #41: Need to support static methods on RTCPeerConnection

fippo:
@KaptenJansson @alvestrand
Not sure if I want to do the same for FF... maybe we should just drop support for versions < 38? Trivial to add though. cc @jan-ivar

### DIFF
--- a/adapter.js
+++ b/adapter.js
@@ -130,6 +130,20 @@ if (typeof window === 'undefined' || !window.navigator) {
     return new mozRTCPeerConnection(pcConfig, pcConstraints); // jscs:ignore requireCapitalizedConstructors
   };
 
+  // wrap static methods. Currently just generateCertificate.
+  if (mozRTCPeerConnection.generateCertificate) {
+    Object.defineProperty(RTCPeerConnection, 'generateCertificate', {
+      get: function() {
+        if (arguments.length) {
+          return mozRTCPeerConnection.generateCertificate.apply(null,
+              arguments);
+        } else {
+          return mozRTCPeerConnection.generateCertificate;
+        }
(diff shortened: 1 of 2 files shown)


KaptenJansson:
Nice work! ;)

Maybe we should look over all version checks? Not sure what versions we have committed to support, be aware of Firefox ESR though which lags behind for a while, I think it's still 38.

Also having a test for this would be nice, otherwise LGTM.

alvestrand:
That was quick :-)
Travis' hiccups seem to be a problem with fetching external content.
Should we add a function that throws a "notSupported" error when the function is not present?

fippo:
@alvestrand code that wants to use it will have to check if that is a function anyway (if anyone is using it at all)

KaptenJansson:
Filed #159 WRT to checking versions just to make sure we do not forget.

fippo:
Is this only available in M49 when setting --enable-experimental-web-platform-features?
Should we do that for beta and canary? I don't want it to be an assumption we make for stable.

KaptenJansson:
LGTM

PR #160, opened 2016-01-15
Add browser version check in the webdriver creation step

KaptenJansson:
This allows to set browser command line flags conditionally based on the browser version. This should address the concern in https://github.com/webrtc/adapter/pull/158#issuecomment-165861005.

I will fix the flaky video tag issue in another PR.

### DIFF
--- a/adapter.js
+++ b/adapter.js
@@ -1,5 +1,5 @@
 /*
- *  Copyright (c) 2014 The WebRTC project authors. All Rights Reserved.
+ *  Copyright (c) 2015 The WebRTC project authors. All Rights Reserved.
  *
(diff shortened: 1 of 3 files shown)


KaptenJansson:
@fippo is it sufficient to enable/disable flags based on version or should we do it per channel (stable, unstable etc) instead?

fippo:
Enabling this also in stable seems fine to me. It will only increase the number of issues we run into and therefore can catch.

LGTM -- I assume travis fails because of #161? Wondering why travis shows a push build.

KaptenJansson:
The reason I added this check is for compound flags like "--enable-experimental-web-platform-features" to ensure we know what we are testing.

PR #161, opened 2016-01-15
Add autoplay to video tags for all tests

KaptenJansson:
Firefox does not update video.videoHeight and video.videoWidth until the first frame.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -342,10 +342,10 @@ test('attachMediaStream', function(t) {
       video.setAttribute('id', 'video');
       video.setAttribute('autoplay', 'true');
       // If attachMediaStream works, we should get a video
-      // at some point. This will trigger onloadedmetadata.
+      // at some point. This will trigger onresize.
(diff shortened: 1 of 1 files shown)


fippo:
LGTM

Should we change the tests (in particular https://github.com/webrtc/adapter/blob/master/test/test.js#L387-L396) to be potentially-async and maybe listen for the video elements onresize?

fippo:
ah... now i understand why this kind-of worked. We are waiting for the element to appear which happens in loadedmetadata. Which apparently is called for videos which don't have autoplay. Maybe we should listen for onresize instead?

fippo:
cross-referencing -- maybe not playing the elements caused https://bugs.chromium.org/p/webrtc/issues/detail?id=4888 ?

KaptenJansson:
Changed to onresize and it seems to work well. Will do async tests in a separate PR if it fails or goes flaky.

PR #163, opened 2016-01-19
Use explicit reference to window.RTCPeerConnection for require context

xdumaine:
This fixes an error occurring in a require context: `"Object.defineProperty called on non-object"`. When used in a require context, I'm getting `RTCPeerConnection` as undefined here, since `this` is not `window`. Since Adapter explicitly is explicitly adding `RTCPeerConnection` to the window, use that object explicitly.

### DIFF
--- a/adapter.js
+++ b/adapter.js
@@ -134,7 +134,7 @@ if (typeof window === 'undefined' || !window.navigator) {
 
   // wrap static methods. Currently just generateCertificate.
   if (mozRTCPeerConnection.generateCertificate) {
-    Object.defineProperty(RTCPeerConnection, 'generateCertificate', {
+    Object.defineProperty(window.RTCPeerConnection, 'generateCertificate', {
       get: function() {
         if (arguments.length) {
           return mozRTCPeerConnection.generateCertificate.apply(null,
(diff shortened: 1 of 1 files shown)


fippo:
makes sense -- now why did it work the other way?

xdumaine:
I'm actually not sure... it's breaking in Canary and Firefox, but not in Chrome stable. Trying to figure that out now...

xdumaine:
Oh, of course, because `webkitRTCPeerConnection.generateCertificate` is not defined in stable chrome, but is in Canary (and `mozRTCPeerConnection.generateCertificate` in FF), so this line was not hit. I guess this was just never tested in a require context where `generateCertificate` was defined.

fippo:
generateCertificate is only there with the experimental flags. But this should still have been caught by the test.. unless the test now no longer tests require'd stuff :-(

xdumaine:
It's there in Firefox 42 without any flags. And appears to be in Chrome 50 without them as well.

KaptenJansson:
LGTM.

fippo:
that is published as 0.2.8?

KaptenJansson:
Yes it should be.

PR #164, opened 2016-01-20
Adapter 2.0 - Separate shims and utility functions into separate files

KaptenJansson:
Currently each browser shim and utility functions is separated into files, a main adapter_core.js uses node.js require() with browserify to bring them into one file (adapter.js) which will now be omitted from being checked in (we will still publish a browserified version on the gh-pages branch).

TODOs/questions:
- Logging tests are failing due to it now always being a module, not sure how we should suppress logging when used as a module (maybe have a toggle?).
- Test this on more browsers (webview, cordova, ios etc)
- Decide if we should only create global variable (window.adapter which is set in the browserifyOptions standalone option) for testing purposes (tests need to have access to this) or always include it (it exposes the browserShim object). We could just have different build rules for "prod" (without it) and "testing" (with it). Should also probably create a global variable for the tests where this object can be defined so that the we do not have to change the global object (adapter) name wherever it's called.
- Create build targets per browser or just for Edge (due to size of the shim and it's not widely used yet).

Feedback is welcome, this PR can be seen as a discussion thread with working code.
@alvestrand, @fippo


KaptenJansson:
Just to make it clear, I know the logging test is failing, reason is in the main description.

src/js/adapter_core.js
```diff
@@ -0,0 +1,62 @@
...
+ *
+ *  Use of this source code is governed by a BSD-style license
+ *  that can be found in the LICENSE file in the root of the source
+ *  tree.
+ */
+
+'use strict';
+
+// Shimming starts here.
+(function() {
```
fippo, line 11: why do we need the module pattern here if we use browserify anyway?

src/js/utils.js
```diff
@@ -0,0 +1,92 @@
...
+      result.version = this.extractVersion(navigator.userAgent,
+          /Edge\/(\d+).(\d+)$/, 2);
+      result.minVersion = 10547;
+      return result;
+    }
+  }
+};
+
+// Export.
+module.exports.utils = utils;
```
fippo: how about just exporting this as-is, i.e.

```
module.exports = utils;
```

You are always requiring this and then go for .utils anyway

src/js/chrome_shim.js
```diff
@@ -0,0 +1,307 @@
...
+ *  tree.
+ */
+'use strict';
+
+// Instantiate utility functions.
+var utils = function() {
+  var utils = require('./utils.js').utils;
+  var utilsObject = {
+    log: utils.log || console.log,
+    browser: utils.detectBrowser().browser,
```
fippo: meta-question: why does utils.js not run detectBrowser() and is just an object that can be accessed?

src/js/adapter_core.js
KaptenJansson, line 11: Might as well have it as a self executing function rather than assigning it and calling it separately? Also I like encapsulating everything even though we use browserify.

src/js/utils.js
KaptenJansson: Makes sense, done!

src/js/chrome_shim.js
KaptenJansson: I wanted it to be called explicitly when needed, but I see your point, the browser, version and minVersion are not likely to change during the same session.

alvestrand: I like having at least one scenario where detectBrowser not run automatically. It makes writing tests for detectBrowser much easier (if you want to fake and mock things before calling detectBrowser, and calling it twice may do something different from calling it once).

That's an advantage of splitting up the files: including adapter.js without a framework like npn SHOULD run detection automatically, while for tests, we can include the files that the tests need.

KaptenJansson:
Yes googlebot, @fippo is the author of these merged PR's ;)

fippo:
hi @googlebot, i think we've met before ;-)

KaptenJansson:
I tried to use browserify alias so that we only need to set the path for the files in the gruntfile rather than having a relative path in each file requiring a module but then the module.exports scope is all wrong due to alias does require on the files even though they are not even used in adapter_core.js. Took me a while to figure this out. Used alias initially to make it possible to have conditional require for edge, however I managed with ignore ;).

KaptenJansson:
Also, I created some more folders to make for a cleaner structure.

KaptenJansson:
Left todo's/questions:
- [x] Add logging toggle + fix logging test
- [ ] Agree upon this design, if the we want to separate even more (like edge_sd.js) bigger functions into separate modules we can always do that later.
- [ ] Change NPM name from webrtc-adapter-test to webrtc-adapter (this will result in the global object expose in the browser to have this name as well unless we override it)
- [ ] How does this work when it's included as a NPM module in a different project? Does it run the Gruntfile.js automagically ?
- [ ] merge latest changes from the master branch before merging.
- [ ] add tests for all build targets ensuring they deliver what's intended (edge shim should not be included in the no_edge target etc)

fippo:
(3): webrtc is still available ,-)
(4): IIUC you're making adapter-core.js the result of the build process and point main in package.json to it. main in package.json should point to a file that requires all the other dependencies.

Should we replace window. by global. which seems to be browserify BCP. Even though I don't see much advantage since this is not going to work in node anyway

KaptenJansson:
Got ya, will change back to adapter.js then.

I do not mind using either global or window, as you say, it will never run in a node context anyway. Maybe just leave it for now and change it later if desired?

KaptenJansson:
PTAL at the logging toggle, not sure if that's the best way....

KaptenJansson:
@calebboyd do you have any input here (considering your usage context)? 

Also a general question, how would developers expect to use this when included via NPM? Should browserified targets be built upon install and included in the node_modules folder or should it be as is (imported using node require and I guess requireJS (have not tested it))?

calebboyd:
@KaptenJansson thanks for pinging. After looking over it briefly It shouldn't present any new issues The pre-existing browser detection problems (`version === null`) still exist but it looks much more open to extension in that regard. 

For the general NPM question, I'd expect to download the built versions (prepublish) when installing the package. And that the built versions would _not_ be available in source control

KaptenJansson:
@calebboyd please check https://github.com/webrtc/adapter/commit/743786375a5a8a28a04437093bd26347b530a75f to see if that meets your expectations. 

Currently when doing npm publish in the adapter repo:
1. Builds all targets
2. Includes them in the published npm package
3. When added as dep in another project, it symlinks all adapter.js build artifacts from out/ to node_modules/.bin

calebboyd:
@KaptenJansson I think it does, except for 3. I think it should be `files` instead of `bin`. I'm pretty sure `bin` is for exposing a [hash](https://docs.npmjs.com/files/package.json#bin) of shell executable `command: script` pairs for `npm run` commands. Unless the symlinks have other benfits.

KaptenJansson:
Ow right, I just thought adding them to .bin would add them to the NPM path and you could then use NPM require for adapter.js or it's shims.

I'm using .npmignore hence I do not really need files (need to use it since I want to include the out/ folder in the npm package but not being part of the source)

KaptenJansson:
@fippo @jan-ivar @calebboyd @alvestrand @juberti I've now merged all the changes up until today. PTAL.

The only thing I'm not sure about at the moment is the NPM name and the version number (currently set to `webrtc-adapter` and `0.3.0`).

`webrtc` is owned by fippo@
`webrtc-adapter` is owned by willscott@ (last update in May 2015 and does not contain a whole lot of shimming)
`webrtc-adapter-test` is the current name but it's only a temporary one.

I suggest we reach out to willscott and ask if we can use webrtc-adapter. Just `webrtc` sounds to generic to me.

PS I do know that one can just publish on top of someone else package but that's IMHO a bit rude.

src/js/firefox/firefox_shim.js
```diff
@@ -0,0 +1,249 @@
...
+              return mozRTCPeerConnection.generateCertificate.apply(null,
+                  arguments);
+            } else {
+              return mozRTCPeerConnection.generateCertificate;
+            }
+          }
+        });
+      }
+
+      window.RTCSessionDescription = mozRTCSessionDescription;
```
juandebravo, line 113: you should do this assignment only if `window.RTCSessionDescription` is undefined, as [_mozRTCSessionDescription_ won't be maintained in future Firefox versions](https://wiki.mozilla.org/Media/WebRTC/ReleaseNotes/44#Noteworthy_Changes):

``` js
if (!window.RTCSessionDescription) {
    window.RTCSessionDescription = mozRTCSessionDescription;
}
```

src/js/firefox/firefox_shim.js
```diff
@@ -0,0 +1,249 @@
...
+      window.RTCIceCandidate = mozRTCIceCandidate;
```
juandebravo, line 114: same here:

``` js
if (!window.RTCIceCandidate) {
    window.RTCIceCandidate = mozRTCIceCandidate;
}
```

src/js/chrome/chrome_shim.js
```diff
@@ -0,0 +1,345 @@
...
+                      deviceId: device.id,
+                      groupId: ''};
+            }));
+          });
+        });
+      }};
+    }
+
+    // A shim for getUserMedia method on the mediaDevices object.
+    // TODO(KaptenJansson) remove once implemented in Chrome stable.
```
juandebravo, line 279: <del>*Chrome stable* -> *Firefox stable*?</del>

src/js/utils.js
```diff
@@ -0,0 +1,99 @@
...
+        'adapter.js logging enabled';
+  },
+
+  log: function() {
+    if (typeof window === 'object') {
+      if (logDisabled_) {
+        return;
+      }
+      console.log.apply(console, arguments);
+    }
```
juandebravo, line 30: a bit simpler:

``` js
if (typeof window === 'object' && !logDisabled_) {
    console.log.apply(console, arguments);
}
```

src/js/utils.js
```diff
@@ -0,0 +1,99 @@
...
+    result.minVersion = null;
+
+    // Non supported browser.
+    if (typeof window === 'undefined' || !window.navigator) {
+      result.browser = 'Not a supported browser.';
+      return result;
+    }
+
+    // Firefox.
+    if (navigator.mozGetUserMedia) {
```
juandebravo, line 64: this will break in two-three releases after Firefox 44 ([Firefox will stop supporting the `mozGetUserMedia`](https://wiki.mozilla.org/Media/WebRTC/ReleaseNotes/44#Noteworthy_Changes:) object)

src/js/utils.js
```diff
@@ -0,0 +1,99 @@
...
+    // Edge.
+    if (navigator.mediaDevices &&
+        navigator.userAgent.match(/Edge\/(\d+).(\d+)$/)) {
+      result.browser = 'edge';
+      result.version = this.extractVersion(navigator.userAgent,
+          /Edge\/(\d+).(\d+)$/, 2);
+      result.minVersion = 10547;
+      return result;
+    }
+  }
```
juandebravo, line 90: return a `non supported browser` at the bottom of this method (if the user agent is Safari, Opera, ...)

src/js/utils.js
jan-ivar, line 64: Note that `mozGetUserMedia` is not on that list, and should be safe for now. The asymmetry stems from `RTCPeerConnection` (a class) supporting both promises and callbacks, whereas `navigator.mediaDevices.getUserMedia` (a function) is promises only. `navigator.mozGetUserMedia` is the last remaining callback API, and AFAIK there are no plans to unprefix it.

src/js/firefox/firefox_shim.js
fippo, line 113: all three of those were unprefixed in the same Firefox version so !window.RTCPeerConnection implies !window.RTCSessionDescription

src/js/utils.js
juandebravo, line 64: umm you're right @jan-ivar, I included the wrong information :blush: .

I saw in several places the warning `navigator.mozGetUserMedia has been replaced by navigator.mediaDevices.getUserMedia` (i.e. https://github.com/amw/jpeg_camera/issues/6, https://github.com/mozilla/fxa-content-server/issues/3150).

_[NavigatorGetUserMediaWarning=navigator.mozGetUserMedia has been replaced by navigator.mediaDevices.getUserMedia](https://hg.mozilla.org/releases/mozilla-b2g44_v2_5/rev/278d5ae8efb7)_

src/js/firefox/firefox_shim.js
juandebravo, line 113: :+1:

src/js/utils.js
fippo, line 64: that warning only pops up when using it -- not when checking for existence ;-)

juandebravo, line 64: oh good to now :) and thanks for the quick answer!!

src/js/utils.js
fippo, line 90: Opera is considered Chrome. This should have been called webkit but the cost of changing this is rather high.

Not returning anything for unsupported browsers (the first check is not checking for a supported browser, it is checking whether this is a browser) is questionable. Maybe change the the first error to 'Not a browser' and return 'Not a supported browser'?

juandebravo, line 90: agree on returning `Not a browser` at the first check and `Not a supported browser` at the bottom of the method

src/js/chrome/chrome_shim.js
KaptenJansson, line 279: ?

src/js/utils.js
KaptenJansson, line 90: Thanks for your comments.

Keep in mind this is just a util method detecting what browser it is, the actual decision is in adapter-core.js (switch 'chrome/edge/firefox').

Not sure what we gain on adding an additional check due to if the aforementioned browsers are not found, adapter-core.js will hit default in the switch statement anyhow. 

This logic is basically copied over from the old adapter.js. I suggest you open a new bug or submit a PR where we can discuss this further. Should not block this PR.

src/js/firefox/firefox_shim.js
KaptenJansson, line 113: These are copied from the old adapter.js, should not be dealt with there (if at all). Should rather care about the implementation details of splitting into files.

src/js/firefox/firefox_shim.js
KaptenJansson, line 114: These are copied from the old adapter.js, should not be dealt with there (if at all). Should rather care about the implementation details of splitting into files. Please file new issues/create PR's.

KaptenJansson:
Flipping the label due to Googlebot is not able to do so after the fact.

KaptenJansson:
I'm going to merge this now since I've not gotten any objections and want to get this before new PR's line up.

PR #166, opened 2016-01-20
move sdp utils to separate module

fippo:
ontop of #164


KaptenJansson:
I was thinking about this but noticed the PeerConnection shim is using it lots to hence I figured you cannot not use the Peerconnection shim without the SDP shim hence they could be in the same file. But this works too, makes it cleaner.

Do you want your PR's to be merged directly on top the of #164 PR?

fippo:
we _might_ start fixing SDP bugs in other browsers, too. https://bugs.chromium.org/p/webrtc/issues/detail?id=4957 I am looking at you ;-)

KaptenJansson:
;)

PR #167, opened 2016-01-20
srcObject is always there in edge

fippo:
shall we remove the shimSourceObject function and the call in core?


KaptenJansson:
Makes sense, copy paste... ;)

PR #169, opened 2016-01-25
deal with string gum constraints

fippo:
```
var vgaConstraints = {
  video: {width: {exact: "640"}, height: {exact: "480"}}
};

navigator.mediaDevices.getUserMedia(vgaConstraints).then(function(stream) { console.log('yay'); }).catch(function(err) { console.log(err); })
```

Both Firefox and Edge deal with this, but the shim does not. I think this should cover the relevant cases. wdyt @jan-ivar?

### DIFF
--- a/adapter.js
+++ b/adapter.js
@@ -383,8 +383,8 @@ if (typeof window === 'undefined' || !window.navigator) {
         return;
       }
       var r = (typeof c[key] === 'object') ? c[key] : {ideal: c[key]};
-      if (r.exact !== undefined && typeof r.exact === 'number') {
-        r.min = r.max = r.exact;
+      if (r.exact !== undefined) {
+        r.min = r.max = r.exact = parseInt(r.exact, 10);
       }
       var oldname = function(prefix, name) {
         if (prefix) {


fippo:
travis failure seems to be related to bad webdriver version?

adapter.js
```diff
@@ -383,8 +383,8 @@ if (typeof window === 'undefined' || !window.navigator) {
         return;
       }
       var r = (typeof c[key] === 'object') ? c[key] : {ideal: c[key]};
-      if (r.exact !== undefined && typeof r.exact === 'number') {
-        r.min = r.max = r.exact;
```
jan-ivar, line 387: Oh I remember why I did this. It's because not all constraints are numeric. For example:

```
deviceId: {exact: "20983-20o198-109283-098-09812"},
facingMode: {exact: "environment"},
```

And these don't have `min` and `max` in the spec.

adapter.js
```diff
@@ -383,8 +383,8 @@ if (typeof window === 'undefined' || !window.navigator) {
...
+      if (r.exact !== undefined) {
+        r.min = r.max = r.exact = parseInt(r.exact, 10);
```
jan-ivar, line 387: Here I worry we'll end up with minDeviceId, maxDeviceId, minFacingMode and maxFacingMode below, which is gibberish of course.

jan-ivar, line 387: Or rather, I would worry had you changed the two other number typeof's below (lines 398 and 409).

KaptenJansson:
LGTM, anything else needed here?

jan-ivar:
Do we care about `{ video: { width: "640", height: "480" } }`? Because I don't think that's working.

fippo:
given that this is more complicated than it looks (thanks @jan-ivar) do we want to bother before the native constraints roll out?

jan-ivar:
Also, wont this convert `{ video: { facingMode: { exact: "environment" } }` to `{ video: { mandatory: { facingMode: 0 } }` ? That's wrong.

jan-ivar:
Is anyone really passing strings here?

fippo:
@chadwallacehart did and asked me when it did not work :-)

fippo:
ah... lets not do this. bigger harm than benefits, also since iiuc the new constraints will be available natively soon

Issue #171, opened 2016-01-28
Add shim for ontrack instead of onaddstream


a2sheppy:
With onaddstream deprecated in favor of using ontrack, it would be helpful to have a shim to allow use of ontrack where it's not yet implemented.

KaptenJansson:
Fixed in #175

PR #175, opened 2016-01-29
Polyfill ontrack on browsers that don't have it.

jan-ivar:
Firefox 47 has this natively, so time for a polyfill.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -1261,6 +1261,7 @@ test('Basic connection establishment', function(t) {
     pc2.ontrack = function(e) {
       window.testPassed.push('pc2.ontrack');
       window.trackEvent = e;
+      window.tracks = e.streams.length ? e.streams[0].getTracks() : [];
       window.receivers = pc2.getReceivers();
     };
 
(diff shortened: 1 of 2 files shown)


jan-ivar:
> ^ 'arrow function syntax (=>)' is only available in ES6 (use esnext option).

@KaptenJansson is using exnext an option by any chance?

fippo:
@a2sheppy pushed you, eh? Looking forward to using this. I should natively polyfill this for Edge though, onaddstream is a hack there.

jan-ivar:
Fixes https://github.com/webrtc/adapter/issues/171

fippo:
@samdutton maybe you know some people with strong opinions re esnext?

jan-ivar:
PTAL

test/test.js
```diff
@@ -1225,6 +1225,14 @@ test('Basic connection establishment', function(t) {
...
     window.testPassed = [];
     window.testFailed = [];
+    var t = {
+      ok: function(ok, msg) {
+        window[ok ? 'testPassed' : 'testFailed'].push(msg);
+      },
+      is: function(a, b, msg) {
+        this.ok((a === b), msg + ' - got ' + b);
+      }
+    };
```
fippo, line 1235: wait, this makes life easy again?!

fippo:
i'm very much in favor of merging this!

test/test.js
jan-ivar, line 1235: Look below.

samdutton:
> samdutton https://github.com/samdutton maybe you know some people with
> strong opinions re esnext?

I think it's fine to use any syntax that's supported by the browsers we're
targeting, if that's what you mean :).

KaptenJansson:
Re esnext, is that message from nodejs? I agree with samdutton@ as long as it supported by browsers and nodejs it's fine.

KaptenJansson:
LGTM.

fippo:
no, a jshint option. But why are we still using jshint/jscs even?

samdutton:
We're using ESLint for samples — just created new issue to do the same for
adapter.js: #176 https://github.com/webrtc/adapter/issues/176.

KaptenJansson:
Yeah I was also wondering why we weren't using eslint everywhere, I think we wanted to see how it worked out in samples first.

PR #178, opened 2016-02-04
use tape-style assertions inside selenium

fippo:
'cause @jan-ivar showed how. Still looking for a way to not duplicate the definition of t in the browser environment...

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -1291,30 +1291,30 @@ test('Basic connection establishment', function(t) {
 
       pc1.createOffer(
         function(offer) {
-          t.ok('pc1.createOffer');
+          t.pass('pc1.createOffer');
           pc1.setLocalDescription(offer,
             function() {
-              t.ok('pc1.setLocalDescription');
+              t.pass('pc1.setLocalDescription');
 
               offer = new RTCSessionDescription(offer);
-              t.ok('created RTCSessionDescription from offer');
+              t.pass('created RTCSessionDescription from offer');
               pc2.setRemoteDescription(offer,
                 function() {
-                  t.ok('pc2.setRemoteDescription');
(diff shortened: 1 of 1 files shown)


test/test.js
```diff
@@ -1231,7 +1231,13 @@ test('Basic connection establishment', function(t) {
       },
       is: function(a, b, msg) {
         this.ok((a === b), msg + ' - got ' + b);
-      }
+      },
+      pass: function(msg) {
+        window.testPassed.push(msg);
```
jan-ivar: or `this.ok(true, msg);`

test/test.js
```diff
@@ -1231,7 +1231,13 @@ test('Basic connection establishment', function(t) {
...
+      },
+      fail: function(msg) {
+        window.testFailed.push(msg);
```
jan-ivar: or `this.ok(false, msg);`

test/test.js
fippo: even better!

test/test.js
```diff
@@ -1285,64 +1291,56 @@ test('Basic connection establishment', function(t) {
 
       pc1.createOffer(
         function(offer) {
-          window.testPassed.push('pc1.createOffer');
+          t.ok('pc1.createOffer');
```
jan-ivar: I think you mean `t.pass` here (and below).

fippo:
note: it would be good to also "shim" t.plan here which might have helped catching #179. Not sure what the best way to do it is though

KaptenJansson:
LGTM.

Not sure you can use t.plan() here anyway due to having t.pass on the onicecandidate event (we do not know how many candidates will be generated), this could ofc be changed to only trigger once which would make t.plan() usable. But I think it's fine as it is for now.

PR #179, opened 2016-02-04
set prototype and fix ontrack

fippo:
had to do that to make #175 work. This should make the 'trackEvent.track is an object' messages appear in travis

### DIFF
--- a/adapter.js
+++ b/adapter.js
@@ -1683,12 +1683,24 @@ if (typeof window === 'object' && !('ontrack' in
   Object.defineProperty(window.RTCPeerConnection.prototype, 'ontrack', {
     get: function() { return this._ontrack; },
     set: function(f) {
+      var self = this;
       if (this._ontrack) {
         this.removeEventListener('track', this._ontrack);
         this.removeEventListener('addstream', this._ontrackpoly);
       }
       this.addEventListener('track', this._ontrack = f);
       this.addEventListener('addstream', this._ontrackpoly = function(e) {
+        if (webrtcDetectedBrowser === 'chrome') {
+          // onaddstream does not fire when a track is added to an existing stream.
+          // but stream.onaddtrack is implemented so we use thたt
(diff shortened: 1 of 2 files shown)


fippo:
also fixed the receivers-contain test which failed in Firefox after some discussion with @jan-ivar

adapter.js
```diff
@@ -1676,7 +1678,8 @@ if (typeof window === 'undefined' || !window.navigator) {
 }
 
 // Polyfill ontrack on browsers that don't yet have it
-if (typeof window === 'object' && !('ontrack' in window.RTCPeerConnection)) {
+if (typeof window === 'object' && !('ontrack' in
+    window.RTCPeerConnection.prototype)) {
```
jan-ivar, line 1682: Should probably have a check in the middle here that window.RTCPeerConnection exists, just to be on the safe side?

jan-ivar:
Thanks for this!

adapter.js
KaptenJansson, line 1682: I agree.

KaptenJansson:
LGTM apart from the comment from @jan-ivar.

adapter.js
fippo, line 1682: had done that earlier but pushed in the wrong branch.

fippo:
Before merging I want to fix that we currently require .ontrack to be set -- we should also support addEventListener('track', ...)

jan-ivar:
@fippo can we do that in a separate PR? It's broken right now...

fippo:
hrm. I would need to add a variable so... i'll do that in the modularized stuff.

KaptenJansson:
Ok, so this is OK to merge then? The follow up PR will be done in after #164 ? I will try to land it this week.

fippo:
can you bump npm and update gh-pages please?

KaptenJansson:
Done

PR #182, opened 2016-02-08
shim addTrack

fippo:
Note that I can't return an RTPSender.

also this recreates the blob url whenever a track is added to a stream attached to a video element.
This causes a slight visual glitch but... we might have native srcObject soon.

Should that also remove the event listeners when the srcObject is changed?

### DIFF
--- a/adapter.js
+++ b/adapter.js
@@ -1701,7 +1701,7 @@ if (typeof window === 'object' && window.RTCPeerConnection && !('ontrack' in
       this.addEventListener('addstream', this._ontrackpoly = function(e) {
         if (webrtcDetectedBrowser === 'chrome') {
           // onaddstream does not fire when a track is added to an existing stream.
-          // but stream.onaddtrack is implemented so we use thたt
+          // but stream.onaddtrack is implemented so we use that
           e.stream.addEventListener('addtrack', function(te) {
             var event = new Event('track');
             event.track = te.track;
--- a/adapter.js
+++ b/adapter.js
@@ -1724,7 +1724,7 @@ if (typeof window === 'object' && window.RTCPeerConnection && !('ontrack' in
 
 if (typeof window === 'object' && window.RTCPeerConnection && !('addTrack' in
     window.RTCPeerConnection.prototype)) {
-  RTCPeerConnection.prototype.addTrack = function(track, stream) {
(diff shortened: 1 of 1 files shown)


adapter.js
```diff
@@ -1713,6 +1722,18 @@ if (typeof window === 'object' && window.RTCPeerConnection && !('ontrack' in
...
 }
 
+if (typeof window === 'object' && window.RTCPeerConnection && !('addTrack' in
+    window.RTCPeerConnection.prototype)) {
+  window.RTCPeerConnection.prototype.addTrack = function(track, stream) {
+    var streams = this.getLocalStreams();
+    if (streams.indexOf(stream) !== -1) {
+      streams[streams.indexOf(stream)].addTrack(track);
+    } else {
+      this.addStream(stream);
```
jan-ivar, line 1732: Assuming I have a `stream` with one video and one audio track, if I do `stream.forEach(track => pc.addTrack(track))` wont the first invocation of `addTrack` here add the stream with both tracks? Wont the second invocation attempt to add a track to the stream it's already in?

jan-ivar, line 1732: Note that `addTrack` on a peer connection is not a proxy for `stream.addTrack`, so the former should never call the latter. See [spec](http://w3c.github.io/webrtc-pc/#widl-RTCPeerConnection-addTrack-RTCRtpSender-MediaStreamTrack-track-MediaStream-streams). It should only add a track to the peer connection, not change any track or stream affiliations in the MediaCapture and Stream spec.

The method instead probably wants to validate that its input arguments already exist in the streams claimed. See [firefox implementation](http://mxr.mozilla.org/mozilla-central/source/dom/media/PeerConnection.js?rev=d2c568567473#1001) for an example.

fippo, line 1732: also: if the user acquires an AV stream but only adds audio this would send video which is a no-no.

Maybe this is not shim-able...

jan-ivar, line 1732: In general, no it is not.

There may be some value perhaps in handling the common case `stream.forEach(track => pc.addTrack(track))` to allow `pc.addStream` to disappear.

fippo:
One way to remedy this would be to require new streams to be still added via addStream -- but that seems dangerous to me since apps might check for the existence of addTrack so shimming a crippled version would be harmful. 

Will add the srcObject-addtrack thing after the module stuff has been merged.

PR #184, opened 2016-02-10
use faucet to print a more human readable output

juandebravo:
[faucet](https://github.com/substack/faucet) defines itself as an _human-readable TAP summarizer_

It helps us to get this output:
screen shot 2016-02-10 at 17 10 23

instead of the current one:
screen shot 2016-02-10 at 17 11 33


juandebravo:
I signed it!

KaptenJansson:
This looks great, could you update your PR to accommodate for the latest changes?

Thanks for doing this!

juandebravo:
thanks @KaptenJansson, master merged to this branch, ready to merge

KaptenJansson:
LGTM!

fippo:
on travis this looks slightly confusing: https://travis-ci.org/webrtc/adapter/jobs/108850206
but that doesn't seem to do any harm so \o/

fippo:
@legastero just showed me https://www.npmjs.com/package/tap-spec which seems slightly more maintained. Maybe we should use that instead? Too many fancy tools!
I do like how faucet hides the details for tests that pass however.

juandebravo:
there're several reporters that we could play with :)

https://github.com/substack/tape

PR #187, opened 2016-02-11
no dep on grunt, set main to adapter_core.js

fippo:
I missed that before... need more :eyes:
I think main should not point to the generated out.js -- and then we don't need grunt as dep, only as dev-dependency.

I also think we don't need to use the module pattern in adapter_core.js but that is a style nit :-)

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "0.3.0",
   "description": "A shim to insulate apps from WebRTC spec changes and browser prefix differences",
   "license": "BSD-3-Clause",
-  "main": "out/adapter.js",
+  "main": "./src/js/adapter_core.js",
   "repository": {
     "type": "git",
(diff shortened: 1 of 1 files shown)


KaptenJansson:
But will this be browserified when installed in another project now?

fippo:
i don't think you need the browserified version when you install it via npm

KaptenJansson:
Yes we do, especially for pages like on samples etc. Yes we could directly link samples to adapter-latest.js in the gh-pages branch but then we introduce some uncertainty (especially considering the tests) by not using a checked in version of adapter. 

Hence why we need these deps.

KaptenJansson:
I do see your point now (for the require() use case), we should at least point to adapter_core.js

KaptenJansson:
LGTM

PR #189, opened 2016-02-11
ontrack shim requires peerconnection shim

fippo:
i think we didn't notice in the tests because lack of t.plan... again.


KaptenJansson:
LGTM

PR #190, opened 2016-02-12
move ontrack test to separate test

fippo:
mostly so we can use t.plan -- we do not know the number of ice candidates in the other tests.

cc @jan-ivar


jan-ivar:
Great!

KaptenJansson:
LGTM

PR #191, opened 2016-02-12
reset blob url on track addition

fippo:
one thing I originally wanted in #182
basically that is required to show video in a scenario like audio->audio/video

Do we want a test for this? I think I can use https://github.com/webrtc/testrtc/blob/master/src/js/camresolutionstest.js#L353-L361 to assert video is shown but that would need an update of the utils first.

### DIFF
--- a/src/js/chrome/chrome_shim.js
+++ b/src/js/chrome/chrome_shim.js
@@ -59,6 +59,13 @@ var chromeShim = {
             this._srcObject = stream;
             // TODO: revokeObjectUrl(this.src) when !stream to release resources?
             this.src = URL.createObjectURL(stream);
+            // We need to recreate the blob url when a track is added or removed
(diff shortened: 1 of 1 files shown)


KaptenJansson:
Do we have to do revoke URL on the old streams as well?

KaptenJansson:
WRT to the tests, they are always welcome however that can be done in a separate PR due to need to add things to utils.

KaptenJansson:
What's needed for revoking? Just keep track of the old stream and revoke it before/after a new track is added?

fippo:
we don't currently revoke so this doesn't change the behaviour. Maybe we should revoke?

fippo:
revoked. @KaptenJansson ^^

KaptenJansson:
Awesome ;). LGTM

PR #203, opened 2016-02-23
bump bowerTo1.0.1

KaptenJansson:


Nek-:
Why not bump to 1.0.2 ? (this fix should have a new release)

KaptenJansson:
Right my bad, was doing things a little to quick. Thanks for pointing this out.

PR #204, opened 2016-02-23
Used another access method to avoid to use the '.new' method

bud-mo:
**Purpose**

Some minifiers like YUI Compressor detect the 'new ' word as reserved keyword.
Using this method to access the property, the compressor will not fail.

### DIFF
--- a/src/js/edge/edge_shim.js
+++ b/src/js/edge/edge_shim.js
@@ -547,18 +547,18 @@ var edgeShim = {
         states[transceiver.dtlsTransport.state]++;
       });
       // ICETransport.completed and connected are the same for this purpose.
-      states.connected += states.completed;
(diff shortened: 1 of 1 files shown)


KaptenJansson:
Thanks for the contribution! Could you also sign the CLA? Thanks!

@fippo PTAL

bud-mo:
@KaptenJansson I'm in trouble to sign the CLA, the form does not allow me to use the secondary address (this one) to sign the CLA. I have submitted my primary address and github account.

bud-mo:
@googlebot Ok ;)

fippo:
LGTM -- i recall not doing this because jshint or jscs complained but if eslint is ok with it... thanks for fixing it!

PR #208, opened 2016-02-24
Update utils.js

srtucker22:
adapter_core.js was throwing an error on my iPhone6 because none of the if statements in detectBrowser were resolving. Maybe there's a more elegant way of handling iOS, but in general, I think it would be best to have a default at the end of the detectBrowser function so it at least returns and doesn't throw errors

### DIFF
--- a/src/js/utils.js
+++ b/src/js/utils.js
@@ -87,6 +87,10 @@ var utils = {
       result.minVersion = 10547;
       return result;
     }
+    
+    // Non supported browser default.
+    result.browser = 'Not a supported browser.';
+    return result;
   }
 };
 


srtucker22:
I signed it!

KaptenJansson:
Thanks for this, it slipped through the cracks!

LGTM.

xdumaine:
@KaptenJansson I'm getting hit by this too, in unit tests with zombie - could you publish a new version with this please? cc @samdutton

samdutton:
Sorry — only just got onto this.

The solution here is fine, but instead I tend to think we should use multiple else/if blocks (not a return for each if block) with one return statement. This would return null for an unsupported browser. 

'Not a supported browser.' feels more like an error message than a property value to me.

KaptenJansson:
I agree, however I just found it cleaner this way when refactoring, previously we had a 500+ line if else statement which makes it intimidating to change anything.

I will take another look at it.

PR #212, opened 2016-03-03
Added missing semicolon on utils.js

bud-mo:
**Description**

Added missing semicolon on utils.js

**Purpose**

Some ancient browsers (like IE 8) Cannot handle correctly the fact that a semicolon is missing at the end of the line.
With this modification the code will run smooth on IE 8.


KaptenJansson:
Well spotted, wonder why the linter did not pick this up.

LGTM.

PR #217, opened 2016-03-04
Edge eventlistener support

fippo:
**Description**
adds EventListener support to the Edge shim

**Purpose**
making .addEventListener('icecandidate') and friends work.

**version considerations**
(not sure if we should have those?)
Not urgent, but should be a minor version. I have some more Edge stuff that i want to merge as well.


KaptenJansson:
LGTM

PR #218, opened 2016-03-04
move minVersion check to utils and clean up browser detection

### LINKED ISSUE #211: Improve browserDetection logic/structure
**Description**
Clean up the browserDetection structure in utils.js

KaptenJansson:
**Description**
Improve browser detection structure
Fixes #211 

**Purpose**
Make the browser detection and minVersion check more succinct, less error prone and cleaner.

### DIFF
--- a/src/js/utils.js
+++ b/src/js/utils.js
@@ -54,34 +54,21 @@ var utils = {
     result.version = null;
     result.minVersion = null;
 
-    function checkMinVersion_(browser, version, minVersion) {
-      // Warn if version is less than minVersion.
-      if (!version && version < minVersion) {
-        utils.log('Browser: ' + browser + ' Version: ' + version + ' < ' +
-            'minimum supported version: ' + minVersion + '\n some things ' +
-            'might not work!');
-      } else {
-        utils.log('Browser version could not be detected, some things might ' +
-            'not work.')
-      }
-    }
-
     if (typeof window === 'undefined' || !window.navigator) {
       result.browser = 'Not a browser.';
+      return result;
     } else if (navigator.mozGetUserMedia) {
       // Firefox.
       result.browser = 'firefox';
       result.version = this.extractVersion(navigator.userAgent,
           /Firefox\/([0-9]+)\./, 1);
       result.minVersion = 31;
-      checkMinVersion_(result.browser, result.version, result.minVersion);
     } else if (navigator.webkitGetUserMedia && window.webkitRTCPeerConnection) {
       // Chrome, Chromium, WebView, Opera and other WebKit browsers.
       result.browser = 'chrome';
       result.version = this.extractVersion(navigator.userAgent,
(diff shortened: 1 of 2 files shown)


KaptenJansson:
@fippo @vganapathy-lifesize PTAL.

fippo:
LGTM. In terms of style I would slightly prefer returning early in case of a non-supported browser and moving the checkMinVersion_ stuff to the end (and not as a function). Your call.

KaptenJansson:
PTAL

src/js/utils.js
```diff
@@ -54,42 +54,42 @@ var utils = {
...
+    } else if(navigator.mediaDevices &&
         navigator.userAgent.match(/Edge\/(\d+).(\d+)$/)) {
+      // Edge.
       result.browser = 'edge';
       result.version = this.extractVersion(navigator.userAgent,
           /Edge\/(\d+).(\d+)$/, 2);
       result.minVersion = 10547;
-      return result;
+    } else {
+      result.browser = 'Not a supported browser.';
```
fippo, line 80: returning here as well will save you the !resultion.version check below

src/js/utils.js
```diff
@@ -54,42 +54,42 @@ var utils = {
...
+    }
+
+    // Warn if version is less than minVersion.
+    if (!result.version  && !result.minVersion && !result.browser &&
+        result.version < result.minVersion) {
+      utils.log('Browser: ' + result.browser + ' Version: ' + version + ' < ' +
+          'minimum supported version: ' + minVersion + '\n some things ' +
+          'might not work!');
+    } else {
+      utils.log('Browser version could not be detected, some things might ' +
```
fippo: move this to the other else

src/js/utils.js
```diff
@@ -54,42 +54,42 @@ var utils = {
...
       result.version = this.extractVersion(navigator.userAgent,
           /Edge\/(\d+).(\d+)$/, 2);
       result.minVersion = 10547;
-      return result;
+    } else {
+      result.browser = 'Not a supported browser.';
+    }
+
+    // Warn if version is less than minVersion.
+    if (!result.version  && !result.minVersion && !result.browser &&
```
fippo: if (result.version < result.minVersion) { ...

KaptenJansson: Still need to check that !result.version or maybe we want to crash and fail because it means our extractVersion is failing?

src/js/utils.js
KaptenJansson: I do not think we need this when changing it like you suggested.

src/js/utils.js
KaptenJansson, line 80: Done

KaptenJansson:
PTAL @samdutton

samdutton:
LGTM.

KaptenJansson:
Thanks!

PR #220, opened 2016-03-09
move getusermedia to separate files

fippo:
**Description**
moves getUserMedia shims to separate files

**Purpose**
so they can be included without pulling the RTCPeerConnection shim code in.

**Open issues**
Not sure if we want to go in that direction long-term. I think we should at least get rid of the intermediate chromeShim and firefoxShim objects and directly set module.exports

### DIFF
--- a/src/js/chrome/chrome_shim.js
+++ b/src/js/chrome/chrome_shim.js
@@ -203,131 +203,6 @@ var chromeShim = {
     });
   },
 
-  shimGetUserMedia: function() {
-    var constraintsToChrome_ = function(c) {
-      if (typeof c !== 'object' || c.mandatory || c.optional) {
-        return c;
-      }
-      var cc = {};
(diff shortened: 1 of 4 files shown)


KaptenJansson:
SGTM. The reason I left chromeShim is that it's visible in the object exposed to the browser, i.e. adapter.chromeShim.shimPeerconnection etc.

But I'm fine removing it. LGTM

fippo:
ah... but chromeShim in src/js/chrome/chrome_shim doesn't become exposed because scoping. Anyway... :-)

can I get a minor version please once #222 is in?

KaptenJansson:
Yeah, I think it did at one point when I was testing stuff but then changed the scope without reflecting over the names.

KaptenJansson:
Will try to create a release soon.

PR #223, opened 2016-03-10
change the way candidates are emitted; add icegatheringstate

fippo:
**Description**
- fixes p2p test when promises are uses
- adds icegathering state

The logic to start emitting candidates after the promise is resolved is getting crazy.
@pthatcherg: is that any different in the browser?


KaptenJansson:
LGTM

PR #225, opened 2016-03-16
minify generated javascript file with grunt uglify

ridakk:
**Description**
Added grunt clean and uglify tasks to minift generated outputs

**Purpose**
To reduce generated file sizes


ridakk:
I signed it!

PR #227, opened 2016-03-16
set src to empty string when setting null srcobject

### LINKED ISSUE #224: Errors on chrome when setting `srcObject = undefined`
Chrome.

**Description**
If `video.src` is set to `undefined` it explodes because `URL.createObjectURL` doesn't like being given undefined as an argument here: https://github.com/webrtc/adapter/blob/master/src/js/chrome/chrome_shim.js#L63

Chatting to @fippo, this should probably be something like this: (haven't tested), also fixes the missing `self` he identified.

``` diff
         set: function(stream) {
+           var self = this;
            // Use _srcObject as a private property for this shim
            this._srcObject = stream;
            if (this.src) {
              URL.revokeObjectURL(this.src);
            }

+           if (!stream) {
+             this.src = undefined;
+             return;
+           }
+
            this.src = URL.createObjectURL(stream);
            // We need to recreate the blob url when a track is added or removed.
            // Doing it manually since we want to avoid a recursion.
            stream.addEventListener('addtrack', function() {
              if (self.src) {
                URL.revokeObjectURL(self.src);
              }
              self.src = URL.createObjectURL(stream);
            });
            stream.addEventListener('removetrack', function() {
              if (self.src) {
                URL.revokeObjectURL(self.src);
              }
              self.src = URL.createObjectURL(stream);
            });
          }
```

fippo:
**Description**
sets src to the empty string when setting the srcObject to null. Also bails out earlier and doesn't try to add event listeners.

**Purpose**
fixes #224

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -714,6 +714,66 @@ test('srcObject set from another object', function(t) {
   });
 });
 
+test('srcObject null setter', function(t) {
+  var driver = seleniumHelpers.buildDriver();
+
+  // Define test.
+  var testDefinition = function() {
+    var callback = arguments[arguments.length - 1];
+
+    var constraints = {video: true, fake: true};
+    navigator.mediaDevices.getUserMedia(constraints)
+    .then(function(stream) {
+      window.stream = stream;
+
+      var video = document.createElement('video');
+      video.setAttribute('id', 'video');
+      video.setAttribute('autoplay', 'true');
+      document.body.appendChild(video);
+      video.srcObject = stream;
+      video.srcObject = null;
+
+      callback(null);
+    })
+    .catch(function(err) {
+      callback(err.name);
+    });
+  };
+
+  // Run test.
+  driver.get('file://' + process.cwd() + '/test/testpage.html')
+  .then(function() {
+    t.plan(3);
+    t.pass('Page loaded');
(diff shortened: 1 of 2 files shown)


KaptenJansson:
LGTM when test is fixed https://travis-ci.org/webrtc/adapter/jobs/116367971

test/test.js
```diff
@@ -714,6 +714,66 @@ test('srcObject set from another object', function(t) {
...
+    // 5 second timeout in case the event does not fire for some reason.
+    return driver.wait(webdriver.until.elementLocated(
+      webdriver.By.id('video')), 3000);
+  })
+  .then(function() {
+    return driver.executeScript(
+        'return document.getElementById(\'video\').src');
+  })
+  .then(function(src) {
+    t.ok(src === 'file://' + process.cwd() + '/test/testpage.html',
```
KaptenJansson: Why are you checking that the src is the test page? Shouldn't be src === ''?

fippo: that is what I assumed initially as well but... maybe one day @samdutton can explain this :-)

fippo: oh. in firefox it is still ''. Odd...

KaptenJansson:
LGTM

PR #228, opened 2016-03-16
fix broken candidate sdp

fippo:
**Description**
I added candidates to the sdp without thinking that candidates are defined without the a= in the ice candidate event...

I think i need some edge-specific tests for all the SDP behaviour :-(


KaptenJansson:
LGMT

Issue #229, opened 2016-03-17
ontrack test failure


fippo:
the ontrack tests currently fails in Firefox. What is rather confusing is that it worked until a few days ago and still works if one runs just this test. @jan-ivar any ideas?

PR #230, opened 2016-03-17
deactivate ontrack test in firefox

fippo:
**Description**
deactivates ontrack test in Firefox until we have had time to investigate #229


KaptenJansson:
LGTM

PR #235, opened 2016-03-17
fix adapter-latest.js typo

ngauthier:
**Description**

Fixes typo in adapter-latest.js symlink (extra s from 68a18a6aabef42b53da540ddfb384cbcf56928aa )

**Purpose**

Fix the link in the readme so that it correctly points to the latest version.

:heart: @ngauthier


KaptenJansson:
Sorry about this again.....So much for doing things like this at 6pm just before going home....

KaptenJansson:
And Thanks for fixing!

KaptenJansson:
And I will make this a more automatic process so that the human errors like this do not happen again.

ngauthier:
NP! Thanks for the quick merge.

KaptenJansson:
semi-automatic publish fix is in #239. Yes there are that many steps required.....Will consider making it more portable and pretty in the future but it will do for now.

PR #240, opened 2016-03-19
Edge: add non-trickle support

fippo:
**Description**
improves non-trickle support for Edge

**Purpose**


KaptenJansson:
Awesome. LGTM

PR #242, opened 2016-03-22
Safari support

agouaillard:
**Description**
Early safari support in the utils (browser detect), adapter-core (main switch), and specific shim.

**Purpose**
Support safari.


KaptenJansson:
Thanks for the contribution!

This is progress, more browsers ;).

One thing though, I'd rather wait submitting this until the shim is usable. Feel free to keep adding commits here and let me know when it's ready.

agouaillard:
hi chris,

I had though about that: when is the code "usable":
1 - after I upgrade the browser detection?
2 - after the shim is "complete" ?
3 - after safari driver is added to the selenium / travis-cl / test part ?

#1 is always good, even if you have no corresponding shim.
#2 is intrinsically flawed, since shim are never complete, and can become
incomplete with each new version of a browser, or each added browser, or
each revision of the spec. So here, going step by step is not a problem.
#3 is a bonus, but today, there is a lot more work involved in supporting
safari than chrome or FF for there reasons:
- the safari driver is less mature
- safari nightly need to be tested, not safari, and the driver does not
  seem to make that difference out of the box
- travis-cl support of maces and safari is also not mature.
  so we should not wait for #3 to commit #1 and #2.

Finally, with fippo pushing commits everyday, rebasing is problematic, the
more we wait, the more likely it is I will have to manually rebase.

I would prefer to commit step by step, especially since I Do not believe
this will break anybody's code ......

On Wed, Mar 23, 2016 at 1:30 AM, Christoffer Jansson <
notifications@github.com> wrote:

> Thanks for the contribution!
> 
> This is progress, more browsers ;).
> 
> One thing though, I'd rather wait submitting this until the shim is
> usable. Feel free to keep adding commits here and let me know when it's
> ready.
> 
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> https://github.com/webrtc/adapter/pull/242#issuecomment-200244597

## 

## Alex. Gouaillard, PhD, PhD, MBA

Principal Architect - Citrix, San Francisco

## President - CoSMo Software Consulting, Singapore

sg.linkedin.com/agouaillard

   -

src/js/adapter_core.js
```diff
@@ -27,9 +27,13 @@
   var chromeShim = require('./chrome/chrome_shim') || null;
   var edgeShim = require('./edge/edge_shim') || null;
   var firefoxShim = require('./firefox/firefox_shim') || null;
+  var safariShim = require('./safari/safari_shim') || null;
 
   // Shim browser if found.
   switch (browserDetails.browser) {
+    // NOTE ALEX: for the time being, just fall through
+    // ...        Eventually use the safari shim when ready.
```
fippo: this require's without actually doing anything?

src/js/safari/safari_shim.js
```diff
@@ -0,0 +1,44 @@
...
+  },
+
+  shimPeerConnection: function() {
+  },
+
+  shimGetUserMedia: function() {
+    // NOTES ALEX: both navigatorusermedia and mediaDevices version exist ...
+    // ...         the former, prefixed with "webkit" is a JS implementation
+    // ...         that wraps the MediaDevices one, as per latest specs.
+    navigator.getUserMedia = navigator.webkitGetUserMedia(constraints, onSuccess, onError);
```
fippo: this doesn't look like it is going to work

src/js/utils.js
```diff
@@ -59,26 +59,42 @@ var utils = {
...
+      } else {
+        result.browser = 'safari';
+        result.version = this.extractVersion(navigator.userAgent,
+          /AppleWebKit\/([0-9]+)\./, 1);
+        result.minVersion = 602;
+      }
+
+    } else if (navigator.mediaDevices &&
       // Edge.
+      navigator.userAgent.match(/Edge\/(\d+).(\d+)$/)) {
```
fippo: can you unbreak that so it doesn't insert the comment in the middle of the condition please?

fippo:
(3) is not a blocker, Edge is still running without CI coverage. And we have had getUserMedia support for Edge without the rest of it so (2) is not either.

But I have some doubts the code works as-is.

src/js/adapter_core.js
agouaillard: Detecting safari properly has its own value. I can comment out the "require" for the time being if you prefer. Please advise. NOTE: I tried to contact you to speak about all this (+IMTC) and failed miserably, could we have a quick chat?

src/js/safari/safari_shim.js
agouaillard: You can see the results [here](http://webrtcbydralex.com/index.php/2016/03/17/updated-webrtc-support-scores/), and admire how well it works :-)

agouaillard:
chris, that should be usable now. I ran a full W3C test on all sorts of browsers.

fippo, FYI, I updated the W3C media stream tests. FF already score much better on mediastreamtrack APIs. You might want to take a look. Much more coming this week and next week.

src/js/safari/safari_shim.js
```diff
@@ -0,0 +1,42 @@
...
+ *  that can be found in the LICENSE file in the root of the source
+ *  tree.
+ */
+'use strict';
+var logging = require('../utils.js').log;
+var browserDetails = require('../utils.js').browserDetails;
+
+var safariShim = {
+  // TODO: alex, should be here, double check against LayoutTests
+  shimOnTrack: function() { },
```
KaptenJansson: Comment out or remove unusable methods. You can always use the other shims as a reference/to do list.

src/js/safari/safari_shim.js
```diff
@@ -0,0 +1,42 @@
...
+
+  // TODO: alex, not sure if the HTMLMediaElement has been updated,
+  // ... double-check anyway
+  shimSourceObject: function() { },
```
KaptenJansson: Comment out or remove unusable methods.

src/js/safari/safari_shim.js
```diff
@@ -0,0 +1,42 @@
...
+  attachMediaStream: function(element, stream) { },
```
KaptenJansson: Comment out or remove unusable methods. Also worth noting is that this will soon be removed as it's a wrapper, not a shim and most browser support srcObject now.

src/js/safari/safari_shim.js
```diff
@@ -0,0 +1,42 @@
...
+  reattachMediaStream: function(to, from) { },
```
KaptenJansson: Comment out or remove unusable methods. Also worth noting is that this will soon be removed as it's a wrapper, not a shim and most browser support srcObject now.

src/js/safari/safari_shim.js
```diff
@@ -0,0 +1,42 @@
...
+
+  // TODO: once the back-end for the mac port is done, add.
+  // TODO: check for webkitGTK+
+  shimPeerConnection: function() { },
```
KaptenJansson: Comment out or remove unusable methods.

src/js/safari/safari_shim.js
```diff
@@ -0,0 +1,42 @@
...
+  // ... just need to remove the prefix. navigator.GUM is
+  // ... but a wrapper for mediaDevices.GUM. 
+  shimGetUserMedia: function() {
+    navigator.getUserMedia = navigator.webkitGetUserMedia;
+  },
+}
+
+// Expose public methods.
+module.exports = {
+  shimOnTrack: safariShim.shimOnTrack,
```
KaptenJansson: Remove exports for non usable methods/shims. Applies to all except shimGetUserMedia.

KaptenJansson:
Linting was disabled for most files, I've now added eslint (which we have planned to use for some time) so you will most likely hit some linting issues. So in addition to addressing my comments, please address the lint issues as well (can be run separately by invoking `grunt lint`). Thanks!

src/js/safari/safari_shim.js
agouaillard: OK

src/js/safari/safari_shim.js
agouaillard: OK

src/js/safari/safari_shim.js
agouaillard: I just checked their unit tests and with the engineers and safari supports srcObject. Will remove this.

src/js/adapter_core.js
```diff
@@ -29,15 +29,18 @@
...
   // Shim browser if found.
   switch (browserDetails.browser) {
+    case 'opera': // fallthrough as it uses chrome shims
     case 'chrome':
       if (!chromeShim || !chromeShim.shimPeerConnection) {
         logging('Chrome shim is not included in this adapter release.');
         return;
       }
-      logging('adapter.js shimming chrome!');
+      logging(`adapter.js shimming ${browserDetails.browser}`
```
KaptenJansson: ECMA 6 stuff, neat! 

Checking https://kangax.github.io/compat-table/es6/ (see template literals section) for templating compatibility it looks like most new browsers support this but not sure where adapter.js is used exactly. Browsers like iOS8/safari 8 or older does not support it. 

Do you think the neatness is worth the risk here? If so then I'm fine with it.

agouaillard: You are right. Looking at the table, first safari 8 is not that old (2014), and then everybody using plugins for safari and IE would have errors. Let me correct that tomorrow with a less elegant solution but which would work with older browsers.

KaptenJansson: I forgot to add that I do realize those older browsers do not have webrtc in them but it's still nice to have the browser detection working properly and returning non supported browser.

Thanks for understanding.

KaptenJansson:
Ping

src/js/utils.js
```diff
@@ -59,28 +59,69 @@ var utils = {
...
+    } else if (navigator.webkitGetUserMedia) {
+      // Chrome, Chromium, Webview, Opera, all use the chrome shim for now
+      if (window.webkitRTCPeerConnection) {
+        if (navigator.userAgent.match(/OPR/)) {
+          result.browser = 'opera';
+        } else {
+          result.browser = 'chrome';
+        }
+        result.version = this.extractVersion(navigator.userAgent,
           /Chrom(e|ium)\/([0-9]+)\./, 2);
```
KaptenJansson, line 85: Does this work on opera?

According to http://www.useragentstring.com/pages/All/ Opera browsers have Opera in the User Agent string.

agouaillard, line 85: I will double check but this is what the userAgent returns for opera:

"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_4) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/49.0.2623.87 Safari/537.36 OPR/36.0.2130.46"

"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_4) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/50.0.2657.0 Safari/537.36 OPR/37.0.2171.0 (Edition developer)"

agouaillard, line 85: the website you link seems quite old ... e.g. latest listed safari version is 7.x ....

fippo, line 85: I don't think it is a good idea to change this to 'opera' if that changes adapter.browserDetails.browser as well.
That would be a breaking change and I don't want to bump the version just for this. (arguably, this kind of detection should move out of adapter but...)

KaptenJansson, line 85: Ok I was just worried that we would not get a version number when Opera was used due to checking for "/Chrom(e|ium)" but my regexp foo is weak, and yes I should have looked at the validity of that page.

Also I agree that this can be bad. Surely there is some library on Github that does this, maybe we should take a look around for a better dedicated browser detection?

src/js/utils.js
```diff
@@ -59,28 +59,69 @@ var utils = {
...
+        //
+        if (navigator.userAgent.match(/Version\/(\d+).(\d+)/)) {
+          result.browser = 'safari';
+          result.version = this.extractVersion(navigator.userAgent,
+            /AppleWebKit\/([0-9]+)\./, 1);
+          result.minVersion = 602;
+
+        // unknown webkit-based browser
+        } else {
+          result.browser = 'Unsupported webkit-based browser \
```
KaptenJansson, line 110: I prefer to use concatenate for this, but the linter should catch this. Will let this pass and fix the linter.

KaptenJansson:
LGTM

src/js/utils.js
fippo, line 85: reverting the opera change in https://github.com/webrtc/adapter/pull/274 -- I don't see what makes opera different from an adapter perspective.

agouaillard, line 85: to KJ:
OPR is the keyword to detect opera in the UA string, but everything else is like chromium, so we can use the same version number detection regex. Confirmed by opera eng.

To both, 

I think browser detection and "adaptation" should be dealt with separately, and actually this is the way it s being done in the code. The fact that we use the same shim for chrome than for opera should not impact the browser detection algorithm. If opera is supported, we should recognize opera first, and then let the adapter decide how to shim.

I have two use cases in mind reported by some of my engineers:
- users get confused when the console shows a message saying "chrome" was detected in opera and report a bug.
- users base their own code on adapter.js, but have special magic for opera, and would like to have a place holder to inject their opera code. Today, they need to rewrite their own browser detection, since adapter.js falsely detect opera as chrome.

Speaking about the breaking change, it's all a matter of backward compatibility. If an error, or oddity, has been merged in in the past, should we correct it and do the right thing, or leave it as it because code in the wild might depend on it? Unless there is a documented backward compatibility policy that commit us to support existing bugs for a certain amount of time, I'd say: let's do the right thing.

fippo, line 85: since opera doesn't behave differently for the purpose of adapter.js I don't see any value in doing a major version bump.

If you want to do browser detection, do it outside adapter. The "bug" here is that adapter does browser detection and exposes it.

agouaillard, line 85: So you deny value in both of the use cases I presented, is that correct?

Shall we open a bug against adapter doing browser detection and remove it from the code? if not, we should do it right.

fippo, line 85: console logging has been disabled by default so (1) is not valid anymore.

For (2), what you are implying is that libraries that rely on adapters browser detection to make decisions about whether to apply certain peerconnections operations should add a special case for 'opera' (as you have done in this PR) even though opera does not behave substantially different from a peerconnection or gum point of view.

While breaking changes are cheap due to semver, I don't see a reason to make one for marginal gains such as this.

agouaillard, line 85: (1) fair enough.

(2) I can live with that. I'm curious though: do we have any example or metric of how much code is actually depending on the value of that field? Within adapter itself, outside of this big switch, nothing depends on it, correct? It might be totally possible that the marginal gain comes at an even smaller price.

KaptenJansson, line 85: Ideally I would like to have separate shims for each browser but since Opera (among a plethora of other browsers but it is probably the biggest) is based on Chromium and have the same API's that we shim (I think?) for Chrome it makes sense to use the Chrome shim. But I've seen cases where API's are missing or are different in Opera in the past, cannot remember exactly which ones now though. 

Now looking back at identifying the browser as Opera, the only thing that would break are the tests unless developers are using the browser detection as well? Since in some of the test we check for if the browser is Chrome or Firefox and adjust the result accordingly.

I guess the concern here is that we are adding support for a browser that will most likely work but we do not have any tests running for it hence we cannot support it properly. I guess this is true for Safari as well but that is at least running it's own shim.

agouaillard, line 85: Fair enough, I will open a ticket for automated opera testing (and safari, for that matter), assign it to me, and revisit this discussion once we have a better safety net.

KaptenJansson, line 85: SGTM. Also FYI just being able to run tests locally is better than no tests at all (downloading browsers automatically using travis is not easy, especially when they require specific OS's like MS edge and Safari). I.e. implementing tests running locally should be a first step.

ibc: Regarding proper browser detection, `bowser` library does it very well:
https://www.npmjs.com/package/bowser

KaptenJansson: aha thanks for the tip!

Issue #243, opened 2016-03-23
Shimming RTCPeerConnection.getStats() does not fix Chrome reports when using Promises


gregor:
**Browsers and versions affected**
Chrome

**Description**
If you implement RTCPeerConnection.getStats() using promises, webRTC adapter fails to map the reports Chrome returns into standards compliant ones.

**Steps to reproduce**

``` coffeescript
pc = new window.RTCPeerConnection pcConfig
pc.getStats()
.then (track_stats) => console.log track_stats
```

**Expected results**
This should return key, value pairs.

**Actual results**
 Instead you get the non-compliant Chrome reports with track_stats.result() etc.

The broken code can be found in:
https://github.com/webrtc/adapter/blob/c3a32562ecbf5b0e7cfb76807a3b522d60b42028/src/js/chrome/chrome_shim.js#L140-L149

The stats will be shimmed if you pass in 1 argument and that argument (the selector) is null. Workaround would be:

``` coffeescript
pc = new window.RTCPeerConnection pcConfig
pc.getStats null
.then (track_stats) => console.log track_stats
```

If you want to make use of a selector, it will currently not work.

fippo:
yeah, the behaviour is bad but somewhat consistent with the chrome behaviour of returning the old stats unless a track selector (or null) is passed

gregor:
Problem is that that it will only return new stats if you pass in null as the selector. If you do pass in a selector it will return Chrome style stats.

fippo:
IIRC the reason for not shimming this was that chrome [doesn't accept a track as first argument anyway](https://bugs.chromium.org/p/webrtc/issues/detail?id=2031). Just returning full stats without looking at the selector is probably going to lead to confusion as well but probably better than the current behaviour.

@alvestrand WDYT?

alvestrand:
That makes no sense. If you're shimming, you can pass the first argument to the shimmed function as the second argument to the built-in function.

It looks as if it is actually being done for the callback case on line 135, but not for the promise case. Why not?

https://github.com/webrtc/adapter/blob/master/src/js/chrome/chrome_shim.js

IDL for Chrome's current getstats:

https://code.google.com/p/chromium/codesearch#chromium/src/third_party/WebKit/Source/modules/mediastream/RTCPeerConnection.idl&q=GetStats%20file:.*.idl&sq=package:chromium&type=cs&l=111

KaptenJansson:
[...]

PR #244, opened 2016-03-23
Enable fixChromeStats_() if window.RTCPeerConnection.getStats() used …

gregor:
**Description**
Enable fixChromeStats_() if window.RTCPeerConnection.getStats() used with a selector is passed in. This only partially fixes the stats compat issues but is a first step.

https://github.com/webrtc/adapter/issues/243

### DIFF
--- a/src/js/chrome/chrome_shim.js
+++ b/src/js/chrome/chrome_shim.js
@@ -138,7 +138,7 @@ var chromeShim = {
 
         // promise-support
         return new Promise(function(resolve, reject) {
-          if (args.length === 1 && selector === null) {
+          if (args.length === 1 && selector === 'object') {
             origGetStats.apply(self, [
                 function(response) {
                   resolve.apply(null, [fixChromeStats_(response)]);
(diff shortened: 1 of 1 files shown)


gregor:
I signed it!

src/js/chrome/chrome_shim.js
```diff
@@ -138,7 +138,7 @@ var chromeShim = {
 
         // promise-support
         return new Promise(function(resolve, reject) {
-          if (args.length === 1 && selector === null) {
+          if (args.length === 1 && selector === 'object') {
```
fippo: && typeof(selector) === 'object'? Which covers null and MediaStreamTrack afaics

gregor: That's what I had in my code locally and messed up opening this PR. No need for the brackets but `typeof` is what I was aiming for. This way you can have a standards compliant Promise request to `RTCPeerConnection.getStats()` even if Chrome might not properly apply the selector.

KaptenJansson:
LGTM

Issue #232, opened 2016-03-17
Add SDP tests for Edge


KaptenJansson:
**Browsers and versions affected**
MS edge

**Description**
Since MS edge has a SDP shim implementation we need specific tests.

KaptenJansson:
Do we need more here or can we close this?

fippo:
I think i'll do them over in https://github.com/fippo/sdp

PR #245, opened 2016-03-26
add edge specific sdp tests

fippo:
**Description**
first stab at #232 
also parses the RED from the rtpmap as fecMechanism.

**Purpose**
moar tests!


test/sdp.js
```diff
@@ -0,0 +1,26 @@
+/*
+ *  Copyright (c) 2015 The WebRTC project authors. All Rights Reserved.
```
KaptenJansson: Should be 2016. Really need to add some auto update mechanism for this.

KaptenJansson:
LGTM with nit.

PR #246, opened 2016-03-27
change resize to loadedmetadata since edge does not implement resize

fippo:
**Description**
Edge doesn't seem to trigger resize so we need to use another event. loadedmetadata seems to work fine but travis will tell us if not :-)

**Purpose**
more green tests for Edge


KaptenJansson:
Yeah this should now work since firefox has fixed their issue with only updating videoWidth/height after receiving a frame which does not happen before loadedmetadata has fired obviously. Not sure if ESR has caught up to it though but since the tests are passing we can probably assume so ;)

LGTM!

PR #247, opened 2016-03-27
fix edge ice failures

fippo:
**Description**
fixes a bug I introduced in a22115f8ea26d71de6e2f461f7dc04c47ddeef8a and is more correct about setting the gathering state to complete.

### DIFF
--- a/src/js/edge/edge_shim.js
+++ b/src/js/edge/edge_shim.js
@@ -117,7 +117,8 @@ var edgeShim = {
               sections[j] += 'a=end-of-candidates\r\n';
             }
           }
-        } else {
+        } else if (event.candidate.candidate.indexOf('typ endOfCandidates')
+            === -1) {
           sections[event.candidate.sdpMLineIndex + 1] +=
               'a=' + event.candidate.candidate + '\r\n';
         }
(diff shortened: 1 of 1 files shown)


src/js/edge/edge_shim.js
```diff
@@ -117,7 +117,8 @@ var edgeShim = {
               sections[j] += 'a=end-of-candidates\r\n';
             }
           }
-        } else {
+        } else if (event.candidate.candidate.indexOf('typ endOfCandidates')
```
KaptenJansson, line 120: should this be type?

fippo, line 120: no, rfc 5245 calls this "typ":

```
cand-type             = "typ" SP candidate-types
```

src/js/edge/edge_shim.js
```diff
@@ -311,13 +308,15 @@ var edgeShim = {
...
-      var sessionpart = sections.shift();
+      var sections;
+      var sessionpart;
       if (description.type === 'offer') {
         if (!this._pendingOffer) {
         } else { 
           // VERY limited support for SDP munging. Limited to:
           // * changing the order of codecs
+          sections = SDPUtils.splitSections(description.sdp);
+          sessionpart = sections.shift();
```
KaptenJansson, line 319: [shift()](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/shift) will return `undefined` if length is 0, will this cause issues? I.e. do we need to catch this or do we want it to fail loudly?

fippo, line 319: that would be an sdp which doesn't contain anything... filed https://github.com/fippo/adapter/issues/22 so I don't forget to add something

KaptenJansson:
LGTM

PR #248, opened 2016-03-28
better way to skip tests

fippo:
**Description**
a better way to skip tests from [the documentation](https://github.com/substack/tape#testname-opts-cb)

**Purpose**


KaptenJansson:
Wow that's really neat! LGTM

PR #249, opened 2016-03-30
Removed unnecessary comma

bud-mo:
**Description**

With this modification the file is parsed correctly by IE

**Purpose**

Some browsers like IE does not like the unnecessary commas and fail to parse the script.


KaptenJansson:
Thanks for this, will fix the linter which currently does work properly. LGTM

PR #250, opened 2016-03-30
don't use shell script for running the tests

fippo:
**Description**
replaces the shell script for running tests and travis-multirunner with a nodejs version 

**Purpose**
make it possible to run tests on windows using 'npm test'


KaptenJansson:
Nice! Should we also fallback to use already installed browsers if not on linux?

fippo:
yeah. Probably by removing the .setChromeBinaryPath et al... let me add that.

fippo:
ok, linux + windows work well. Will test on mac in a bit.

fippo:
WFM on mac too

agouaillard:
Works on mac, for chrome and firefox stable.
@fippo are the dev versions supported as well (is yes, what's the VER tag?)

fippo:
@agouaillard not yet, we need a cross-platform replacement for setBinary and setChromeBinaryPath. This is currently only done on linux

agouaillard:
OK, I'll take a look on mac then, this it s my main platform for now. If I understand correctly, travis scripts are populating the web driver's set[Chrome]Binary[Path]() with the right binary on linux for the time being, correct?
And on non-linux, it just takes the default install for each?

KaptenJansson:
LGTM

PR #251, opened 2016-03-30
Conditionally do 'console.log' call

bud-mo:
**Description**

Verify the presence of the console.log function before to call it.

**Purpose**

Some browsers fail on console object


KaptenJansson:
LGTM

PR #255, opened 2016-03-30
reactivate ontrack test

### LINKED ISSUE #229: ontrack test failure

fippo:
**Description**
trying to fix #229, works locally for me...


KaptenJansson:
LGTM

PR #258, opened 2016-03-30
make it work on the secret sauce

fippo:
**Description**
test edge

@KaptenJansson will send you my credentials that you need to set as an env variable in travis.

### DIFF
--- a/test/selenium-lib.js
+++ b/test/selenium-lib.js
@@ -12,6 +12,7 @@
 var webdriver = require('selenium-webdriver');
 var chrome = require('selenium-webdriver/chrome');
 var firefox = require('selenium-webdriver/firefox');
+var edge = require('selenium-webdriver/edge');
 var fs = require('fs');
 var os = require('os');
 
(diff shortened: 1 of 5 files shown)


fippo:
using my fork of travis-multirunner until we get this working and can make a upstream PR. will squish later

KaptenJansson:
LGTM, will merge as soon as I get credentials.

fippo:
TODO: replace driver.get with executeScript(fs.readFileSync('out/adapter.js') -- preferably for all browsers but doing it for Edge would be ok if things break a lot otherwise (they currently do in Chrome)

KaptenJansson:
When we need stuff in the DOM like using the video tag etc, can all that be done just via javascript without HTML page loaded?

fippo:
yes. However, you need to reload the page between runs which I currently don't do and which leads to interesting effects.

saucelabs is going to be difficult because of admin-permissions for registry keys :-/

fippo:
ok... I set up a VM in azure using the TH2/stable release of windows 10 from [these images](https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/windows/) provided by Microsoft. Downloaded the linux image, converted to vhd, then uploaded and installed git, node (not sure why, probably local testing), java (for selenium) and manycam. Finally started a selenium hub and node.
Will probably have to repeat that for the current insider windows version as well.

@KaptenJansson you just got email with the value SELENIUM_SERVER needs to be set to in the travis setting pane.

test/selenium-lib.js
```diff
@@ -82,23 +83,42 @@ function buildDriver() {
...
   // callbacks to be caught in tests.
-  sharedDriver.manage().timeouts().setScriptTimeout(5000);
+  sharedDriver.manage().timeouts().setScriptTimeout(10 * 1000);
 
   return sharedDriver;
 }
 
+// loads the dummy page that includes adapter.js.
+// In Microsoft Edge (via selenium) this directly injects adapter.js.
+function loadTestPage(driver) {
```
fippo, line 114: do we always want to use this about:blank+executeScript style?

KaptenJansson:
SELENIUM_SERVER variable set.

test/selenium-lib.js
KaptenJansson, line 114: If it works for all browsers why not, since the tests are running we are actually testing that window.adapter object works (i.e. the DOM is not needed to verify that).

KaptenJansson, line 114: Also, I would want this to be ported to the utils repo.

fippo, line 114: made an issue in utils repo to track that

fippo:
uhm. now travis does not like me anymore?

fippo:
Ok... as far as I am concerned it is sufficient to run this before you publish a new version. Which means we can store the URL (will change soon) in travis without exposing it.

For local development on non-windows this means I have to set SELENIUM_SERVER but that works for me.

Sounds good? Then i'll squash and ping you when this can be merged.

KaptenJansson:
LGTM

PR #262, opened 2016-04-09
make ontrack pass again

fippo:
ontrack seems flaky. It seems that ice connection state can go to connected or completed before ontrack fires. Which is really, really odd. @jan-ivar you might want to take a look.

This adds a one-second sleep which seems sufficient to fix that.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -2197,6 +2197,7 @@ test('ontrack', function(t) {
   .then(function(pc1ConnectionStatus) {
     t.ok(pc1ConnectionStatus === 'completed' || 'connected',
       'P2P connection established');
+    driver.sleep(1000); // flaky in firefox
     return driver.executeScript('return window.testPassed');
   })
   .then(function(testPassed) {


KaptenJansson:
I do not like "sleep" but I do not see an immediate solution to this. LGTM

jan-ivar:
@fippo @KaptenJansson They do indeed race, but where in the spec does it say `ontrack` must fire before `"connected"`?

fippo:
ontrack must fire during SRD so tracks can be rejected
this is even mentioned in an issue in the spec:

```
For the answerer, ontrack will continue to fire during setRemoteDescription,
as it does today (this is necessary to allow the answerer to reject offered
tracks by stopping them).
```

http://w3c.github.io/webrtc-pc/#processing-remote-mediastreamtracks seems like the right place to describe that this happens after SRD and before the SRD callback. Maybe that is hidden some in JSEP?

jan-ivar:
Thanks, it is indeed described there (step 7), and this _dispatch a receiver_ procedure is called synchronously from setRD. Turns out this is [bug 998546](https://bugzil.la/998546).

PR #263, opened 2016-04-09
remove unnecessary code in generateCertificate wrapper

fippo:
**Description**
removes unnecessary code. As @jan-ivar pointed out, arguments.length for the getter will always be 0. Not sure how I came up with that code...

**Purpose**
reducing code size. We still want this to disappear.


KaptenJansson:
LGTM

PR #264, opened 2016-04-09
re-enable icetransports relay test in FF

fippo:
this test was previously disabled because FF ESR was 38 and iceTransportPolicy was [implemented in FF 42](https://bugzilla.mozilla.org/show_bug.cgi?id=1187775)


KaptenJansson:
LGTM

PR #265, opened 2016-04-10
improved ice server filter

fippo:
**Description**
improved ice server filter. Should make apprtc work but if you could test it before merging that would be great.

### DIFF
--- a/src/js/edge/edge_shim.js
+++ b/src/js/edge/edge_shim.js
@@ -95,9 +95,10 @@ var edgeShim = {
         this.iceOptions.iceServers = config.iceServers.filter(function(server) {
           if (server && server.urls) {
             server.urls = server.urls.filter(function(url) {
-              return url.indexOf('transport=udp') !== -1;
+              return url.indexOf('turn:') === 0 &&
+                  url.indexOf('transport=udp') !== -1;
             })[0];
-            return true;
+            return !!server.urls;
           }
           return false;
         });


KaptenJansson:
LGTM

Tinet-AaronAn:
Thanks fippo for the reply, so i can not use stun? must use turn for Edge?

Tinet-AaronAn:
Micosoft msdn said that they support stun.

https://msdn.microsoft.com/zh-cn/library/mt502501(v=vs.85).aspx

fippo:
yet this still does not work and results in an InvalidAccessError:

```
var g = new RTCIceGatherer({iceServers:[{urls: "stun:stun.l.google.com:19302?transport=udp"}]})
```

Tinet-AaronAn:
hi fippo, I got this work on new RTCPeerConnection, but it dont get the correct public ip addr.
iceOptions.gatherPolicy= 'all';
iceOptions.iceServers = {urls: "stun:stun.l.google.com:19302"}; //not array but just object.

fippo:
STUN is not supported so if you just supply a STUN server no ice servers will be used and you don't get a public ip. No way around this, use a TURN server.

Tinet-AaronAn:
Thanks a lot, I have gave up run webRTC on Edge.
By the way, I use sipjs (sipjs.com) and it works on chrome very well.

PR #267, opened 2016-04-14
Edge: if mid is not defined, choose a random one

fippo:
since I saw @alvestrand merge https://github.com/w3c/webrtc-pc/pull/535 earlier today... ;-)
(wondering if that should also apply to rejected m-lines but _shrug_)


KaptenJansson:
LGTM

PR #269, opened 2016-04-15
Make a deep copy of constraints before applying any browser-specific …

jnoring:
**Description**

If we don't make a deep copy of constraints submitted to getUserMedia, then the supplied constraints are modified to that browser's desired constraint format, be they spec compliant or no.  For example, after calling getUserMedia on Chrome, constraints.audio.deviceId becomes constraints.audio.mandatory.sourceId

**Purpose**

It's a party foul to modify input arguments unless the spec mandates it.  I don't believe it does.

### DIFF
--- a/src/js/chrome/getusermedia.js
+++ b/src/js/chrome/getusermedia.js
@@ -63,14 +63,15 @@ module.exports = function() {
   };
 
   var getUserMedia_ = function(constraints, onSuccess, onError) {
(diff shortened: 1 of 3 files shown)


jnoring:
I am covered by a corporate agreement with my employer, HireVue.  Let me know if I need to supply any additional info, or sign an individual CLA.

PR #272, opened 2016-04-18
This MR could be made much more simple by changing the object to whic…

jnoring:
Simplify my previous MR based on feedback from @jan-ivar


KaptenJansson:
LGTM (I missed this too ;))

PR #276, opened 2016-04-20
bump to 1.1.0

KaptenJansson:
**Description**

**Purpose**


fippo:
since this adds safari, isn't it minor at least?

KaptenJansson:
Good point! 2nd time using npm version and did not reflect the version

PR #277, opened 2016-04-20
Shim maplike getStats support.

jan-ivar:
**Description**

Shims maplike getStats support in the three browsers, while still being backwards compatible.

**Purpose**

Firefox 48 [adds](https://bugzil.la/1213056) maplike getStats support, so here is a shim that lets adapter.js users use this right away without worry.

Note: had some trouble testing this in Edge, but trouble seemed unrelated to this patch, so I put this up to see if tests succeed there.

### DIFF
--- a/src/js/chrome/chrome_shim.js
+++ b/src/js/chrome/chrome_shim.js
@@ -130,9 +130,19 @@ var chromeShim = {
           return standardReport;
         };
 
+        // shim getStats with maplike support
+        var makeMapStats = (stats, legacyStats) => {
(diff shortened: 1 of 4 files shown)


fippo:
I can test in Edge later @jan-ivar. No automated tests yet.

\o/

KaptenJansson:
LGTM. (FYI ES 6 syntax template literals is fine due to all the affected browser + node js support it, we just need to be careful in adapter-core.js since it has to run in all browsers™)

PR #285, opened 2016-05-05
improve promise handling when callback method signature is used

xdumaine:
**Description**

When callback functions (success and error) are provided, invoke these using `promise.then` and `promise.catch` instead of at resolve/reject time.

**Purpose**

By putting these into `then` and `catch` you can ensure that the promise chain is handled by the error handler, instead of causing `Uncaught error in promise chain` errors. I.e., If the promise is rejected, but it's handled by a callback, it should be done in the catch.


KaptenJansson:
LGTM!

KaptenJansson:
Thanks for this btw!

PR #286, opened 2016-05-05
Edge: ice-lite support

fippo:
**Description**
this adds ice-lite support. Used to work with @ggarber's SFU and still does with mine.

Hard to test without an SFU (or other server) unfortunately.


KaptenJansson:
LGTM

PR #287, opened 2016-05-09
Fix Chrome WebRTC promise-poly to never reject dummy promise from legacy cb calls

jan-ivar:
**Description**
Fixes https://github.com/w3c/webrtc-pc/issues/615#issuecomment-216864923.

**Purpose**
Makes sure legacy code doesn't trigger unhandled promise detection code on properly handled failure.


KaptenJansson:
cc @fippo

KaptenJansson:
LGTM

PR #292, opened 2016-05-12
shim facingMode in Chrome

jan-ivar:
**Description**
Shim "user" and "environment" facingMode constraint in Chrome for mobile.

**Purpose**
I got tired of pointing people in the wrong direction on this on SO. :)

### DIFF
--- a/src/js/chrome/getusermedia.js
+++ b/src/js/chrome/getusermedia.js
@@ -62,17 +62,49 @@ module.exports = function() {
     return cc;
   };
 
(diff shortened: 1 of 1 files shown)


KaptenJansson:
Awesome, LGTM although I'm a bit hesitant to the ECMA6 use but I guess most "Chrome" based browsers does accept it. If not, people will let us know (as long as we do not use it in the generic pieces of adapter which is always executed regardless of the browser).

PR #293, opened 2016-05-14
Shim NotAllowedError.

jan-ivar:
**Description**
Shim getUserMedia to fail with NotAllowedError (Chrome, Firefox) and OverconstrainedError (Chrome), to match the spec.

**Purpose**
Make it easier to handle errors across browsers and versions uniformly.

### DIFF
--- a/src/js/chrome/getusermedia.js
+++ b/src/js/chrome/getusermedia.js
@@ -115,7 +115,7 @@ module.exports = function() {
   var getUserMedia_ = (constraints, onSuccess, onError) =>
     shimConstraints_(constraints,
                      c => navigator.webkitGetUserMedia(c, onSuccess,
-                         e => Promise.reject(shimError_(e))));
+                         e => onError(shimError_(e))));
 
   navigator.getUserMedia = getUserMedia_;
 
--- a/src/js/chrome/getusermedia.js
+++ b/src/js/chrome/getusermedia.js
@@ -101,9 +101,21 @@ module.exports = function() {
     return func(constraints);
   };
 
+  var shimError_ = e => ({
+    name: {PermissionDeniedError: 'NotAllowedError',
+           ConstraintNotSatisfiedError: 'OverconstrainedError'
+          }[e.name] || e.name,
+    message: e.message,
+    constraint: e.constraintName,
+    toString: function() {
+      return this.name + (this.message && ': ') + this.message;
+    }
+  });
+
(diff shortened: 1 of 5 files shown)


jan-ivar:
Need to test Edge to see if it needs a shim as well.

src/js/chrome/getusermedia.js
```diff
@@ -101,9 +101,21 @@ module.exports = function() {
     return func(constraints);
   };
 
+  var shimError_ = e => ({
+    name: {PermissionDeniedError: 'NotAllowedError',
+           ConstraintNotSatisfiedError: 'OverconstrainedError'
+          }[e.name] || e.name,
+    message: e.message,
+    constraint: e.constraintName,
+    toString: function() {
```
KaptenJansson, line 110: I'm curious to why you did not use an arrow function here?

src/js/firefox/getusermedia.js
```diff
@@ -112,4 +129,10 @@ module.exports = function() {
       });
     };
   }
+  if (browserDetails.version < 49) {
+    var origGetUserMedia = navigator.mediaDevices.getUserMedia.
```
KaptenJansson, line 133: I would rather have the dot on the new line but since we do not have linting in place for this I will fix it afterwards when I add the new [rule](http://eslint.org/docs/rules/dot-location).

KaptenJansson:
LGTM

fippo:
@jan-ivar: what Edge version did you test with? Might be worth fixing upstream, @shijuns might be able to arrange this ;-)

src/js/chrome/getusermedia.js
jan-ivar, line 110: I didn't want to assume strict mode. http://stackoverflow.com/questions/36717376/arrow-function-in-object-literal

Are we guaranteeing adapter.js will always use strict?

jan-ivar:
I tested Edge 25.10586.0.0, which appears to use `PermissionDeniedError`.

ShijunS:
@fippo, @Jan-Ivar, We haven't made much change to the Edge implementation recently.  Now with the spec being finalized soon, we will catch up in the next product cycles in a few areas including the error event and error code.

src/js/chrome/getusermedia.js
KaptenJansson, line 110: I'm pretty sure our linters (and browsers) will let us know if we do not follow strict conventions since we have 'use strict' at the top of each file.

PR #266, opened 2016-04-11
Shim away RTCSessionDescription/RTCIceCandidate.

jan-ivar:
**Description**
Shim implicit creation of (now redundant) `RTCSessionDescription`/`RTCIceCandidate` interfaces in `setLocalDescription`/`setLocalDescription`/`addIceCandidate`.

**Purpose**
https://github.com/w3c/webrtc-pc/pull/302. I started fixing this in Firefox, then realized people wouldn't benefit in years without having a polyfill.

PR #294, opened 2016-05-15
add addIceCandidate(null)

fippo:
**Description**
this allows calling addIceCandidate with the null candidate, a relatively recent spec change.

@taylor-b you might know the issue number in chrome
@nils-ohlmeier same for mozilla

**Purpose**
spec-compliant code, today :-)

Detailed comments inline

### DIFF
--- a/src/js/chrome/chrome_shim.js
+++ b/src/js/chrome/chrome_shim.js
@@ -230,7 +230,7 @@ var chromeShim = {
     var nativeAddIceCandidate =
         RTCPeerConnection.prototype.addIceCandidate;
     RTCPeerConnection.prototype.addIceCandidate = function() {
-      return arguments[0] === null ? Promise.resolve
+      return arguments[0] === null ? Promise.resolve()
           : nativeAddIceCandidate.apply(this, arguments);
     };
   },
--- a/src/js/chrome/chrome_shim.js
+++ b/src/js/chrome/chrome_shim.js
@@ -225,6 +225,14 @@ var chromeShim = {
             };
           });
     }
+
+    // support for addIceCandidate(null)
+    var nativeAddIceCandidate =
+        RTCPeerConnection.prototype.addIceCandidate;
+    RTCPeerConnection.prototype.addIceCandidate = function() {
+      return arguments[0] === null ? Promise.resolve
+          : nativeAddIceCandidate.apply(this, arguments);
+    };
   },
 
   // Attach a media stream to an element.
--- a/src/js/edge/edge_shim.js
+++ b/src/js/edge/edge_shim.js
@@ -905,38 +905,44 @@ var edgeShim = {
     };
 
     window.RTCPeerConnection.prototype.addIceCandidate = function(candidate) {
-      var mLineIndex = candidate.sdpMLineIndex;
-      if (candidate.sdpMid) {
-        for (var i = 0; i < this.transceivers.length; i++) {
-          if (this.transceivers[i].mid === candidate.sdpMid) {
-            mLineIndex = i;
-            break;
+      if (candidate === null) {
+        this.transceivers.forEach(function(transceiver) {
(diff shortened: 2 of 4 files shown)


test/test.js
```diff
@@ -1332,19 +1332,16 @@ test('Basic connection establishment', function(t) {
...
-            // TODO: Decide if we are interested in adding all candidates
-            // as passed tests.
-            tc.pass('addIceCandidate ' + counter++);
-          },
-          function(err) {
-            tc.fail('addIceCandidate ' + err.toString());
-          }
-        );
-      }
+      pc.addIceCandidate(event.candidate,
```
fippo, line 1335: we did not use promises here on purpose, right?

src/js/chrome/chrome_shim.js
```diff
@@ -225,6 +225,14 @@ var chromeShim = {
             };
           });
     }
+
+    // support for addIceCandidate(null)
```
fippo, line 227: I decided not to mix this with the promise-ification. Which we might have to remove soon for canary?
Maybe we should separate https://github.com/webrtc/adapter/pull/266/files as well?

test/test.js
KaptenJansson, line 1335: I cannot remember if it was on purpose or just out of habit. I assume most are not using promises for this yet?

src/js/chrome/chrome_shim.js
KaptenJansson, line 227: SGTM on both counts. Separate PR though.

src/js/edge/edge_shim.js
```diff
@@ -905,38 +905,44 @@ var edgeShim = {
...
-        // A dirty hack to make samples work.
-        if (cand.type === 'endOfCandidates') {
-          cand = {};
-        }
-        transceiver.iceTransport.addRemoteCandidate(cand);
+        var transceiver = this.transceivers[mLineIndex];
+        if (transceiver) {
+          var cand = Object.keys(candidate.candidate).length > 0 ?
+              SDPUtils.parseCandidate(candidate.candidate) : {};
+          // Ignore Chrome's invalid candidates since Edge does not like them.
```
KaptenJansson, line 926: Invalid candidates? Should we file a bug?

fippo, line 926: the tcp ones with port 0. Those are gone now I think.

KaptenJansson, line 926: Aha right, yes that is fixed.

KaptenJansson:
LGTM

test/test.js
fippo, line 1335: yeah. which makes it important to test the non-promise versions. So lets assume this was a conscious choice

KaptenJansson, line 1335: SGTM

nils-ohlmeier:
No tickets existed yet, so I created https://bugzilla.mozilla.org/show_bug.cgi?id=1273319

taylor-b:
Just a heads up: this may change in the spec at some point. There's an ambiguous situation where if addIceCandidate is called in the middle of an ICE restart, you don't know if it's a post-restart or pre-restart candidate.

The ICE working group decided to trickle the ufrag to solve this ambiguity. But this ambiguity also exists with a trickled "end-of-candidates". So we could end up putting a ufrag on the end-of-candidates indication, meaning it won't be a simple "null" any longer.

See the discussion here: https://github.com/rtcweb-wg/jsep/issues/251#issuecomment-195171292

fippo:
@taylor-b gee. How about checking for !arguments[0].candidate? That seems safe no matter what is going to be the result of the ufrag/pwd discussion.

taylor-b:
I think this is good for now; it's possible the API won't even change. I just wanted to keep you in the loop.

fippo:
sounds good. I think this will allow doing the right thing in the samples -- just forward all the output of onicecandidate to the peer and feed it into addIceCandidate.

PR #288, opened 2016-05-10
use native RTCPeerConnection promises in Chrome 51

fippo:
**Description**
does not attempt to shim promises in Chrome 51+

**Purpose**
make adapter superflous

PR #295, opened 2016-05-16
shim away RTCSessionDescription/RTCIcecandidate

fippo:
basically #266 but separate so it doesn't collide with #288
This is still required in Chrome 51, even though not if promises are used. See [here](https://bugs.chromium.org/p/chromium/issues/detail?id=612118)

cc @jan-ivar, sorry for breaking this


KaptenJansson:
LGTM. One thing though, we should have test that catch this. I will start looking at separating the tests into separate files per API since it's way too big now.

Issue #297, opened 2016-05-17
SyntaxError: Unexpected token '>' (11:01:31:703 | error, javascript)   at adapter-latest.js:726


fatalfeel:
http://webrtc.github.io/adapter/adapter-latest.js

SyntaxError: Unexpected token '>' (11:01:31:703 | error, javascript)
  at adapter-latest.js:726

KaptenJansson:
Hi,

Where are you running this? It cannot run a ECMA 6 arrow functíon.

Is your browser/environment not ECMA 6 capable?

fatalfeel:
firefox 41

KaptenJansson:
Why are you using a browser from September 2015? Even FF ESR is on version 45. Please update.

cc @jan-ivar

fatalfeel:
I am 32 bits windows system the lasetest version is v43 /

KaptenJansson:
No, that is not true according to https://www.mozilla.org/en-US/firefox/46.0.1/system-requirements/.

KaptenJansson:
Closing issue due to using ancient browser.

fatalfeel:
SyntaxError: Unexpected token '>' (11:01:31:703 | error, javascript) fixed
///
but failed here
change to adapter.js or adapter-latest.js both failed too  wss can not work on firefox v46
https://picasaweb.google.com/106185541018774360364/IMx6#6285590650062588018

at chrome is ok
https://picasaweb.google.com/106185541018774360364/IMx6#6285592528799490258

davies147:
I agree that suddenly using ECMA6 '=>' arrow functions is a bit out of
order. A useful function of adapter.js has always been to load adapter.js
to determine whether you can use WebRTC. There are companies running IE10,
IE11 etc out there because they have strict upgrade policies internally
(yes it might be stupid, but it happens).

IMHO adapter.js should be able to load on IE, older Firefox, and older
Chrome, and will fail gracefully rather than throwing a compile-time
exception. Perhaps a solution to this is to add some 'babel'ified targets
for people who need them?

At the very least the README.md should warn that backward compatibility
just got thrown out?

Cheers,
Steve

KaptenJansson:
Yes I agree. I did not realize the impact it would have. Will merge @fippo's fix.

KaptenJansson:
This has now been fixed. Thanks for your feedback. Will look into adding babelified targets soonish for legacy browsers.

fatalfeel:
[...]

PR #299, opened 2016-05-17
dont use arrow functions (yet)

fippo:
**Description**
ES6 breaks things like #297. While I'm in favor of doing that we should make that change at a major version, not a minor one.

@jan-ivar: don't worry, there is a [codemod to arrow-ify things](https://github.com/cpojer/js-codemod/blob/master/transforms/arrow-function.js) ;-)

### DIFF
--- a/src/js/chrome/chrome_shim.js
+++ b/src/js/chrome/chrome_shim.js
@@ -136,10 +136,12 @@ var chromeShim = {
         };
 
         // shim getStats with maplike support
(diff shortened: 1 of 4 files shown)


KaptenJansson:
@jan-ivar could you take a look at this as well?

KaptenJansson:
LGTM (will address comments later if needed, just want to get this in ASAP (all adapter + samples test pass with this and code looks code)).

jan-ivar:
lgtm.

PR #301, opened 2016-05-17
Crude initial attempt to create ES5 safe targets

davies147:
Add 4 "duplicate" targets which are babelified while being browserified.

This is in order to automatically create safe versions for old browsers without restricting coders to pre-ES6 sources.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -54,9 +54,71 @@ module.exports = function(grunt) {
           ignore: [
             './src/js/edge/edge_shim.js',
             './src/js/edge/edge_sdp.js'
-          ],
+          ]
         }
       },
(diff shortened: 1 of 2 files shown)


davies147:
I signed it!

KaptenJansson:
Awesome ;) Thanks for contributing.

One question for the users requiring this. The initial report was from someone using adapter-latest.js (it's actually a symlink to adapter.js) served on the gh-pages branch. Should we add more variants there with ES5 appended to the name?

fippo:
I think we should leave adapter-latest as transpiled es5 for now. Thank you @davies147!

KaptenJansson:
LGTM

PR #311, opened 2016-05-25
Remove es6 arrow functions, add specific babel preset and npm 3+ is now required

### LINKED ISSUE #309: Package Babel presets conflict with local Babel presets
My build system (Gulp), throws the error ' Couldn't find preset "es2015" ', as a result of the babel preset specified in package.json og adapter.js . It can't be found because I am using es2015-without-strict.

I am using NPM to get the package, and only want to use a prebuilt version.

I don't use grunt so haven't been able to test but think this could be resolved by moving the preset declaration inline.  e.g. transform: [['babelify', {"presets": ["es2015"]}]]

KaptenJansson:
**Description**
- Be ES5 compliant since do not need any ES6 features yet (nice to have to make code more compact but not needed).
- Indicate that we expect npm to be at least 3.0.0
- Add specific babel preset to ensure better compatibility when included in other projects. Fixes #309 

**Purpose**
Make adapter.js easier to use.
…

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -58,7 +58,7 @@ module.exports = function(grunt) {
           browserifyOptions: {
             // Exposes shim methods in a global object to the browser.
             standalone: 'adapter',
-            transform: 'babelify'
+            transform: [['babelify', {'presets': ['es2015']}]]
           }
         }
       },
@@ -69,7 +69,7 @@ module.exports = function(grunt) {
         dest: './out/adapter_no_global_es5.js',
         options: {
           browserifyOptions: {
-            transform: [['babelify', {"presets": ["es2015"]}]]
+            transform: [['babelify', {'presets': ['es2015']}]]
           }
         }
       },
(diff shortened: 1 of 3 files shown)


KaptenJansson:
There we go, travis woke up ;)

KaptenJansson:
@fippo @jan-ivar FYI (lots of other projects not being es6 capable are failing now)

Gruntfile.js
```diff
@@ -69,7 +69,7 @@ module.exports = function(grunt) {
         dest: './out/adapter_no_global_es5.js',
         options: {
           browserifyOptions: {
-            transform: 'babelify'
+            transform: [['babelify', {"presets": ["es2015"]}]]
```
samdutton: Single quotes? 

`'presets': ['es2015']`

KaptenJansson: Doh ;) Sorry about that.

samdutton:
LGTM.

PR #312, opened 2016-05-25
Move 'babelify' settings into grunt file to isolate from outside world

davies147:
Move 'babelify' settings into grunt file to isolate from outside world

It is reported that putting babelify settings into package.json can 'contaminate' other NPM projects, so the settings are moved inline into the grunt config.


davies147:
Tested this change briefly, and on my local setup it seems to "change nothing" - Would appreciate a double check from elsewhere.

Hope the format of the pull request is clean. It seemed to want to take my merge along with it for some reason.

davies147:
OK, looks like I was beaten to it, so ignore/reject this :)

PR #313, opened 2016-05-26
Fix publish copy routine, use out/adapter.js as adapter.js rather than es5

KaptenJansson:
**Description**
Remove global babel preset since it breaks builds, now uses only Gruntfile.js presets.
Use out/adapter.js for gh-pages rather than the es5 target since we should now be es5 complaint regardless since just creating es5 compliant targets does not help when trying to build es6 code to es5 compliant targets when build environments only support es5 (true for projects having adapter.js as a dependency)

**Purpose**
Make adapter.js universally usable ;)

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,18 +8,13 @@
     "type": "git",
     "url": "https://github.com/webrtc/adapter.git"
   },
-  "babel": {
-    "presets": [
-      "es2015"
-    ]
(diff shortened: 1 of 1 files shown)


KaptenJansson:
Merging this in order to fix publishing. No code changes per see.

PR #316, opened 2016-05-26
move, not copy, getusermedia to separate file.

jan-ivar:
**Description**
Remove duplicate `firefoxShim.shimGetUserMedia`.

**Purpose**
Less code, and the duplicate was confusing people.


KaptenJansson:
I wonder why I hadn't spotted this before ;). Thanks for this. LGTM.

PR #322, opened 2016-05-28
allow ice-lite without candidates

fippo:
i'm not sure if 5245 really allows ice-lite with no candidates and using trickle.
Previously, my interpretation was 'no'. On what basis does hangouts combine trickle and lite @juberti @pthatcherg?


KaptenJansson:
CC @taylor-b

taylor-b:
I don't know why an ICE lite endpoint would need to trickle candidates. But I don't see any problem with allowing it.

KaptenJansson:
OK, LGTM.

PR #326, opened 2016-06-02
support for BUNDLE when using max-bundle

fippo:
I felt an urge to be somewhat compatible with applications that set max-bundle.

This is a laymans interpretation of the BUNDLE and doesn't claim compliance with any of the 29 versions of the draft. It is really more a "just use a single ice and dtls transport".


KaptenJansson:
LGTM

PR #328, opened 2016-06-07
allow urls to be a string

### LINKED ISSUE #327: Edge not supporting `filter`

Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/46.0.2486.0 Safari/537.36 Edge/13.10586

**Description**

Above user agent causes JS error:

```
Object doesn't support property or method 'filter'
```

At this line:

https://github.com/webrtc/adapter/blob/d6e8b1a45add33f382fed872f32908ea225a1996/src/js/edge/edge_shim.js#L97

**Steps to reproduce**

Create webrtc peer connection with an edge browser and give it ice servers.

**Expected results**

No js error. Perhaps a for loop?

**Actual results**

JS error thrown above.

p.s. thank you very much for this library! If you need anything else, let me know.

fippo:
**Description**
fixes #327 -- much as I think the spec is doing a bad thing by allowing this :-)

### DIFF
--- a/src/js/edge/edge_shim.js
+++ b/src/js/edge/edge_shim.js
@@ -94,13 +94,18 @@ var edgeShim = {
         // Edge does not like
         // 1) stun:
         // 2) turn: that does not have all of turn:host:port?transport=udp
-        this.iceOptions.iceServers = config.iceServers.filter(function(server) {
+        var iceServers = JSON.parse(JSON.stringify(config.iceServers));
+        this.iceOptions.iceServers = iceServers.filter(function(server) {
           if (server && server.urls) {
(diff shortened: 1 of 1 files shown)


ngauthier:
Nice, thank you. Why the stringify and parse? Just curious :)

fippo:
deep-copy. Unrelated to the issue but i had just seen weird behaviour when modifying the config in ways the browser did not expect.

KaptenJansson:
Yes modifying objects can cause unwanted behaviors especially if objects are reused. LGTM.

PR #329, opened 2016-06-11
dont use native promise for createOffer yet

fippo:
since promise-ified createOffer is broken... :-(
see https://bugs.chromium.org/p/chromium/issues/detail?id=619289

createAnswer is collateral damage. Seemed better than to split createOffer and createAnswer, I assume this is quite temporary.

### DIFF
--- a/src/js/chrome/chrome_shim.js
+++ b/src/js/chrome/chrome_shim.js
@@ -187,23 +187,23 @@ var chromeShim = {
       });
     }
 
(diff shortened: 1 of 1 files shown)


fippo:
cc @alvestrand

travis failure in Firefox... seems unrelated. Maybe kick it?

KaptenJansson:
LGTM.

Re Firefox, noticed this in other tests as well, seems like webdriver is not supporting latest stable :(

PR #332, opened 2016-06-16
move null candidate shim

fippo:
the order of shimming things is important. The Firefox shim does it right, Chrome does not.
The null candidate has to be ignored at the outermost layer of this little onion, otherwise the implicit creation shim tries to call RTCIceCandidate(null) which errors.


KaptenJansson:
LGTM

Issue #192, opened 2016-02-12
bump minversion and get rid of legacy code


fippo:
if we do a major bump anyway we should update the minimum versions supported. Currently this is 
- Firefox 31(released 2014-07-21)
- Chrome 38 (released 2014-10-07)
- Edge 10547 (insider-only version I think? 10587 is from Nov 2015 and the first stable version with ORTC support)

I think we can remove a fair amount of code that is no longer needed.
@jan-ivar: thoughts on a better Firefox version? Last ESR version maybe?
@KaptenJansson @alvestrand same for Chrome? Chrome 43 with the srcObject shim would be good imo

I don't think there are relevant changes in Edge between 10547 and 10587 which is the current stable so no need to bump that.

KaptenJansson:
+1 for removing old code ;). Let's settle on latest Chrome stable and FF esr then?

jan-ivar:
Discussed with @fippo, even though Firefox ESR 45 shipped yesterday, I agree we should probably support 38 (the previous ESR) for at least another month.

fippo:
fab. Will go for chrome 48 as min version. I don't see any shim code that was obsoleted between 38 and 48 unfortunately. Looking forward to the constraints changes.

fippo:
little point in doing this. Lets just remove adapter.js alltogether one day ;-)

PR #334, opened 2016-06-20
remove attachmediastream and reattachmediastream

fippo:
**Description**
this remove attachmediastream and reattachmediastream. When moving to 1.0 we hid them in such a ugly way that hopefully nobody uses them anyway.

This needs a major version bump. I think we should do #192 and support for the new native getUserMedia at the same time.


KaptenJansson:
LGTM, I've filed https://github.com/webrtc/apprtc/issues/322 to keep track since it uses them still.

PR #336, opened 2016-06-21
dont shim GUM if unprefixed gum exists

fippo:
lets see if the tests pass. This needs some more discussion before actually merging.

The tests that fail (locally) are conversion tests which i'd expect to fail.

### DIFF
--- a/src/js/chrome/getusermedia.js
+++ b/src/js/chrome/getusermedia.js
@@ -11,9 +11,6 @@ var logging = require('../utils.js').log;
 
 // Expose public methods.
 module.exports = function() {
-  if (navigator.getUserMedia) {
-    return;
-  }
   var constraintsToChrome_ = function(c) {
     if (typeof c !== 'object' || c.mandatory || c.optional) {
       return c;
@@ -120,15 +117,15 @@ module.exports = function() {
     };
   };
 
-  var getUserMedia_ = function(constraints, onSuccess, onError) {
-    shimConstraints_(constraints, function(c) {
-      navigator.webkitGetUserMedia(c, onSuccess, function(e) {
-        onError(shimError_(e));
+  if (!navigator.getUserMedia) {
+    navigator.getUserMedia = function(constraints, onSuccess, onError) {
+      shimConstraints_(constraints, function(c) {
+        navigator.webkitGetUserMedia(c, onSuccess, function(e) {
+          onError(shimError_(e));
+        });
       });
-    });
-  };
-
-  navigator.getUserMedia = getUserMedia_;
+    };
+  }
 
   // Returns the result of getUserMedia as a Promise.
   var getUserMediaPromise_ = function(constraints) {
--- a/src/js/chrome/getusermedia.js
+++ b/src/js/chrome/getusermedia.js
@@ -11,6 +11,9 @@ var logging = require('../utils.js').log;
 
 // Expose public methods.
 module.exports = function() {
+  if (navigator.getUserMedia) {
+    return;
+  }
   var constraintsToChrome_ = function(c) {
     if (typeof c !== 'object' || c.mandatory || c.optional) {
       return c;
--- a/test/test.js
+++ b/test/test.js
@@ -1251,8 +1251,7 @@ test('Check getUserMedia legacy constraints converter', function(t) {
     // decision if the test should be run or not is in the Test definition
     // rather than the preferred Run test section (Webdriver).
     // FIXME: Move the decision to // Run test.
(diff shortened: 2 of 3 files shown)


fippo:
uhm... travis?!

fippo:
https://travis-ci.org/webrtc/adapter/jobs/139182194 shows the canary output. This looks great!

KaptenJansson:
@alvestrand FYI

KaptenJansson:
So this means we have to disable the gUM constraints conversion tests for <=M52 versions, right?

alvestrand:
My instinctive reaction is that this decision belongs in the Chrome shim - that is, chromeShim.shimGetUserMedia() should be the one making the decision.

So this line from src/js/chrome/getusermedia.js:

 navigator.getUserMedia = getUserMedia_;

should become

if (!navigator.getUserMedia) {
   // getuserMedia was added in 53
   navigator.getUserMedia = getUserMedia_;
}

That way, we can keep stuff isolated if we discover ways in which the new getUserMedia require shimming.

alvestrand:
The constraints conversion tests should still pass - the old constraints are still supported, the tests explicitly target the prefixed "webkitGetUserMedia" interface, and the tests are targeted at testing the conversion functions, not how Chrome handles them.

fippo:
the tests target webkitGetUserMedia -- but the getUserMedia call and we do not shim navigator.webkitGetUserMedia (should we?)

Just returning isn't enough, this misses the error shim. Or did you already implement that in M53 as well?

KaptenJansson:
I do not think we should be shimming prefixed API's once they have been unprefixed.

KaptenJansson:
This LGTM now,  @alvestrand any concerns?

KaptenJansson:
Ping

KaptenJansson:
Are we ready to merge this ?

src/js/chrome/getusermedia.js
```diff
@@ -11,6 +11,9 @@ var logging = require('../utils.js').log;
 
 // Expose public methods.
 module.exports = function() {
+  if (navigator.getUserMedia) {
+    return;
+  }
```
alvestrand: Keep an eye on this one. It turns of _all_ shimming if getUserMedia is detected unprefixed, including the error shim. Would be nice if it's correct.

test/test.js
```diff
@@ -1251,7 +1251,8 @@ test('Check getUserMedia legacy constraints converter', function(t) {
     // decision if the test should be run or not is in the Test definition
     // rather than the preferred Run test section (Webdriver).
     // FIXME: Move the decision to // Run test.
-    if (window.adapter.browserDetails.browser === 'chrome') {
+    if (window.adapter.browserDetails.browser === 'chrome'
+        && window.adapter.browserDetails.browserVersion < 53) {
```
alvestrand: This needs to move. The test for browserVersion < 53 should be specifically around the back-conversion of new-style constraints; the test for old constraints passing through unharmed still needs to be run.
We also need a new test - that new-style constraints are _not_ converted if browserVersion >= 53.

alvestrand:
Sorry for being slow.

dandv:
:+1:

I'm building an [ImageCapture takePhoto()](https://www.w3.org/TR/image-capture/#dom-imagecapture-takephoto) polyfill and calling the [Promise-style MediaDevices.gUM() API](https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/getUserMedia) results in an odd error, given than Chrome 53 supports `navigator.getUserMedia()` unprefixed:

``` js
navigator.getUserMedia({video: true}).then(mediaStream => {
    ...
}).catch(error => {
    ...
});
```

The error is "Uncaught TypeError: Failed to execute **_'webkitGetUserMedia'**_ on 'Navigator': The callback provided as parameter 2 is not a function.".

http://jsbin.com/yerojuhudu/edit?html,css,output

fippo:
@dandv navigator.getUserMedia doesn't do promises (which is one of the reasons it is deprecated)

dandv:
@fippo: right, my point was that the error was surprising in that it mentioned `webkitGetUserMedia` instead of `navigator.getUserMedia`.

fippo:
@dandv until this is merged (which is blocked by fixing the tests; i'm slow) we are overriding navigator.getUserMedia with the legacy webkit thing

KaptenJansson:
what's the latest here?

fippo:
i don't have the cycles to fix the tests right now :-/

fippo:
also removing the constraints shim doesn't seem prudent, see #408

PR #337, opened 2016-06-21
edge: fix bad assumptions about order of tracks

fippo:
if the local stream had and audio and a video track but
the sdp was video-only this created the rtpsender with an audio
track which led to funny errors when calling send. This fix still
has bad assumptions but is more explicit about being limited at least.


KaptenJansson:
LGTM

PR #338, opened 2016-07-01
remove minVersion

fippo:
minVersion is not really usable right now. The original plan was to allow
rejecting old browser versions based on this. However, this doesn't work.
Suppose we reject Chrome-based things < Chrome 50. Well...
Opera is at version 38 right now and gets detected as 38. So this would
just block all Opera users! Platform detection is something the app needs to
do unfortunately unless we want to have more complicated rules here (e.g.
by doing more UA string parsing using the platform module)

### DIFF
--- a/src/js/utils.js
+++ b/src/js/utils.js
@@ -49,7 +49,7 @@ var utils = {
   /**
    * Browser detector.
    *
-   * @return {object} result containing browser, version and minVersion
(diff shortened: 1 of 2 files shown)


KaptenJansson:
I agree, can you also remove the test?

KaptenJansson:
LGTM.

PR #341, opened 2016-07-16
Edge: determine common rtcp feedback

fippo:
This improves getCommonCapabilities to also determine the intersection of .rtcpFeedback.

I have a minimal test in https://jsfiddle.net/h5bjgjxk/ but am not sure where to put it... :-/


KaptenJansson:
LGTM.

Re tests, hmm yes, maybe create a separate edge test file?

PR #342, opened 2016-07-22
edge: filter chrome tcp candidates with port 9

fippo:
to avoid adding candidates with the same ip/port for different m-lines


KaptenJansson:
LGTM

PR #343, opened 2016-07-24
gUM's provided media MUST include one track of each requested type.

jan-ivar:
**Description**
Shim gUM in Chrome and Firefox 48- to fail unless tracks exist for all requested media types.

**Purpose**
Comply with spec.

### DIFF
--- a/src/js/chrome/getusermedia.js
+++ b/src/js/chrome/getusermedia.js
@@ -167,7 +167,16 @@ module.exports = function() {
         bind(navigator.mediaDevices);
     navigator.mediaDevices.getUserMedia = function(cs) {
       return shimConstraints_(cs, function(c) {
-        return origGetUserMedia(c).catch(function(e) {
+        return origGetUserMedia(c).then(function(stream) {
+          if (c.audio && !stream.getAudioTracks().length ||
+              c.video && !stream.getVideoTracks().length) {
+            stream.getTracks().forEach(function(track) {
+              track.stop();
+            });
+            throw new DOMException('', 'NotFoundError');
+          }
+          return stream;
+        }, function(e) {
           return Promise.reject(shimError_(e));
         });
       });
--- a/src/js/chrome/getusermedia.js
+++ b/src/js/chrome/getusermedia.js
@@ -167,13 +167,7 @@ module.exports = function() {
         bind(navigator.mediaDevices);
     navigator.mediaDevices.getUserMedia = function(cs) {
       return shimConstraints_(cs, function(c) {
-        return origGetUserMedia(c).then(function(stream) {
-          if (c.audio && !stream.getAudioTracks().length ||
-              c.video && !stream.getVideoTracks().length) {
-            throw new DOMException('', 'NotFoundError');
-          }
-          return stream;
-        }, function(e) {
+        return origGetUserMedia(c).catch(function(e) {
           return Promise.reject(shimError_(e));
         });
       });
--- a/src/js/chrome/getusermedia.js
+++ b/src/js/chrome/getusermedia.js
@@ -167,7 +167,13 @@ module.exports = function() {
         bind(navigator.mediaDevices);
     navigator.mediaDevices.getUserMedia = function(cs) {
       return shimConstraints_(cs, function(c) {
-        return origGetUserMedia(c).catch(function(e) {
+        return origGetUserMedia(c).then(function(stream) {
+          if (c.audio && !stream.getAudioTracks().length ||
+              c.video && !stream.getVideoTracks().length) {
+            throw new DOMException('', 'NotFoundError');
+          }
+          return stream;
+        }, function(e) {
           return Promise.reject(shimError_(e));
         });
       });
--- a/src/js/firefox/getusermedia.js
+++ b/src/js/firefox/getusermedia.js
@@ -133,7 +133,15 @@ module.exports = function() {
     var origGetUserMedia = navigator.mediaDevices.getUserMedia.
         bind(navigator.mediaDevices);
     navigator.mediaDevices.getUserMedia = function(c) {
-      return origGetUserMedia(c).catch(function(e) {
+      return origGetUserMedia(c).then(function(stream) {
+        // Work around https://bugzil.la/802326
+        if (c.audio && !stream.getAudioTracks().length ||
+            c.video && !stream.getVideoTracks().length) {
+          throw new DOMException('The object can not be found here.',
+                                 'NotFoundError');
+        }
+        return stream;
+      }, function(e) {
         return Promise.reject(shimError_(e));
       });
     };
--- a/src/js/firefox/getusermedia.js
+++ b/src/js/firefox/getusermedia.js
@@ -137,6 +137,9 @@ module.exports = function() {
         // Work around https://bugzil.la/802326
         if (c.audio && !stream.getAudioTracks().length ||
             c.video && !stream.getVideoTracks().length) {
+          stream.getTracks().forEach(function(track) {
+            track.stop();
+          });
           throw new DOMException('The object can not be found here.',
                                  'NotFoundError');
         }


fippo:
i'm worried that this will have a large unexpected impact on applications. Fortunately, I can tell you how big:

```
select count(*), calledgetusermediarequestingaudio as wantaudio, 
    calledgetusermediarequestingvideo as wantvideo, 
    firstaudiotracklabel != '' as gotaudio,
    firstvideotracklabel != '' as gotvideo
    from dataset
    where getusermediasuccess = 't' and getusermediaerror = '0'
    group by calledgetusermediarequestingaudio, calledgetusermediarequestingvideo, 
    firstaudiotracklabel != '', firstvideotracklabel != '';
```

Basically I query how many GUM calls from a 100k call dataset of calls that requested audio and video (without running into any error) and look whether I got a audio/video track label.

```
 count | wantaudio | wantvideo | gotaudio | gotvideo 
-------+-----------+-----------+----------+----------
 88867 | t         | t         | t        | t
  1706 | t         | t         | f        | t
  9427 | t         | t         | t        | f
```

This would break the current behaviour in 11% of the cases. Which is a lot.

I don't particularly care for appear.in, we try again with different constraints when getting a NotFoundError.

KaptenJansson:
@alvestrand WDYT?

src/js/chrome/getusermedia.js
```diff
@@ -167,7 +167,13 @@ module.exports = function() {
         bind(navigator.mediaDevices);
     navigator.mediaDevices.getUserMedia = function(cs) {
       return shimConstraints_(cs, function(c) {
-        return origGetUserMedia(c).catch(function(e) {
+        return origGetUserMedia(c).then(function(stream) {
+          if (c.audio && !stream.getAudioTracks().length ||
+              c.video && !stream.getVideoTracks().length) {
+            throw new DOMException('', 'NotFoundError');
```
fippo, line 176: don't you need to stop the tracks?

KaptenJansson, line 176: Well spotted, yes if there is an audio but no video track, the audio track needs to be stopped and vice versa.

jan-ivar, line 176: @fippo Ooh, good one. Will fix.

jan-ivar:
> This would break the current behaviour in 11% of the cases. Which is a lot.

@fippo Good point, though do we have a way of knowing if these applications are otherwise successful? I mean, do developers expect and handle the absence of these tracks, or does the app fail?

Not sure what to do about this since the spec mandates it. We can try to change the spec I suppose, but we're running out of time for that as this is shipping in Firefox 49!

fippo:
@jan-ivar: I don't know, don't you have telemetry on this? @KaptenJansson likewise, does Chrome gather stats?

Maybe we should send an email to discuss-webrtc, asking people how/if this would affect them?

jan-ivar:
@fippo thanks for the numbers. Is there a way to break those numbers down by browser?

Unlike Firefox (<49), Chrome doesn't offer users a choice of what to share, so the only case it returns a single stream is when a user is missing either a camera or microphone.

Also, 11% seems awfully high. My sql is a bit rusty, are you positive this total is limited to requests for both?

We have telemetry in firefox, but only counts for media types requested, and separately for what plays, but no way to correlate sessions unfortunately, the way you are doing here. We'd basically need to have asked this specific question in telemetry ahead of time.

fippo:
@jan-ivar we always request both in appear.in -- no enumerateDevices yet. While we have a fallback strategy that was elimiated by the getusermediaerror = '0'. Note to self: data set limited by 
`create or replace view dataset as select * from features_import where getusermediasuccess = 't' and getusermediaerror = '0' order by date desc limit 100000;`

Earlier numbers grouped by browser type:

```
browsertype | count | wantaudio | wantvideo | gotaudio | gotvideo 
-------------+-------+-----------+-----------+----------+----------
 moz         |  6383 | t         | t         | t        | t
 moz         |   728 | t         | t         | f        | t
 moz         |  2568 | t         | t         | t        | f
 webkit      | 82484 | t         | t         | t        | t
 webkit      |  6859 | t         | t         | t        | f
 webkit      |   978 | t         | t         | f        | t
```

Total for Firefox is 9679, total for chrome 90321. So this would affect 8% of Chrome users but 34% of Firefox users?!

(and yeah, telemetry is not flexible enough to answer this. Asking the right questions here is really hard. Answering them with the right query even more so)

jan-ivar:
Wow, those are surprising numbers. Granted Firefox actually asks users and gives them a choice, whereas Chrome doesn't. But is that enough to explain a 26% difference? If so, then Firefox sure seems to have a lot of shy or cautious users!

fippo:
Taking a step back... I started gathering this data a while back to see what errors were most frequent. In Chrome, the second most common error was DevicesNotFoundError which seems to be thrown already when audio and video have been requested but not both are available? In that case we just need to shim the error code.

I have also seen a very small proportion of Firefox calls failing with NotFoundError when another GUM call succeeded, mostly in FF 49 and 50.

So maybe Chrome already behaves this way? @alvestrand @KaptenJansson?

jan-ivar:
@fippo My internal MBP cam is busted atm, so my system is microphone-only unless I plug in something USB, so I know for a fact that both Chrome and Firefox <49 give me an audio-only stream in response to a request for cam+mic in that configuration, no error (49+ gives me NotFoundError, following the spec).

> I have also seen a very small proportion of Firefox calls failing with NotFoundError when another GUM call succeeded, mostly in FF 49 and 50.

What does "when another GUM call succeeded" mean?

jan-ivar:
...yet when I try it this morning I get DevicesNotFoundError. I could have sworn it worked differently just a few days ago. Did a Chrome update just land on me?

jan-ivar:
OK, I'm not crazy. The DevicesNotFoundError is a bug in Chrome that confused me (and @fippo I think). On my MBP with busted camera (i.e. mic-only), I can do this:
1. Plug in USB camera, do gUM call, and then block permission.
2. Now gUM fails with `NotAllowedError`.
3. Pull out USB camera.
4. Now gUM fails with `DevicesNotFoundError` :-P

In contrast, when things are unblocked, I get no error and an audio-only stream, so this patch (and @fippo 's analysis) still seem valid.

fippo:
yes, i've seen these constraints
`{"audio":{"optional":[{"sourceId":"default"}]},"video":{"optional":[{"sourceId":"default"},{"minWidth":640},{"maxWidth":640},{"minHeight":480},{"maxHeight":480},{"minFrameRate":25},{"maxFrameRate":25}]}}`
result in a video-only stream: `{"id":"zl5ffKUjrCaEn2T126OqduwvQVI1LCX8SQY7","tracks":[{"id":"60696a65-d9e3-48a8-8c2f-6eb1481d582f","kind":"video","label":"A4 TECH USB2.0 PC Camera E (0ac8:3450)","readyState":"live"}]}`
(in Opera 38) which explains my numbers.

jan-ivar:
The test in my last comment is with `{ video: true, audio: true }` fyi.

fippo:
that mostly leaves us with the behaviour change and a couple of chrome bugs to file. We need to make a major version anyway e.g. because of #338 so lets do it.

KaptenJansson:
@alvestrand what is your take on it?

alvestrand:
I think this is correct. We're violating the spec, and adapter.js should hide it.

KaptenJansson:
LGTM. Should we do a patch, minor or major release for this? I think at least minor.

alvestrand:
Every patch should have a patch release number change. I assume we don't push every patch version to the NPM repo, but some people might pull it from our working repo, so that helps sanity.

Major version pushes to NPM means that people with default setups for NPM packages stop tracking the current release train. Which is reasonable if we break compatibility and don't want them to call, but rather stay on the old version until they read the release notes.

Do we have enough accumulated incompatibiltiy that we should set them loose?

KaptenJansson:
We did an incompatible push not long ago which got a major version. So guess no, if this is not deemed breaking, then patch will do.

PR #345, opened 2016-07-27
chrome/edge interop: hardcode chrome default ssrcs

fippo:
working around https://bugs.chromium.org/p/webrtc/issues/detail?id=4740
Guarded by an extra "is this chrome?", hoping that nobody else uses this ancient version of MSID.

I suspect this is causing NACKs and PLIs not to be sent/received properly which leads to pictures like [this](https://twitter.com/HCornflower/status/741706976703479812), [this](https://twitter.com/HCornflower/status/738079011750940672) and [this](https://twitter.com/HCornflower/status/736456618758574080).

Hard to tell if that actually fixes things, this was rare to begin with.

### DIFF
--- a/src/js/edge/edge_shim.js
+++ b/src/js/edge/edge_shim.js
@@ -524,6 +524,18 @@ var edgeShim = {
             recvEncodingParameters =
                 SDPUtils.parseRtpEncodingParameters(mediaSection);
 
+            // Chrome does announce its SSRCs for recvonly streams.
+            // https://bugs.chromium.org/p/webrtc/issues/detail?id=4740
+            if (recvEncodingParameters.length === 0
+                && SDPUtils.matchPrefix(sessionpart,
+                'a=msid-semantic: WMS').length) {
+              if (kind === 'audio') {
+                recvEncodingParameters.push({ssrc: 0xfa17fa17});
+              } else if (kind === 'video') {
+                recvEncodingParameters.push({ssrc: 0x1});
+              }
+            }
+
             var mid = SDPUtils.matchPrefix(mediaSection, 'a=mid:');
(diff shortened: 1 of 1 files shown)


fippo:
@pthatcherg do yo I really have to or will ya fix 4740 instead? (I assume that this will be required for the transition but...)

KaptenJansson:
@pthatcherg ping

src/js/edge/edge_shim.js
```diff
@@ -524,6 +524,18 @@ var edgeShim = {
             recvEncodingParameters =
                 SDPUtils.parseRtpEncodingParameters(mediaSection);
 
+            // Chrome does announce its SSRCs for recvonly streams.
+            // https://bugs.chromium.org/p/webrtc/issues/detail?id=4740
```
pthatcherg, line 528: What do you mean it "announces" the SSRCs for recvonly streams?  It's more like it hardcodes them.

src/js/edge/edge_shim.js
```diff
@@ -524,6 +524,18 @@ var edgeShim = {
...
+            if (recvEncodingParameters.length === 0
+                && SDPUtils.matchPrefix(sessionpart,
+                'a=msid-semantic: WMS').length) {
+              if (kind === 'audio') {
+                recvEncodingParameters.push({ssrc: 0xfa17fa17});
+              } else if (kind === 'video') {
+                recvEncodingParameters.push({ssrc: 0x1});
+              }
+            }
+
```
pthatcherg, line 538: These are the SSRCs Chrome will use to send RTCP RRs.   Why would you put those in the encoding parameters for receiving RTP?

PR #346, opened 2016-07-27
edge: filter ipv6 turn addresses

fippo:
@shijuns told me that the RTCIceGatherer doesn't support ipv6 turn servers (like the ones used by apprtc)


KaptenJansson:
LGTM

PR #347, opened 2016-07-29
remove support for ancient firefox versions

fippo:
if we do a major version bump I think we should remove some legacy code. This breaks compat with Firefox < 41.

### DIFF
--- a/src/js/firefox/firefox_shim.js
+++ b/src/js/firefox/firefox_shim.js
@@ -8,8 +8,6 @@
  /* eslint-env node */
 'use strict';
 
-var browserDetails = require('../utils').browserDetails;
-
 var firefoxShim = {
   shimOnTrack: function() {
     if (typeof window === 'object' && window.RTCPeerConnection && !('ontrack' in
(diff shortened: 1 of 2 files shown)


fippo:
@jan-ivar WDYT? Do you know what versions we would loose if we removed https://github.com/webrtc/adapter/blob/master/src/js/firefox/getusermedia.js#L101-L118 as well?

TODO: remove all the constraint rewriting tests (which probably fail right now).

jan-ivar:
@fippo How do version bumps (for devs) correlate with dropping backwards support (for users)?

We added `navigator.mediaDevices.getUserMedia` in 36, `enumerateDevices` in 39.

41 seems aggressive to me. Backwards compatibility is one of the advantages of adapter.js.

I was unable to pull internal numbers atm, but going by http://caniuse.com/usage-table I see a significant tail, which translates into millions of users who might stumble across (hopefully working) WebRTC.

fippo:
@jan-ivar: we can't do it in a way which will automatically have people pull this in because it might break things which has to be a conscious decision.

There is a big difference between webrtc and other features: old versions of webrtc will offer a much lower user experience in terms of audio/video quality. I actually did look at our numbers and we have 0.38% of Firefox users who would be affected by this. Which includes version numbers all the way down to FF27. I am quite sure there is a large number of bugs in those versions which we do not work around.

But yeah, this needs discussion :+1:

jan-ivar:
@fippo I understand developers can't pull in new breaking versions automatically, but that's **_API**_ breakage, which has nothing to do with how many browser versions back will continue to work for users once the developer pulls it in. The lowest user experience is not working at all.

Are you saying 99.62% of Firefox users are on 41 or newer? Not sure I believe that, and it doesn't match http://caniuse.com/usage-table. I would believe 0.38% of **_all**_ desktop and mobile browsers, which against ~2 billion would be 7.6 million _potential_ users who might stumble across WebRTC for the first time.

What specific harm does leaving these firefox polyfills in cause?

fippo:
we don't support Chrome < 38 -- if I recall correctly this is when offerToReceive constraints started working. We do not shim this. Should we? Maybe... 1.02% of users are Chrome 29.

The minimum Firefox version which was said to be 31 (see #338; this minVersion check does not work for a number of reasons, including different version numbers of Opera and Chrome). We don't support the Firefox version which did not have a getLocalStreams() (22/23? One of the first ones with unflagged WebRTC) nor do we support the Chrome < 25 which did not have getLocalStreams() either.

if we continue to "support" those old versions, how are they tested? I am silently shrugging off reports of things not working in much newer browser versions because I know of too many bugs.

fippo:
@jan-ivar convinced me that it is too soon. Those versions will still get "best effort" support.

PR #349, opened 2016-08-02
edge: end-of-candidates can be at session level

fippo:
a=end-of-candidates is currently only parsed at m-line level but it can be a session-level attribute, 
see https://tools.ietf.org/html/draft-ietf-mmusic-trickle-ice-sip-04#section-8.2

kudos to @freeswitch for pointing this out


KaptenJansson:
LGTM

PR #350, opened 2016-08-04
edge: no longer ignore stun

fippo:
since 14393 Edge will ignore stun: urls itself. This PR allows forward compability and will just start to work when Edge actually supports stun:


KaptenJansson:
LGTM

PR #352, opened 2016-08-04
Bump version to 2.0.0

KaptenJansson:
**Description**
This includes breaking changes like removing attach/reattachMediastream helpers (everyone should use srcObject directly), minVersion check has been removed hence really old Chrome browsers ( < 50 which we do not want nor can support) will stop working.

Also lots of Edge fixes has been included such as stun support, filter IPV6 turn addresses (edge do no support this yet) and more.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "webrtc-adapter",
-  "version": "1.4.0",
+  "version": "2.0.0",
   "description": "A shim to insulate apps from WebRTC spec changes and browser prefix differences",
   "license": "BSD-3-Clause",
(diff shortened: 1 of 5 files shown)


fippo:
"minVersion check has been removed hence really old Chrome browsers ( < 50 which we do not want nor can support)" -- nah, I don't think we did breaking changes to old browsers (even though with Chrome < 44 you are now on your own). In particular we should continue to support Chrome 49 which is the last version available to poor old Windows XP

KaptenJansson:
By that statement I mean this is nothing we test. We support stable and >.

PR #360, opened 2016-08-26
Bower version is deprecated

bennycode:
**Description**

The Bower version attribute is deprecated ([source](https://github.com/bower/spec/blob/master/json.md#version)) and Bower will make use of `git` or `svn` tags. Since "webrtc-adapter" has proper [release versions](https://github.com/webrtc/adapter/releases), it is recommended to drop the `version` property in `bower.json`.

Less maintenance. Yay!! :speedboat: :pineapple:

**Purpose**

Bower version is deprecated and having that property can lead to version mismatch problems like this one:

screenshot

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,5 @@
 {
   "name": "webrtc-adapter",
-  "version": "2.0.2",
   "description": "A shim to insulate apps from WebRTC spec changes and browser prefix differences",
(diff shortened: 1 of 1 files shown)


KaptenJansson:
Yeah this was a pain, thanks for doing this. LGTM!

PR #366, opened 2016-10-03
re-enable firefox tests with geckodriver

fippo:
**Purpose**

this re-enables tests in Firefox 47+ using geckodriver

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,17 +12,14 @@ env:
     - BROWSER=firefox BVER=beta
     - BROWSER=firefox BVER=nightly
     - BROWSER=firefox BVER=esr
-    - BROWSER=MicrosoftEdge BVER=stable
 
 matrix:
   fast_finish: true
 
   allow_failures:
     - env: BROWSER=chrome  BVER=unstable
-    - env: BROWSER=firefox BVER=stable
-    - env: BROWSER=firefox BVER=beta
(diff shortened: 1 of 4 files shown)


fippo:
tests look broken -- they should wait for loadedmetadata. Will try to fix that as well.

fippo:
hrm. Tests fail in ESR (45) now which doesn't support geckodriver. What the...

fippo:
@jan-ivar @KaptenJansson wdyt? i'd rather have tests in current firefox and not in ESR than the other way round

KaptenJansson:
There is now way to run gecko for current and the old for ESR? I guess it's hard to figure out if its ESR from within the code. 

Your reasoning is fine by me.

fippo:
I tried and failed. Its either me or the new selenium stuff not working on the old ESR...

fippo:
also removed Edge since the VM I spun up is down again.

KaptenJansson:
LGTM. Nice refactoring of the tests too!

fippo:
LOADED_DATA is courtesy of @pehrsons ;-)

(errr.. HAVE_ENOUGH_DATA)

Pehrsons:
Hmm, you mean readyState 4? That's HAVE_ENOUGH_DATA.

If you could make it get triggered by the "canplaythrough" event instead of using a timeout though, that would be even better.

PR #369, opened 2016-10-10
edge: remove rtx from recv encoding parameters

fippo:
Edge 14942 throws errors if the rtx field is present in the receiving encoding parameters.


fippo:
@KaptenJansson ping ;-)

KaptenJansson:
LGTM

Issue #370, opened 2016-10-18
Chrome 53 broke facingMode polyfill


jan-ivar:
**Browsers and versions affected**
Chrome 53-56

**Description**
Chrome 53+ returns true for `navigator.mediaDevices.getSupportedConstraints().facingMode` even though it's not supported. This confuses adapter, which tests for support before polyfilling.

**Steps to reproduce**
- In Chrome for Android, click [Back] button in https://jsfiddle.net/j2oe36eL/

**Expected results**
Back camera

**Actual results**
Front camera.

KaptenJansson:
Filed https://bugs.chromium.org/p/chromium/issues/detail?id=656920

bradisbell:
This change gets the right camera for me.  Thanks!  However, using your branch, I am unable to get a 16:9 video from any camera.  Front camera works fine in 16:9 on the main fork of adapter.js 2.0.6.

Using the following constraints on a Galaxy S6 with Chrome:

```
    {
      "audio": {
        "echoCancellation": false,
        "channelCount": 2
      },
      "video": {
        "width": { "min": 640, "ideal": 1280, "max": 1920 },
        "height": { "min": 480, "ideal": 720, "max": 1080 },
        "frameRate": { "ideal": 30, "max": 30 },
        "facingMode": "environment"
      }
    }
```

jan-ivar:
@bradisbell This patch also turns off the polyfilling of constraints in Chrome 54 and on (which is what's breaking the test atm). Can you test to see if you have success with your constraints in Chrome 54+ _without_ adapter?

jan-ivar:
For 16:9 video I mean.

bradisbell:
No.  Chrome v54.0.2840.85, with no adapter.js and the constraints above gives me 4:3 video from the front-facing camera on the Galaxy S6.

Patrick-Vogt:
Did you found a solution for this problem?
I tried to enumerate the devices and and used sourceid = foundDeviceWithBackCamera, but it seems not to be a proper solution...

PR #371, opened 2016-10-18
Unbreak facingMode polyfill on Chrome 53-56 (#370)

jan-ivar:
**Description**
Ignore `navigator.mediaDevices.getSupportedConstraints().facingMode` on 53-56 because it lies.

**Purpose**

Fix https://github.com/webrtc/adapter/issues/370.

### DIFF
--- a/src/js/chrome/getusermedia.js
+++ b/src/js/chrome/getusermedia.js
@@ -72,7 +72,7 @@ module.exports = function() {
       // Shim facingMode for mobile, where it defaults to "user".
       var face = constraints.video.facingMode;
       face = face && ((typeof face === 'object') ? face : {ideal: face});
-      var getSupportedFacingModeLies = browserDetails.version < 57;
+      var getSupportedFacingModeLies = browserDetails.version < 59;
 
       if ((face && (face.exact === 'user' || face.exact === 'environment' ||
                     face.ideal === 'user' || face.ideal === 'environment')) &&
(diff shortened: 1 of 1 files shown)


KaptenJansson:
LGTM after the constraints tests have been updated.

KaptenJansson:
Ping

jan-ivar:
I'll update this to leave the constraints polyfill alone in Chrome.

KaptenJansson:
OK.

@alvestrand do you have any updates on facingMode?

alvestrand:
FacingMode CLs have been submitted to Chrome, and the flag is now Experimental (so that everyone who experiments gets it) - as of yesterday.
If you turn on experimental flags and call getSettings on a video track on a Chromebook, you should see the facing mode turn up.
Suggestion: Don't let the polyfill touch the facing mode when it's already present, only touch it when it's not.
Note: I don't think we select on facing mode yet. @guidou knows about that.

jan-ivar:
@alvestrand when you say you don't select on facingMode yet, do you mean the constraint is not working? If so, how can the polyfill know when it's not needed?

KaptenJansson:
@alvestrand @guidou ping

guidou:
The constraint resolution algorithm in chrome is not spec compliant and will not select a nondefault input device based on a constraint that is not deviceID.
The plan is to fix this in Chromium M58, though, so that a nondefault device can be selected based on the other constraints.

jan-ivar:
@KaptenJansson Looks like this never merged? Travis apparently failed for an unrelated reason, so I've rebased it and it's green now.

I've verified that it works in Chrome for Android in [this fiddle](https://jsfiddle.net/jib1/jsm6rLh0/) using a temp bundle from this tip.

KaptenJansson:
Sorry ;) LGTM

PR #373, opened 2016-10-20
edge: add profile level asymmetry flag to H264

fippo:
this currently breaks interop between Edge and Chrome M56, see
https://bugs.chromium.org/p/webrtc/issues/detail?id=6552

cc @alvestrand @magjed


KaptenJansson:
LGTM

PR #376, opened 2016-10-24
addIceCandidate(null): call callbacks

fippo:
currently this only resolves the promise.

Yes, I had code which was not using promises since it could not rely on them being natively available :-)


KaptenJansson:
LGTM

PR #378, opened 2016-10-31
add support for addIceCandidate(undefined)

fippo:
this adds support for addIceCandidate(undefined) in addition to addIceCandidate(null)

This would allow writing single-page samples and tests that just feed the event.candidate from onicecandidate into addIceCandidate which I think is the direction we want to go.

While the spec currently seems to be going into a different direction wrt to the exact format this is no worse than addIceCandidate(null)

@alvestrand

### DIFF
--- a/src/js/chrome/chrome_shim.js
+++ b/src/js/chrome/chrome_shim.js
@@ -239,11 +239,11 @@ var chromeShim = {
           };
         });
 
-    // support for addIceCandidate(null)
+    // support for addIceCandidate(null or undefined)
     var nativeAddIceCandidate =
         RTCPeerConnection.prototype.addIceCandidate;
     RTCPeerConnection.prototype.addIceCandidate = function() {
-      if (arguments[0] === null) {
+      if (!arguments[0]) {
         if (arguments[1]) {
           arguments[1].apply(null);
         }
--- a/test/test.js
+++ b/test/test.js
@@ -1464,6 +1464,8 @@ test('addIceCandidate with null', function(t) {
 
     var pc1 = new RTCPeerConnection(null);
     pc1.addIceCandidate(null)
+    // callback is called with either the empty result
+    // of the .then or the error from .catch.
     .then(callback)
     .catch(callback);
   };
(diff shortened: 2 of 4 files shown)


fippo:
wait, this failed when I added a test...

test/test.js
```diff
@@ -1485,6 +1485,35 @@ test('addIceCandidate with null', function(t) {
...
+test('addIceCandidate with undefined', function(t) {
+  var driver = seleniumHelpers.buildDriver();
+
+  var testDefinition = function() {
+    var callback = arguments[arguments.length - 1];
+
+    var pc1 = new RTCPeerConnection(null);
+    pc1.addIceCandidate(undefined)
+    .then(callback)
+    .catch(callback);
```
alvestrand, line 1501: Curious: How does this construct distinguish between success and error?
To me it seems to invoke the last argument passed to the function by "executeAsyncScript" no matter what it does. But I don't understand selenium yet.....

fippo: in this test I just test that either the callback or the error is thrown.
Maybe I should fail if the error is thrown -- which typically suggests that the native addIceCandidate has been called

fippo: ah no. The .then resolves without any argument while the error .catch will be resolved with an error. In line 1506 we check that err === null

alvestrand: so that's inside a promise that's returned from "driver.executeAsyncScript".
So 'driver.executeAsyncScript' calls 'testDefinition' in the driven browser, with a last argument that is a function for saying it's done. And the arguments passed to the callback get passed to the resolution of the executeAsyncScript promise.... can you add a comment after .then() saying "// No argument returned" and a comment after .catch saying "// Error is returned"?

This implicit argument-passing makes my head hurt.

alvestrand (approved):
I like it now that I understand it. But please add comments, or the understanding will slip away around 30 seconds from now.

test/test.js
fippo: not quite -- callback is what you're supposed to call when you're done in the executeAsyncScript part. So callback is executed in either the then or the catch and is called with what the promise resolves. will add two comments (also for the other one)

fippo:
phew, @jan-ivar pointed out that the argument in addIceCandidate is nullable (right now) so this is even the right thing to do spec-wise!

src/js/chrome/chrome_shim.js
```diff
@@ -239,11 +239,11 @@ var chromeShim = {
           };
         });
 
-    // support for addIceCandidate(null)
+    // support for addIceCandidate(null or undefined)
     var nativeAddIceCandidate =
         RTCPeerConnection.prototype.addIceCandidate;
     RTCPeerConnection.prototype.addIceCandidate = function() {
-      if (arguments[0] === null) {
+      if (arguments[0] === null || arguments[0] === undefined) {
```
jan-ivar: Why not simply `if (!arguments[0]) {` ?

fippo: that is probably the better way to do it now that I know this is legal.

fippo:
lets wait until after the wg interim before merging this since there is a related topic on the agenda.

fippo:
Hrm. I think the spec is going to take a different route here if I recall what @taylor-b proposed.
Shall we still merge this?

Or explicitly call addIceCandidate(null) in https://github.com/w3c/web-platform-tests/pull/4114

fippo:
Actually http://w3c.github.io/webrtc-pc/#simple-peer-to-peer-example shows exactly this "send anything even undefined event.candidate" so I think this is the right thing until we start to change the spec.

alvestrand:
@fippo somehow I think the changed API will still be backwards compatble (breaking changes are a Bad Thing), so merging this should be a Good Thing. Wil you add the explanatory comments I asked for in my Nov 1 note?

fippo:
already done [here](https://github.com/webrtc/adapter/pull/378/files#diff-c1129c8b045390789fa8ff62f2c6b4a9R1467)

fippo:
@KaptenJansson new release please? I think this is feature. Unless we consider it a bug that we didn't accept undefined ;-)

PR #379, opened 2016-10-31
Edge: improve common capability determination

fippo:
This improves the determination of common capabilities
such that two codecs will be considered equal even if
the number of channels does not match. In that case,
the lower number will be used.


KaptenJansson:
LGTM

PR #386, opened 2016-11-17
Don't shim maplike getStats in Firefox past 47.

jan-ivar:
**Description**
Remove maplike getStats shim in Firefox 48+

**Purpose**
Firefox 48 adds native support, and this redundant shim was triggering deprecation warnings there.


KaptenJansson:
LGTM

PR #390, opened 2016-12-09
update chrome shim

fippo:
this updates the chrome shim to no longer override the unprefixed RTCPeerConnection and don't attempt to shim promises if they're natively available (on the unprefixed connection)

If we do a major version bump soon i'd merge it then

### DIFF
--- a/src/js/chrome/chrome_shim.js
+++ b/src/js/chrome/chrome_shim.js
@@ -144,7 +144,13 @@ var chromeShim = {
           var standardStats = {
             id: report.id,
             timestamp: report.timestamp,
-            type: report.type
+            type: {
+              inboundrtp: 'inbound-rtp',
+              outboundrtp: 'outbound-rtp',
+              candidatepair: 'candidate-pair',
+              localcandidate: 'local-candidate',
+              remotecandidate: 'remote-candidate'
+            }[report.type] || report.type
           };
           report.names().forEach(function(name) {
             standardStats[name] = report.stat(name);
--- a/src/js/chrome/chrome_shim.js
+++ b/src/js/chrome/chrome_shim.js
@@ -133,6 +133,11 @@ var chromeShim = {
         return origGetStats.apply(this, arguments);
       }
 
+      // When spec-style getStats is supported, return those.
+      if (origGetStats.length === 0) {
+        return origGetStats.apply(this, arguments);
+      }
+
       var fixChromeStats_ = function(response) {
         var standardReport = {};
         var reports = response.result();
--- a/src/js/chrome/chrome_shim.js
+++ b/src/js/chrome/chrome_shim.js
@@ -112,7 +112,7 @@ var chromeShim = {
       };
       window.RTCPeerConnection.prototype = webkitRTCPeerConnection.prototype;
       // wrap static methods. Currently just generateCertificate.
-      if (!RTCPeerConnection.generateCertificate) {
+      if (webkitRTCPeerConnection.generateCertificate) {
         Object.defineProperty(window.RTCPeerConnection, 'generateCertificate', {
           get: function() {
             return webkitRTCPeerConnection.generateCertificate;
--- a/src/js/chrome/chrome_shim.js
+++ b/src/js/chrome/chrome_shim.js
@@ -97,21 +97,31 @@ var chromeShim = {
 
   shimPeerConnection: function() {
     // The RTCPeerConnection object.
-    window.RTCPeerConnection = function(pcConfig, pcConstraints) {
-      // Translate iceTransportPolicy to iceTransports,
-      // see https://code.google.com/p/webrtc/issues/detail?id=4869
-      // this was fixed in M56 along with unprefixing RTCPeerConnection.
-      logging('PeerConnection');
-      if (pcConfig && pcConfig.iceTransportPolicy) {
-        pcConfig.iceTransports = pcConfig.iceTransportPolicy;
+    if (!window.RTCPeerConnection) {
+      window.RTCPeerConnection = function(pcConfig, pcConstraints) {
+        // Translate iceTransportPolicy to iceTransports,
+        // see https://code.google.com/p/webrtc/issues/detail?id=4869
+        // this was fixed in M56 along with unprefixing RTCPeerConnection.
+        logging('PeerConnection');
+        if (pcConfig && pcConfig.iceTransportPolicy) {
+          pcConfig.iceTransports = pcConfig.iceTransportPolicy;
+        }
+
+        return new webkitRTCPeerConnection(pcConfig, pcConstraints);
+      };
+      window.RTCPeerConnection.prototype = webkitRTCPeerConnection.prototype;
+      // wrap static methods. Currently just generateCertificate.
+      if (!RTCPeerConnection.generateCertificate) {
+        Object.defineProperty(window.RTCPeerConnection, 'generateCertificate', {
+          get: function() {
+            return webkitRTCPeerConnection.generateCertificate;
+          }
+        });
       }
+    }
 
-      return new webkitRTCPeerConnection(pcConfig, pcConstraints);
-    };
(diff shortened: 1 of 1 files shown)


fippo:
probably easier to review individual commits

src/js/chrome/chrome_shim.js
```diff
@@ -97,142 +97,147 @@ var chromeShim = {
...
+        logging('PeerConnection');
+        if (pcConfig && pcConfig.iceTransportPolicy) {
+          pcConfig.iceTransports = pcConfig.iceTransportPolicy;
+        }
+
+        return new webkitRTCPeerConnection(pcConfig, pcConstraints);
+      };
+      window.RTCPeerConnection.prototype = webkitRTCPeerConnection.prototype;
+      // wrap static methods. Currently just generateCertificate.
+      if (!RTCPeerConnection.generateCertificate) {
```
fippo: ugh, this is wrong

src/js/chrome/chrome_shim.js
```diff
@@ -97,142 +97,147 @@ var chromeShim = {
...
-                args[1].apply(null, []);
-              },
-              function(err) {
-                if (args.length >= 3) {
-                  args[2].apply(null, [err]);
+      // add promise support -- natively available in Chrome 51
+      if (browserDetails.version < 51) {
+        ['setLocalDescription', 'setRemoteDescription', 'addIceCandidate']
+            .forEach(function(method) {
+              var nativeMethod = RTCPeerConnection.prototype[method];
```
alvestrand: This code is only going to execute in versions < 51. (Opera alert?)
Here we move the shimming from the underlying object to the shim object. Is that the Right Thing to do?
How do we test that it still works on version 50? Would rather leave code explicitly targeted at <51 unchanged as far as possible.

alvestrand:
If we do a major version bump, perhaps we should set minVersion to 51 and delete all the code that is flagged as < 51, instead of trying to update it?

src/js/chrome/chrome_shim.js
fippo: Sadly this code has so far shimmed promises on _all_ opera versions...

fippo: Maybe I should just remove the version check here? This would assume lack of promise support for _all_ webkitRTCPeerConnection and shim promises using callbacks. Not nice but should work ok.

alvestrand:
Opera user agent:
"Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/54.0.2840.99 Safari/537.36 OPR/41.0.2353.69"

Chrome user agent:
"Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/55.0.2883.75 Safari/537.36"

With that string, it seems that Opera 41 should be detected as Chrome 54, which might be OK - we shouldn't get Opera's version number in there without looking for that OPR/ piece.

fippo:
oh great. I was assuming we did smarter things here like [platform](https://www.npmjs.com/package/platform) does but if we're not trying to be smart...

The last commit has been purged ;-)

fippo:
wrt to moving from webkitRTCPeerConnection to RTCPeerConnection: since we set RTCPeerConnection.prototype to webkitRTCPeerConnection.prototype those should be identical. This way we don't get to see the ugly prefixed thing that often ;-)

KaptenJansson:
@alvestrand any other thoughts?

alvestrand:
Agree with merging after a major version bump. Is there a label for that?

KaptenJansson:
There are other things that also warrants a major bump that has been merged, could just merge this and do the major bump now if this PR is ready for that?

src/js/chrome/chrome_shim.js
alvestrand: Another thread indicated that on Opera, we exract the underlying Chrome version, so we've been ignoring which Opera version we're in fact using all along

Shimming promises when we don't need to do so is not OK.

alvestrand:
Then just merge it!

fippo:
let me make one or two more commits here... 
1) return spec stats if supported -- unless the first argument is a function
2) change type names to include the hypens similar to what is done for Edge and Firefox. This is an easy change that will make life easier for implementors

src/js/chrome/chrome_shim.js
fippo: Ok. I'll also change createOffer/createAnswer to only add promise support in < M52. M52 has to be shimmed because of https://bugs.chromium.org/p/chromium/issues/detail?id=619289
So far we haven't un-promisified this.

fippo:
Did I say "two"? Well, only four real commits.
Shall I squash post-review pre-merge?

KaptenJansson:
I can squash, LGTM.

src/js/chrome/chrome_shim.js
```diff
@@ -97,117 +97,101 @@ var chromeShim = {
...
+
+      var fixChromeStats_ = function(response) {
+        var standardReport = {};
+        var reports = response.result();
+        reports.forEach(function(report) {
+          var standardStats = {
+            id: report.id,
+            timestamp: report.timestamp,
+            type: {
+              inboundrtp: 'inbound-rtp',
```
foolip, line 148: Is this in anticipation of Chrome shipping stats *without* hyphens, like Firefox, even though that is not the plan?

henbos: These are Chrome's old getStats? These are the types that has:
https://cs.chromium.org/chromium/src/third_party/webrtc/api/statstypes.cc?type=cs&q=statstypes.cc:33&sq=package:chromium&l=33

inboundrtp/outboundrtp/candidatepair are Firefox concepts, not Chromium.
localcandidate/remotecandidate are applicable, but most members are called different things in the spec and what Chromium returns anyway, so I'm not sure changing the type of these helps anyone?

Here's the mapping from new getStats/the spec to old stats:
```
RTCIceCandidateStats : RTCStats {
  type           = "local-candidate" or "remote-candidate" instead of "localcandidate" or "remotecandidate"
  ip             = [localcandidate/remotecandidate].ipAddress
  port           = [localcandidate/remotecandidate].portNumber
  protocol       = [localcandidate/remotecandidate].transport
  candidateType  = [localcandidate/remotecandidate].candidateType
  priority       = [localcandidate/remotecandidate].priority
  url            = not implemented, also missing from old stats

  In old that is not in new:
  [localcandidate/remotecandidate].networkType
}
```

Even with "fixing" the type you would still have to treat chrome and firefox differently for these stats.

fippo: this is mostly about localcandidate and remotecandidate -- reusing the hyphenator from Firefox and Edge made slightly more sense than just having these two. Can change if that improves clarity

fippo: https://github.com/webrtc/adapter/pull/402 -- I think it does improve clarify so until that map moves into common code...

henbos: Agreed that both Firefox and Chrome should update to local/remote-candidate, they're similar to each other. (Commented on #402)

alvestrand:
Please .... write tests!!!!!!!!!!!!!!

fippo:
We have no test coverage for getStats values :-/

alvestrand:
Test coverage for the code. Synthesize inputs and outputs, but exercise the code in a test.

Tests in adapter don't test the implementation - they test the adapter.

fippo:
We have tests running in the browser but currently we don't have such tests (which I assume would run in node). Yes, we should :-/

alvestrand:
Specifically I'm talking about the tests run under https://github.com/webrtc/adapter/tree/master/test and run by travis.

Put it this way: Update those tests to call the code you've just written, and check that the result looks right. When someone comes along and breaks your code by accident, you'll be glad you did.
(Probably you'll never know, because Travis will catch him, and he'll not admit to the problem....)

Issue #389, opened 2016-12-08
need upgrade path to chrome's spec stats


fippo:
Chrome Canary just landed spec-style stats and they're great.
Now we have an issue we avoided for a very long time: how to upgrade from chrome-stats to legacy stats.

When getStats is called with anything not a function as the first argument we shim from the legacy stats:
https://github.com/webrtc/adapter/blob/master/src/js/chrome/chrome_shim.js#L114-L118
This was good because we could determine whether the user was interested in the chrome stats or a more spec-style statistics. Now the defintion of "spec-style" has changed since we made that decision.

How do we determine whether
a) the user wants spec stats
b) the browser supports it. webrtcdetectedbrowser >= 57 doesn't work since that would mean legacy stats in opera for a looooong time (see also [here](https://lists.w3.org/Archives/Public/public-webrtc/2016Dec/0006.html))

Also how to deal with old chrome versions (real-world concern, even though appear.in officially only supports the latest two chrome versions we don't enforce this). I wonder if it makes sense to take @henbos' document and take another stab at writing the JS shim...

KaptenJansson:
@henbos and @alvestrand could you please share you thoughts?

alvestrand:
Hm. Is it possible to detect that there are 2 getStats functions, and what the difference is?

We can detect what the user wants:
- arg is function: Old-style
- arg is missing: new-style
- arg is object: new-style

but we have to detect what the browser supports too.
(the webrtcdetectedbrowser thing just shows that we need to be able to separate Opera from Chrome.)

fippo:
I'm in favor of arguments.length === 0 => new style. So far the samples have shown track with an explicit null and I use that in my code as well. This should be a major version bump for adapter.

henbos:
Note: It's available in Canary *behind a flag* (Experimental Web platform feature), PSA: https://groups.google.com/forum/#!topic/discuss-webrtc/7TXEazn9kTg

The two getStats functions (RTCPeerConnection.idl):
```
[CallWith=ScriptState, LegacyInterfaceTypeChecking] Promise<void> getStats(RTCStatsCallback successCallback, optional MediaStreamTrack? selector);
[CallWith=ScriptState, RuntimeEnabled=RTCPeerConnectionNewGetStats] Promise<RTCStatsReport> getStats(optional MediaStreamTrack? selector);
```

[...]

PR #391, opened 2016-12-09
edge: remove typ endOfCandidates hack

fippo:
this removes the endOfCandidates hack. The samples will be updated
to simply pass every event.candidate (including undefined) to the
other addIceCandidate which removes the need for this.

Needs a major version bump after merging. Lets do it together with the bump for #389


KaptenJansson:
LGTM

PR #393, opened 2016-12-09
edge: change getStats

fippo:
while we're breaking getStats i thought it might be nice to do the same for Edge as well.
This adds hyphens to the stats types and removes the map-object hybrid (h/t @jan-ivar)


KaptenJansson:
LGTM

PR #395, opened 2016-12-09
Do not set remote candidates when setting local answer

ggarber:
**Description**
Setting the local answer for a remote offer is setting the remote candidates in the ice transport again (they were already set during the processing of the remote offer), raising an IllegalState error.   

**Purpose**
Remove setting remote candidates on SLD(answer) as they were already set on SRD(offer).


fippo:
LGTM -- thanks @ggarber!

fippo (approved):

PR #396, opened 2016-12-10
test: fix ontrack test

fippo:
the ontrack test failed in Edge because it returned 8 results
instead of 7. The correct number of tests to pass is 8 if
getReceivers() is supported.

This re-enables the test in Firefox as well.


KaptenJansson:
LGTM

PR #398, opened 2016-12-13
Add node 6 under engines in package.json

KaptenJansson:
**Description**
Add node v 6.0.0 to engines

**Purpose**
Highlight that at least node 6 is required for this repo (due to webdriver)


KaptenJansson:
fyi @fippo

PR #401, opened 2016-12-19
FIX bug to avoid triggering addstream multiple times

ggarber:
**Description**

**Purpose**


fippo (approved):
geee... can you think of a test that fails without this?
Good catch!

KaptenJansson:
LGTM

PR #402, opened 2016-12-20
chrome: only hyphenate what is required

fippo:
only hyphenate localcandidate and remotecandidate since those
are the only almost-spec stats chrome returns.

### DIFF
--- a/src/js/chrome/chrome_shim.js
+++ b/src/js/chrome/chrome_shim.js
@@ -145,9 +145,6 @@ var chromeShim = {
             id: report.id,
             timestamp: report.timestamp,
             type: {
-              inboundrtp: 'inbound-rtp',
-              outboundrtp: 'outbound-rtp',
-              candidatepair: 'candidate-pair',
               localcandidate: 'local-candidate',
               remotecandidate: 'remote-candidate'
             }[report.type] || report.type


KaptenJansson:
@henbos @foolip is this as you expect?

foolip:
It gets rid of the mappings that should be dead code, but I'll leave it to @henbos to say if Chrome's current "localcandidate" and "remotecandidate" are similar enough to the per-spec "local-candidate" an "remote-candidate" to usefully map them.

henbos:
The Chrome and Firefox localcandidate/remotecandidate are similarly incorrect, e.g. have "ipAddress" and "portNumber" instead of spec's "ip" and "port", so given that Firefox is updated with hyphens it makes sense to update this too.

fippo:
ugh. When was _that_ changed in the spec. Might make sense to backfill those values (even though I really don't want to go the route of trying to backfill the complete stats again)

jan-ivar:
Thanks, filed https://bugzil.la/1324788.

KaptenJansson:
@henbos ping

henbos (approved):
LGTM.

KaptenJansson:
Any last objections before merging this?

fippo:
go ahead. I think that is the last thing that keeps me from "announcing" 3.0

jan-ivar:
@fippo I have one more thing I'd like to get in before 3.0... I'd like to iterate on your hyphenate poly for firefox so it doesn't kill our deprecations warnings. Working on it now.

KaptenJansson:
LGTM

PR #405, opened 2017-01-03
ADD required config to RTCPeerConnection test

ggarber:
**Description**

**Purpose**

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -299,7 +299,7 @@ test('Create RTCPeerConnection', function(t) {
     t.plan(2);
     t.pass('Page loaded');
     return driver.executeScript(
-      'return typeof(new RTCPeerConnection()) === \'object\'');
+      'return typeof(new RTCPeerConnection(null)) === \'object\'');
   })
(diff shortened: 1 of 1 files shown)


fippo (commented):
why? Don't tell me chromes unprefixed RTCPeerConnection requires an argument here?

ggarber:
Chrome 55:

```
> new webkitRTCPeerConnection()
VM145:1 Uncaught TypeError: Failed to construct 'RTCPeerConnection': 1 argument required, but only 0 present.
    at <anonymous>:1:1
(anonymous) @ VM145:1
> new webkitRTCPeerConnection(null)
RTCPeerConnection { …}
```

fippo:
Right, that is the old behaviour of webkitRTCPeerConnection. But adapter is hiding that error, no?

ggarber:
You are right, it was a problem in my code not supporting the constructor without arguments.   But in rest of adapter tests you always use RTCPeerConnection(null), so maybe you want to homogenize it at some point.

Feel free to close it.   Thx

PR #407, opened 2017-01-04
edge: fix two issues

fippo:
see individual commits. tl;dr: the first commit is from the "how did this ever work?" category. The second fixes a regression I introduced when removing the typ endOfCandidates hack

### DIFF
--- a/src/js/edge/edge_shim.js
+++ b/src/js/edge/edge_shim.js
@@ -315,7 +315,9 @@ var edgeShim = {
             // Also emits null candidate when all gatherers are complete.
             switch (self.iceGatheringState) {
               case 'new':
-                self._localIceCandidatesBuffer.push(event);
+                if (!end) {
+                  self._localIceCandidatesBuffer.push(event);
+                }
                 if (end && complete) {
                   self._localIceCandidatesBuffer.push(
                       new Event('icecandidate'));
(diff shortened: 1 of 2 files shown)


fippo:
still getting a test failure for the basic test with promises (but callbacks are ok) so I might need another commit to fix things. Will delete this when its ready to merge :-/

fippo:
the tests with promises failed for some reason. Only in the test environment, not when copying them to a html file. @KaptenJansson do you remember why this copy is necessary?

KaptenJansson:
Test environment being nodejs or browser context? AFAIK a browser requires a HTML context for running javascript unless it's an extension.

fippo:
the basic connection establishment with promises test failed but copying that code to a html page worked. Which I find somewhat confusing.

But it works now after not copying the null candidate.

KaptenJansson:
LGTM

tgabi333:
@fippo what is the current status of communicating with edge? does video work somehow?

fippo:
@tgabi333 in the insider builds where H264 is enabled by default -- in the current release build you need to activate a flag

tgabi333:
Thanks @fippo , i didn't get to work it even if i switch experimental H264 support on to transmit video  :S 

Do you have any sample page?

fippo:
@tgabi333 https://simplewebrtc.com/video.html works fine here (latest insider build)

PR #392, opened 2016-12-09
firefox: shim hyphens in getStats types

fippo:
changes the types of the getStats reports from (e.g.) localcandidate
to local-candidate. Workaround until
  https://bugzilla.mozilla.org/show_bug.cgi?id=1322503
is fixed.

part of the major version bump planned for #389

PR #409, opened 2017-01-10
Limit spec-hyphens in type names to promise getStats and <FF53.

jan-ivar:
**Description**
Follow-up to https://github.com/webrtc/adapter/pull/392 now that [bug 1322503](https://bugzil.la/1322503) has landed.

**Purpose**
Limit spec hyphens to promise version of getStats and to FF versions <53.


jan-ivar:
Basically Firefox supported `forEach` long before maplike, so we leave callback version alone. (we're going to deprecate callback version soon, but warnings before breaking).

KaptenJansson:
LGTM

PR #411, opened 2017-01-17
Shim URL.createObjectURL(stream) using srcObject.

jan-ivar:
**Description**
Shim the old `URL.createObjectURL(stream)` using the new `elem.srcObject = stream`, with deprecation warning.

Other uses like `URL.createObjectURL(mediaSource)` and `URL.createObjectURL(blob)` are left intact.

**Purpose**
Mitigation of potential future removal of `URL.createObjectURL(stream)` in browsers, as discussed in https://github.com/w3c/mediacapture-main/issues/404#issuecomment-272217013.

### DIFF
--- a/src/js/utils.js
+++ b/src/js/utils.js
@@ -119,6 +119,39 @@ var utils = {
     }
 
     return result;
+  },
+
+  shimCreateObjectURL: function() {
+    var nativeCreateObjectURL = URL.createObjectURL.bind(URL);
+    var nativeRevokeObjectURL = URL.revokeObjectURL.bind(URL);
+    var streams = new Map(), newId = 0;
+
+    URL.createObjectURL = function(stream) {
+      if ('getTracks' in stream) {
+        var url = 'polyblob:' + (++newId);
+        streams.set(url, stream);
+        console.log('URL.createObjectURL(stream) is deprecated! ' +
+                    'Use elem.srcObject = stream instead!');
+        return url;
+      }
+      return nativeCreateObjectURL(stream);
+    };
+    URL.revokeObjectURL = function(url) {
+      nativeRevokeObjectURL(url);
+      streams.delete(url);
+    };
+
+    var dsc = Object.getOwnPropertyDescriptor(window.HTMLMediaElement.prototype,
+                                              'src');
+    Object.defineProperty(window.HTMLMediaElement.prototype, 'src', {
+      get: function() {
+        return dsc.get.apply(this);
+      },
+      set: function(url) {
+        this.srcObject = streams.get(url) || null;
+        return dsc.set.apply(this, [url]);
+      }
+    });
   }
 };
 
@@ -127,5 +160,6 @@ module.exports = {
   log: utils.log,
   disableLog: utils.disableLog,
   browserDetails: utils.detectBrowser(),
-  extractVersion: utils.extractVersion
+  extractVersion: utils.extractVersion,
+  shimCreateObjectURL: utils.shimCreateObjectURL
 };
--- a/src/js/adapter_core.js
+++ b/src/js/adapter_core.js
@@ -12,12 +12,13 @@
 // Shimming starts here.
 (function() {
   // Utils.
-  var logging = require('./utils').log;
-  var browserDetails = require('./utils').browserDetails;
+  var utils = require('./utils');
+  var logging = utils.log;
+  var browserDetails = utils.browserDetails;
   // Export to the adapter global object visible in the browser.
   module.exports.browserDetails = browserDetails;
-  module.exports.extractVersion = require('./utils').extractVersion;
-  module.exports.disableLog = require('./utils').disableLog;
+  module.exports.extractVersion = utils.extractVersion;
+  module.exports.disableLog = utils.disableLog;
 
   // Uncomment the line below if you want logging to occur, including logging
   // for the switch statement below. Can also be turned on in the browser via
@@ -46,6 +47,7 @@
       chromeShim.shimGetUserMedia();
       chromeShim.shimMediaStream();
       chromeShim.shimSourceObject();
+      utils.shimCreateObjectURL();
       chromeShim.shimPeerConnection();
       chromeShim.shimOnTrack();
       break;
@@ -59,6 +61,7 @@
       module.exports.browserShim = firefoxShim;
 
       firefoxShim.shimGetUserMedia();
+      utils.shimCreateObjectURL();
       firefoxShim.shimSourceObject();
       firefoxShim.shimPeerConnection();
       firefoxShim.shimOnTrack();
@@ -73,6 +76,7 @@
       module.exports.browserShim = edgeShim;
 
       edgeShim.shimGetUserMedia();
+      utils.shimCreateObjectURL();
       edgeShim.shimPeerConnection();
       break;
     case 'safari':
--- a/test/test.js
+++ b/test/test.js
@@ -480,7 +480,7 @@ test('createObjectURL shim test', function(t) {
   // Run test.
   seleniumHelpers.loadTestPage(driver)
   .then(function() {
-    t.plan(4);
+    t.plan(5);
     t.pass('Page loaded');
     return driver.executeAsyncScript(testDefinition);
   })
--- a/test/test.js
+++ b/test/test.js
@@ -360,9 +360,6 @@ test('Video srcObject getter/setter test', function(t) {
     return driver.executeScript(
         'return document.getElementById(\'video\').srcObject.id')
     .then(function(srcObjectId) {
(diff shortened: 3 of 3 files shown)


alvestrand:
As a general thing: Can you add some tests for this?

I'm particularly eager to see that createObjectURL returns distinct URLs for multiple calls, and that video.src = createObjectURL(stream) results in the stream being connected to "video".
If it's not tested, it will break....

KaptenJansson:
LGTM,  thanks for doing this, I guess this warrants at least a minor version update?

alvestrand:
Yes, definitely a version bump. 3.1.0 may be appropriate, since it affects codepaths that used to be unaffected.

jan-ivar:
Actually, ideally, I think we'd want this in even without a major version number, since it's a mitigation feature (it's the tearing away again of the feature that needs to be a major version number). But there's probably no immediate rush either.

KaptenJansson:
Its already in 3.1

fippo:
i've just gotten reports that this causes errors for users on a rasperry pi (which might be a chrome version that is very old and does not have native srcObject) which leads to Maximum call stack size exceeded at
    Function.o.shimCreateObjectURL.URL.createObjectURL

I think this needs to check that srcObject is in the native prototype.

fippo:
Ugh. Also: ```Mixed Content: The page at 'https://...' was loaded over HTTPS, but requested an insecure video 'polyblob:1'. This content should also be served over HTTPS.```

jan-ivar:
@fippo OK sorry about that. I guess the polyfill should test for the existence of srcObject?

jan-ivar:
@fippo Where did you get the Mixed content warning? Do you have STRs?

flo-sch:
Hey, I just tested a WebRTC call from Chrome 55 to Chrome 49 ;

The person using Chrome v49 (which unfortunately cannot update it further ---> stuck with OS X 10.6.8 at the moment) had a `Maximum call stack size exceeded` on line 2720 of adapter.js :
`streams.set(url, stream);`

I am currently using a `<video>` element with ReactJS, and handle the stream this way :
(`RemoteStreamStore.getState()` returns either a valid `MediaStream` object, either `null`)

```
_onStreamChange() {
    let stream = RemoteStreamStore.getState(),
        url = this.state.url,
        videoElement = ReactDOM.findDOMNode(this.refs.remotePlayer);

    if (typeof videoElement.srcObject !== 'undefined') {
        videoElement.srcObject = stream;

        url = null;
    } else {
        if (url) {
            URL.revokeObjectURL(url);
        }

        url = URL.createObjectURL(stream);

        videoElement.src = url;
    }

    this.setState({
        stream: stream,
        url: url
    });
}

// [...]

render() {
    return (
        <video
            ref='remotePlayer'
            autoPlay
            />
    );
}
```

Am I missing something about the use of adapter.js ?

jan-ivar:
@flo-sch Thanks, yeah this patch broke pre-52 Chrome. I'll add a fix today to check for the existence of `srcObject` before attempting to polyfill with it. Thanks for reporting, and sorry for the trouble!

flo-sch:
@jan-ivar thanks for the reactivity :)

No worries for the troubles, I just wanted to be sure it wasn't coming from a bad practice from me...

And thanks a lot for providing such polyfills in this huge mess of standards (can't wait to see when WebRTC specs will actually be stable!)

Anyway, just wondering, will the test on `srcObject` I am doing still be necessary, or will the polyfill automatically fallback on `createObjectURL(stream)` when trying to use an undefined `srcObject` ?
(not sure if it's even possible)

fippo:
@flo-sch you should just set srcObject. That will work in Chrome 45+ which is 99.9% of the population (including, sadly, those users on Windows XP stuck on M49)

Your check looks wrong though... you should check on the prototype. See https://github.com/webrtc/adapter/blob/master/src/js/chrome/chrome_shim.js#L59

flo-sch:
Alright @fippo thanks for noticing, I will update my component as soon as the new version of the polyfill is released :)

jan-ivar:
As I recall, adapter polyfills srcObject on older browsers. That's probably why we're blowing the stack in fact.

jan-ivar:
@flo-sch and @fippo if you could test with https://github.com/webrtc/adapter/issues/420 that would be awesome. Old Chrome is hard to come by.

src/js/utils.js
```diff
@@ -119,6 +119,39 @@ var utils = {
     }
 
     return result;
+  },
+
+  shimCreateObjectURL: function() {
+    var nativeCreateObjectURL = URL.createObjectURL.bind(URL);
+    var nativeRevokeObjectURL = URL.revokeObjectURL.bind(URL);
+    var streams = new Map(), newId = 0;
```
foolip, line 127: Is Map supported on all browser versions that you want to support? (I checked that it is in Edge 14, but not when it was added to Chrome and Firefox.)

src/js/utils.js
```diff
@@ -119,6 +119,39 @@ var utils = {
...
+
+    URL.createObjectURL = function(stream) {
+      if ('getTracks' in stream) {
+        var url = 'polyblob:' + (++newId);
+        streams.set(url, stream);
+        console.log('URL.createObjectURL(stream) is deprecated! ' +
```
foolip, line 133: `console.warn` gives you a nice little warning triangle, just like built-in deprecations.

src/js/utils.js
```diff
@@ -119,6 +119,39 @@ var utils = {
...
     return result;
+  },
+
+  shimCreateObjectURL: function() {
+    var nativeCreateObjectURL = URL.createObjectURL.bind(URL);
+    var nativeRevokeObjectURL = URL.revokeObjectURL.bind(URL);
+    var streams = new Map(), newId = 0;
+
+    URL.createObjectURL = function(stream) {
+      if ('getTracks' in stream) {
```
foolip, line 130: That `stream instanceof MediaStream` wouldn't work reliably is kind of sad, maybe in 5 years :)

src/js/utils.js
```diff
@@ -119,6 +119,39 @@ var utils = {
...
+        return url;
+      }
+      return nativeCreateObjectURL(stream);
+    };
+    URL.revokeObjectURL = function(url) {
+      nativeRevokeObjectURL(url);
+      streams.delete(url);
+    };
+
+    var dsc = Object.getOwnPropertyDescriptor(window.HTMLMediaElement.prototype,
```
foolip, line 144: This `window.` prefix isn't needed.

src/js/utils.js
```diff
@@ -119,6 +119,39 @@ var utils = {
...
+                                              'src');
+    Object.defineProperty(window.HTMLMediaElement.prototype, 'src', {
+      get: function() {
+        return dsc.get.apply(this);
+      },
+      set: function(url) {
+        this.srcObject = streams.get(url) || null;
```
foolip, line 151: Setting it to null here seems unhelpful, maybe set srcObject if there is a stream, and otherwise forward to the original setter?

foolip (commented):
Just a few nits, AFAICT the code will work fine as-is.

src/js/utils.js
jan-ivar: [Yes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map#Browser_compatibility).

src/js/utils.js
jan-ivar: I didn't know that was standard. Good idea.

src/js/utils.js
jan-ivar: You're right, the `|| null` is redundant since `undefined` produces the same result (`null` I checked). However, nulling it on regular URLs is important because `srcObject` trumps `src`.

fippo:
@jan-ivar still getting 
Failed to load resource: net::ERR_UNKNOWN_URL_SCHEME
I don't think making up URL schemes is the right thing to do here :-(

I can't repro outside our production angular environment. I noticed that src is set (to polyBlob:1) but srcObject is not.

jan-ivar:
@fippo We can try s/polyblob/blob/ perhaps... Think that would help?

alvestrand:
Making up URL schemes SHOULD be the right thing to do, because that will result in exactly this kind of error when the URLs go where you haven't catered to them.
There's some function (platform, framework or whatever) that is trying to use the URL as an URL instead of a key into the adapter data structure. We should expect that problem when generating URLs.
Making up blob: URLs should give you "illegal URL value" instead - even harder to debug.

Far-fetched: Is it possible to install a scheme handler from JS?

fippo:
@jan-ivar the issue (see #438) is that srcObject is not set in that case so the video object falls back to the blob link which doesn't lead anywhere.

PR #417, opened 2017-01-22
edge: reject if there are no common video codecs

fippo:
also has to check that there are actually codec-codecs
in common and not just things like RTX and RED.


KaptenJansson (approved):
LGTM

PR #418, opened 2017-01-26
chrome: restrict new getStats

fippo:
this will only call the new getStats when there are no arguments
supplied or when the selector argument is null.

Implements @henbos's suggestion in  https://github.com/webrtc/samples/issues/863#issuecomment-275436014

### DIFF
--- a/src/js/chrome/chrome_shim.js
+++ b/src/js/chrome/chrome_shim.js
@@ -132,9 +132,11 @@ var chromeShim = {
         return origGetStats.apply(this, arguments);
       }
 
-      // When spec-style getStats is supported, return those.
-      if (origGetStats.length === 0) {
-        return origGetStats.apply(this, arguments);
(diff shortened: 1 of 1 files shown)


henbos:
We need adapter.js to do the following when new getStats is available:
1. If called with no parameter, invoke new getStats.
2. If called with 1 parameter and it is not a callback function (e.g. it is null, a track, or something else), invoke new getStats - ignoring the parameter.
3. In all other cases, invoke the old getStats, passing the parameters along.

Instead of 2. we could explicitly check for null but there will likely be a selector argument added to the spec, so 2. is safer for the future.

fippo:
since we're JS and not webidl (where from what I heard this is a problem) we can check (2) as  typeof(arguments[0] !== 'function')

KaptenJansson:
LGTM, @henbos ?

henbos:
LGTM

PR #419, opened 2017-01-26
edge: re-enable RTX

fippo:
all the stuff required for RTX should have been hooked up in Edge 15007
so its no longer necessary to delete this.


KaptenJansson:
LGTM

PR #420, opened 2017-01-27
Don't shim createObjectURL on old browsers wo/srcObject to avoid infinite recursion.

jan-ivar:
**Description**
Fix for https://github.com/webrtc/adapter/pull/411#issuecomment-275085135.

**Purpose**
Oops.

### DIFF
--- a/src/js/adapter_core.js
+++ b/src/js/adapter_core.js
@@ -46,8 +46,8 @@
 
       chromeShim.shimGetUserMedia();
       chromeShim.shimMediaStream();
-      chromeShim.shimSourceObject();
       utils.shimCreateObjectURL();
+      chromeShim.shimSourceObject();
       chromeShim.shimPeerConnection();
       chromeShim.shimOnTrack();
       break;
--- a/src/js/chrome/chrome_shim.js
+++ b/src/js/chrome/chrome_shim.js
@@ -72,7 +72,7 @@ var chromeShim = {
 
             if (!stream) {
               this.src = '';
-              return;
+              return undefined;
             }
             this.src = URL.createObjectURL(stream);
             // We need to recreate the blob url when a track is added or
--- a/src/js/utils.js
+++ b/src/js/utils.js
@@ -121,7 +121,7 @@ var utils = {
     return result;
   },
 
-  // shimCreateObjectURL must be called before shimSourceObject to avoid loop!
+  // shimCreateObjectURL must be called before shimSourceObject to avoid loop.
 
   shimCreateObjectURL: function() {
     if (!(typeof window === 'object' && window.HTMLMediaElement &&
--- a/src/js/utils.js
+++ b/src/js/utils.js
@@ -121,7 +121,15 @@ var utils = {
     return result;
   },
 
+  // shimCreateObjectURL must be called before shimSourceObject to avoid loop!
+
   shimCreateObjectURL: function() {
+    if (!(typeof window === 'object' && window.HTMLMediaElement &&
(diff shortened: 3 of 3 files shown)


alvestrand:
1) Do you have a test for this?
2) It seems that Firefox blows up on Travis, which is odd since the patch doesn't touch Chrome code, and previous builds passed. Can you investigate? (at least run the test locally)

fippo:
This works locally but it seems that we have _some_ issue in Firefox 52 which is now beta. The last successful builds we had with beta were
https://travis-ci.org/webrtc/adapter/jobs/194167230#L727
https://travis-ci.org/webrtc/adapter/jobs/192888091#L741
both of which were FF51.

KaptenJansson:
Yeah it might be worth to add a test for this.

flo-sch:
Works now on Chrome 49 (same environment as yesterday) :)

At least we don't have errors while setting the streams ;
Remote video player has a defined srcObject element, containing both audio and video MediaStreamTracks, which are all enabled ;
RTCPeerConnection is established without error (both session descriptions and ice candidates) ;

But we don't see or hear each other...
Not sure to understand why?

Could it be related to the network, even if ICECandidate are successfully set? (Not using any STUN or TURN server at the moment)

alvestrand:
If possible, can you see what happens if you use a 2.x version of adapter.js?
If that works, and adapter-latest doesn't, it's an adapter bug.

flo-sch:
Hum, yep, same result with adapter@2.1, so it's not related to adapter.js. Sorry about that! I will investigate on the network issues.

jan-ivar:
@alvestrand [#411](https://github.com/webrtc/adapter/pull/411/files#diff-c1129c8b045390789fa8ff62f2c6b4a9R445) added a test for this. It seems to be the one failing in fact. Odd. We don't test old versions of Chrome AFAIK? Unfortunately, I've not been able to run adapter's npm tests locally for over a year, so I've had to test manually with browserify, or by committing and looking at Travis. There's a chance it's provoking a real intermittent we're chasing. Is there a way to re-trigger Travis?

alvestrand:
when you view the run on travis-ci, there's a "restart" button in the upper righthand corner if you have the right to restart; I think it may need to have you in some list of contributors or owners to the repo.
@KaptenJansson  may know more.

fippo:
that button newer worked for me so might require permissions neither @jan-ivar nor me have

jan-ivar:
OK green now, with no change. Looks like an intermittent. :-(

fippo:
@KaptenJansson can you squash before merging? :-)

jan-ivar:
I suspect the 52 intermittent is [bug 1335494](https://bugzil.la/1335494) or [bug 1317126](https://bugzil.la/1317126).

KaptenJansson:
LGTM

flo-sch:
Great :)

Any possible tag for this one ? As it solves a compatibility break, I'd love to update the npm package as a stable release (`3.1.1` maybe?) instead of the commit hash.

KaptenJansson:
@flo-sch 3.1.1 is now published to npm.

flo-sch:
You're the best :)
Thanks a lot!

fippo:
@KaptenJansson https://www.npmjs.com/package/webrtc-adapter still shows 3.1.0

flo-sch:
`npm info webrtc-adapter` makes reference to `3.1.1`, so it's working.

But maybe the NPM page doesn't automatically listen on new releases ? (webhook GitHub ?)

fippo:
it should... but as long as its actually published we're good. Thanks!

PR #425, opened 2017-01-27
test: fix basic-promise in edge

fippo:
Edge was throwing an exception when trying to apply dictionary() to
the null/undefined candidate. Trying to convert this again made no sense.


KaptenJansson:
LGTM

PR #429, opened 2017-02-02
 Revert "edge: reject if there are no common video codecs"

fippo:
The change caused weird issues where the audio m-line was rejected on the PC1 demo. I'll have to do that in a differen way (and this time with tests...)

### DIFF
--- a/src/js/edge/edge_shim.js
+++ b/src/js/edge/edge_shim.js
@@ -593,26 +593,6 @@ var edgeShim = {
                 .filter(function(cand) {
                   return cand.component === '1';
                 });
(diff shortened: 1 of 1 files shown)


KaptenJansson:
LGTM. I forget but did we evaluate https://www.appveyor.com/ for adding windows support?

fippo:
main problem is that we need the insider version. We had stuff running via a remote selenium at times but it turned out to be tedious.

And tests didn't catch this for some reason. Well... post-upgrade checklist!

KaptenJansson:
Can't we find a MS person that could own this? ;)

PR #430, opened 2017-02-02
Revert "edge: re-enable RTX"

fippo:
This reverts commit 3a6d68a4321f841050ccde599ba9ca4f38057ea1.

While RTX works, there is more stuff that needs to be hooked up
(like an additional SSRC etc). Also that patch was incomplete
and only covered 1 out of three places where RTX was filtered out.

(landed in #419)


KaptenJansson:
LGTM

PR #433, opened 2017-02-03
Edge: remove workaround that required track cloning

fippo:
removes a workaround that required cloning the track in addStream
which made muting via .enabled very hard.

Less bugs. This one was _very_ annoying


KaptenJansson (approved):
LGTM

PR #435, opened 2017-02-05
edge: add icegatheringstatechange event

fippo:
adds the ice gathering state change event in Edge.

This would be shim-able in other browsers (by listening for
onicecandidate with the null candidate) but does not seem
important enough.


KaptenJansson (approved):
LGTM

Issue #438, opened 2017-02-08
URL.createObjectURL shim doesn't work with setAttribute


fippo:
@jan-ivar found a repro:
https://jsfiddle.net/4sq7cors/

Apparently out.setAttribute isn't affected by the shim. What the ...

fippo:
probably the issue is rather that the src setter isn't called in this case. Why have one way to do it when you an have two :-(

fippo:
fixed by #439

PR #439, opened 2017-02-10
Shim Element.setAttribute('src', URL.createObjectURL(stream)).

jan-ivar:
**Description**
Fix for https://github.com/webrtc/adapter/issues/438.

**Purpose**
Make URL.createObjectURL shim work with element.setAttribute('src', ...).

### DIFF
--- a/src/js/utils.js
+++ b/src/js/utils.js
@@ -161,12 +161,11 @@ var utils = {
       }
     });
 
-    var nativeSetAttribute = Element.prototype.setAttribute;
-    Element.prototype.setAttribute = function() {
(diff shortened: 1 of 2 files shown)


KaptenJansson:
Thanks for this, would a test be possible as well since this would be hard to detect if it breaks in the wild?

jan-ivar:
Also narrowed shim to HTMLMediaElement (forgot to say in commit msg). Thanks @fippo!

Test coming up.

jan-ivar:
Test is up (forgot to say).

jan-ivar:
@KaptenJansson Test is up (forgot to say).

KaptenJansson:
Thank you! LGTM

PR #441, opened 2017-02-11
test: rewrite extractVersion tests

fippo:
rewrites the extractVersion test. Also changes the regexp to use \d+ instead of [0-9]+

### DIFF
--- a/src/js/utils.js
+++ b/src/js/utils.js
@@ -68,18 +68,18 @@ var utils = {
     if (navigator.mozGetUserMedia) {
       result.browser = 'firefox';
       result.version = this.extractVersion(navigator.userAgent,
-          /Firefox\/([0-9]+)\./, 1);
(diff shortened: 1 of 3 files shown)


fippo:
@alvestrand since I know you like tests... ;-)

meta question... I marked some tests with "by design" because we expect to call extractVersion only after getting a general idea of what browser we test for. Should we change that (not in this PR)?

KaptenJansson:
@fippo sorry for this but you could resolve the conflicts? There is a mix between /d and [0-9] for regexp and I'm not sure if there is a reason to use the one over the other or it just a matter of changing it where you update the code. I do not want to risk breaking things.

fippo:
went for \d+ everywhere

KaptenJansson:
Awesome, thanks! LGTM

PR #442, opened 2017-02-13
edge: add partial support for unified plan [dont merge]

fippo:
adds partial support for Unified Plan in Edge.
This is limited to creating offers after calling addStream twice (or more).

On the receiving end this adds support for parsing MSID and overrides
the native stream and track ids with the stream and track ids from the
SDP.

### DIFF
--- a/src/js/edge/edge_shim.js
+++ b/src/js/edge/edge_shim.js
@@ -532,7 +532,7 @@ var edgeShim = {
     window.RTCPeerConnection.prototype.setRemoteDescription =
         function(description) {
           var self = this;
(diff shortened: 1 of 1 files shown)


fippo:
I do have a bunch of tests in [this branch](https://github.com/fippo/adapter/tree/edge-latte) but they're not ready for upstreaming yet.

fippo:
don't merge just yet, i'm getting invalidaccesserror on the pc1 sample with this. In particular it looks this generates an sdp with a video and a recvonly audio m-line.

Will go back to the drawing board.

Issue #447, opened 2017-02-16
The MediaStreamTrack provided is not an element of a MediaStream that's currently in the local streams set


voiceactivity:
**Browsers and versions affected**
Chrome, Firefox

**Description**
Updated to the latest version of the adapter.js (master). Run the app and it fails during the following call: pc.addStream(localStream);

I found that this code responsible for that:
adapter.js:

chromeShim.shimGetSendersWithDtmf(); // line 550
...
  shimGetSendersWithDtmf: function() {
    if (typeof window === 'object' && window.RTCPeerConnection &&
        !('getSenders' in RTCPeerConnection.prototype) &&
        'createDTMFSender' in RTCPeerConnection.prototype) {
      RTCPeerConnection.prototype.getSenders = function() {
        return this._senders;
      };
      var origAddStream = RTCPeerConnection.prototype.addStream;

      RTCPeerConnection.prototype.addStream = function(stream) {
        var pc = this;
        pc._senders = pc._senders || [];
        origAddStream.apply(pc, [stream]);
        stream.getTracks().forEach(function(track) {
          pc._senders.push({
            track: track,
            get dtmf() {
              if (!this._dtmf) {
                this._dtmf = pc.createDTMFSender(track); //line 670 Error
              }
              return this._dtmf;
            }
          });
        });
      };
    }
  }, 

**Steps to reproduce**
Create the real peer connection between two peers with signaling (not the toy from the example page).
Try to establish connection between peers. 

When I commented out the aforementioned code, the error gone.

**Expected results**
1. LocalStream assigned to the localVideo.
2. PeerConnection create with all required callbacks.
3. pc.addStream(localStream) should start the negotiation between peers.
4. pc.onaddstream -> add the remote stream to the remoteVideo
5. Video chat started.

**Actual results**
1. UserMedia assigned to the local video.
2. During adding the localStream to the pc (pc.addStream(localStream)) the following error occurs: 

Failed to execute 'createDTMFSender' on 'RTCPeerConnection': The MediaStreamTrack provided is not an element of a MediaStream that's currently in the local streams set.
...Error: Failed to execute 'createDTMFSender' on 'RTCPeerConnection': The MediaStreamTrack provided is not an element of a MediaStream that's currently in the local streams set.
    at Object.get dtmf [as dtmf] (adapter.js:670)
    ...

3. Remote peer doesn't receive oncandidate request. All communication chain between peers broken.

jan-ivar:
[...]

PR #448, opened 2017-02-16
Make sender.dtmf be null for video tracks + shim removeStream.

jan-ivar:
**Description**
 * `sender.dtmf` should be `null` for video tracks.
 * Also shims `removeStream` so senders don't accumulate after renegotiations.

**Purpose**
Fix for https://github.com/webrtc/adapter/issues/447 + make it more robust for real use.


KaptenJansson:
LGTM

Issue #403, opened 2016-12-20
Add test for #401


KaptenJansson:
Spawned from #401

fippo:
we're done here (-:

PR #453, opened 2017-02-21
edge: add mocha unit tests

fippo:
also fixes an issue with the sorting of tracks. Now it does audio-video-audio-video.

The tests contain some unified-plan tests which are currently deactivated.

When hta is back we should talk about whether those tests make sense in web-platform-tests.


fippo:
also covers #403

KaptenJansson (approved):
Elegant! LGTM

PR #454, opened 2017-02-22
edge: test for ice server filtering

fippo:
**Description**
adds a unit test for the filtering of ICE servers.

**Purpose**
moar tests


KaptenJansson (approved):
LGTM

PR #456, opened 2017-03-01
edge: add gatheringstatechange event for gathering

fippo:
adds a missing event for the ice gathering state change
to "gathering".


KaptenJansson:
LGTM

PR #458, opened 2017-03-04
edge: add RTX

fippo:
this adds parsing and serialization for all things RTX
and removes the pieces of code removed rtx from the list of
codecs.

Took a bit longer to test since simplewebrtc was still stuck on adapter 1.x and 2.x deep in the dependencies but... seems to work just fine.

### DIFF
--- a/src/js/edge/edge_shim.js
+++ b/src/js/edge/edge_shim.js
@@ -264,11 +264,39 @@ var edgeShim = {
             headerExtensions: [],
             fecMechanisms: []
           };
+
+          var findCodecByPayloadType = function(pt, codecs) {
+            pt = parseInt(pt, 10);
+            for (var i = 0; i < codecs.length; i++) {
+              if (codecs[i].payloadType === pt ||
+                  codecs[i].preferredPayloadType === pt) {
+                return codecs[i];
+              }
+            }
+          };
+
+          var rtxCapabilityMatches = function(lRtx, rRtx, lCodecs, rCodecs) {
+            var lCodec = findCodecByPayloadType(lRtx.parameters.apt, lCodecs);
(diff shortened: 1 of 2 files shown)


src/js/edge/edge_shim.js
```diff
@@ -14,10 +14,10 @@ var browserDetails = require('../utils').browserDetails;
 // sort tracks such that they follow an a-v-a-v...
 // pattern.
 function sortTracks(tracks) {
-  let audioTracks = tracks.filter(function(track) {
+  var audioTracks = tracks.filter(function(track) {
```
fippo: this was a whoops. We are still trying to be ES5, right? Then why didn't eslint yell at me...

test/unit/edge.js
```diff
@@ -755,5 +763,64 @@ describe('Edge shim', () => {
...
+            expect(answer.sdp).not.to.contain('a=ssrc-group:FID');
+            done();
+          });
+        });
+      });
+
+      describe('with a local track', () => {
+        beforeEach(() => {
+          const videoTrack = new MediaStreamTrack();
+          videoTrack.kind = 'video';
```
KaptenJansson, line 805: Why use `const` when you change a property?

KaptenJansson (changes requested):
LGTM with nit

test/unit/edge.js
fippo: const is basically "I am not going to reassign this" -- https://mathiasbynens.be/notes/es6-const
Not like the c++ const "I am not going to change this". Sadly... I miss c++ const.

KaptenJansson: aha, thanks for clarifying!

KaptenJansson (approved):

KaptenJansson:
LGTM

PR #459, opened 2017-03-04
edge: dont use ES6 let in sortTracks

fippo:
separated from #458 

this was a whoops. We are still trying to be ES5, right? Then why didn't eslint yell at me...


KaptenJansson:
My bad as well, I think it's fine for tests and test frameworks but maybe not yet in the actual adapter code since it introduces yet another compatibility layer (not sure how much of the es6 spec all the browsers support).

LGTM

fippo:
let and const should be in all browsers we support. At some point we should consider moving to ES6 though and generate adapter-latest with a es5 transpiler.

At some point ;-)

PR #460, opened 2017-03-04
edge: fix dіrectionality attributes

fippo:
fixes an issue with the direction attribute and unskips three tests.

### DIFF
--- a/src/js/edge/edge_shim.js
+++ b/src/js/edge/edge_shim.js
@@ -656,13 +656,16 @@ var edgeShim = {
                 ssrc: (2 * sdpMLineIndex + 2) * 1001
               }];
 
-              rtpReceiver = new RTCRtpReceiver(transports.dtlsTransport, kind);
+              if (direction === 'sendrecv' || direction === 'sendonly') {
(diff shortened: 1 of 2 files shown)


src/js/edge/edge_shim.js
```diff
@@ -656,13 +656,16 @@ var edgeShim = {
...
 
-              track = rtpReceiver.track;
-              receiverList.push([track, rtpReceiver]);
-              // FIXME: not correct when there are multiple streams but that is
-              // not currently supported in this shim.
-              stream.addTrack(track);
+                track = rtpReceiver.track;
+                receiverList.push([track, rtpReceiver]);
+                // FIXME: not correct when there are multiple streams but that
+                // is not currently supported in this shim.
```
KaptenJansson, line 666: What happens in this case? Throw an error or does it silently ignore the other streams?

fippo: I have a branch for unified plan support which also improves the behaviour here

KaptenJansson: I see

KaptenJansson:
LGTM

PR #463, opened 2017-03-07
edge: create offers with bundle by default

fippo:
creates offers with a a=group:BUNDLE line by default.
Non-bundled offers can be achieved by ѕetting bundlePolicy
to max-compat.

(yes, this was easier than fixing a bug in my usage...)


fippo:
(still waiting for a edge webdriver version that matches my edge version before I can run the tests)

fippo:
tests run with one failure (the ice gathering state change test) which is unrelated. ready-to-merge

KaptenJansson (changes requested):
LGTM

KaptenJansson (approved):

PR #465, opened 2017-03-10
Edge lowercase algo

davies147:
Edge requires Dtls fingerprint 'algorithm' parameter to be lower case

Fixes an instance of the c004e001 error. When using Asterisk generated SDP, the line:
  a=fingerprint:SHA-256 ...
is RFC valid, but results in an algorithm of 'SHA-256'; Edge requires 'sha-256'


davies147:
Oops, apologies for the many unwanted commits - all except the last one are no-ops because I branched from the wrong place.

davies147:
Not entirely sure why Travis failed that...

fippo:
I think i'd rather have this in sdp.js -- should be a 1-line change in https://github.com/fippo/sdp/blob/master/sdp.js#L239

davies147:
@fippo - I agree in terms of clean code, and that is initially how I fixed it, but "fixing" something that is not broken just because Edge is stupid? That burns :( 

OTOH... Will you update https://github.com/fippo/sdp/blob/master/sdp.js#L239 or should I submit a pull request?

fippo:
@davies147 I saw an opportunity to increase test coverage -- https://github.com/fippo/sdp/pull/7 :-)

davies147:
Closing this PR because the change has been made in https://github.com/fippo/sdp/blob/master/sdp.js#L239 as discussed above.

PR #467, opened 2017-03-13
edge: add partial support for unified plan

fippo:
adds partial support for Unified Plan in Edge.
This is limited to creating offers after calling addStream multiple times.
Renegotiation is not supported.

On the receiving end this adds support for parsing MSID and overrides
the native stream and track ids with the stream and track ids from the
SDP.

### DIFF
--- a/src/js/edge/edge_shim.js
+++ b/src/js/edge/edge_shim.js
@@ -622,7 +622,7 @@ var edgeShim = {
     window.RTCPeerConnection.prototype.setRemoteDescription =
         function(description) {
           var self = this;
-          var stream = new MediaStream();
+          var streams = {};
           var receiverList = [];
           var sections = SDPUtils.splitSections(description.sdp);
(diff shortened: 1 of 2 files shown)


src/js/edge/edge_shim.js
```diff
@@ -1104,7 +1147,7 @@ var edgeShim = {
       tracks.forEach(function(mline, sdpMLineIndex) {
         var transceiver = transceivers[sdpMLineIndex];
         sdp += SDPUtils.writeMediaSection(transceiver,
-            transceiver.localCapabilities, 'offer', self.localStreams[0]);
+            transceiver.localCapabilities, 'offer', mline.stream);
```
fippo, line 1150: this should be named mline.track. But I can fix that later, internal detail

KaptenJansson: OK

KaptenJansson (approved):
LGTM

PR #468, opened 2017-03-13
test: move log supression test

fippo:
makes the log suppression test a unit test (since it doesn't require webdriver)


KaptenJansson:
LGTM

PR #470, opened 2017-03-13
edge: RTX bugfix

fippo:
Edge was responding to an offer which did not contain RTX with
a ssrc-group line and an additional ssrc.


KaptenJansson (approved):
LGTM

PR #473, opened 2017-03-17
test: rewrite ontrack test

fippo:
rewrites the ontrack test to use just a single peerconnection, similar
to how the web-platform-tests test works.

This also showed that we had a JS exception in chrome which was silently swallowed :-/


KaptenJansson:
LGTM

Issue #472, opened 2017-03-17
test: firefox fails due to lack of removeStream


fippo:
```
# dtmf
ok 103 Page loaded
not ok 104 PeerConnection failure: TypeError: pc1.removeStream is not a function
  ---
    operator: fail
    at: seleniumHelpers.loadTestPage.then.then (/home/fippo/webrtc/adapter/test/test.js:1539:11)
  ...
```
@jan-ivar can you take a look? Saw this when running tests locally with FF52... which doesn't have removeStream so the test is bad probably

jan-ivar:
Whoops, my bad. Tried [too hard to make the test work with Chrome](https://github.com/webrtc/adapter/commit/51d0a52190d579076e678bd70675f4ec9a2939dd).

PR #474, opened 2017-03-17
Fix test of getSenders polyfill in Chrome to not break FF52.

jan-ivar:
**Description**
Fix for https://github.com/webrtc/adapter/issues/472.

**Purpose**
Make test work for FF52 again.


KaptenJansson:
LGTM

PR #481, opened 2017-03-24
edge: export shimReplaceTrack

fippo:
I wonder if we should just do ```module.exports =``` instead of creating an object that is only used in a single place. Same for other browsers...


KaptenJansson:
I think I did that initially but something did not work. Also currently it's easy to disable a specific API shim.

KaptenJansson:
LGTM

PR #482, opened 2017-03-30
chrome: backfill RTCIceServer.urls from .url

fippo:
apparently there is a significant number of folks who still have
not upgraded from RTCIceServer.url to RTCIceServer.urls.
Let adapter do this for them.

cc @foolip

### DIFF
--- a/src/js/chrome/chrome_shim.js
+++ b/src/js/chrome/chrome_shim.js
@@ -165,6 +165,35 @@ var chromeShim = {
           }
         });
       }
+    } else {
+      // migrate from non-spec RTCIceServer.url to RTCIceServer.urls
+      var OrigPeerConnection = RTCPeerConnection;
+      window.RTCPeerConnection = function(pcConfig, pcConstraints) {
+        if (pcConfig && pcConfig.iceServers) {
+          var newIceServers = [];
+          for (var i = 0; i < pcConfig.iceServers.length; i++) {
+            var server = pcConfig.iceServers[i];
+            if (!server.hasOwnProperty('urls') &&
+                server.hasOwnProperty('url')) {
+              console.warn('RTCIceServer.url is deprecated! Use urls instead.');
+              server = JSON.parse(JSON.stringify(server));
+              server.urls = server.url;
+              newIceServers.push(server);
+            } else {
+              newIceServers.push(pcConfig.iceServers[i]);
+            }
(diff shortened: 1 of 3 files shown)


fippo:
see also https://bugs.chromium.org/p/chromium/issues/detail?id=659133

jan-ivar:
With the exception of https://github.com/webrtc/adapter/pull/411 our motto for adapter.js has been to polyfill the spec on older browsers, not the other way around.

This would be the second issue to deviate from that, which may be fine, but we should have a plan to tear both band-aids off in a major version bump a year from now, so this library doesn't become an extended-support library, contributing to the spread of non-spec API usage.

src/js/chrome/chrome_shim.js
```diff
@@ -165,6 +165,28 @@ var chromeShim = {
...
+    } else {
+      // migrate from non-spec RTCIceServer.url to RTCIceServer.urls
+      var OrigPeerConnection = RTCPeerConnection;
+      window.RTCPeerConnection = function(pcConfig, pcConstraints) {
+        if (pcConfig && pcConfig.iceServers) {
+          for (var i = 0; i < pcConfig.iceServers.length; i++) {
+            var server = pcConfig.iceServers[i];
+            if (server.hasOwnProperty('url') &&
+                !server.hasOwnProperty('urls')) {
+              server.urls = server.url;
```
jan-ivar: Don't land this without at least a deprecation warning to console!

Also note you're modifying the passed-in object here. Might wanna modify a copy.

fippo:
@jan-ivar this is mostly an attempt to drive down the numbers for non-spec usage that @foolip is seeing (unsure if we can impact them) in order to justify removal of the non-spec behaviour.

Whether people update adapter often enough for this to make an impact... :-/

src/js/chrome/chrome_shim.js
fippo: done. Also added a console.warn inside the Edge shim. I assume we don't need it in Firefox

KaptenJansson:
How do we know the devs who are using non spec apis are using adapter? :)

fippo:
we don't. But as @jan-ivar said in the bug we can probably attribute the 60% urls to showing the right stuff in the samples!

KaptenJansson:
I see. LGTM

PR #483, opened 2017-04-04
Fix Chrome facingMode polyfill to work with some desktop devices.

jan-ivar:
**Description**
[Apparently](http://stackoverflow.com/q/43211991/918910) not working w/Surface Pro 4 which defaults to back camera (surprising the old polyfill).

**Purpose**
Also extended lease.


KaptenJansson:
LGTM

PR #486, opened 2017-04-05
test: pin geckodriver to 1.4.0

fippo:
to work around https://github.com/SeleniumHQ/selenium/issues/3625
(needed until a new selenium version comes around)


fippo:
doesn't fix the issues we see on travis but at least fixes the issues I saw locally. Thanks @dminor!

KaptenJansson:
LGTM

PR #487, opened 2017-04-06
edge: fix bundle

fippo:
when using bundle this did not change the transport of the RTPSender / RTPReceiver
causing it to use transport that always remained in the `new` state.


fippo:
needs patch release, regression

fippo:
somewhat wondering if a video.readystate test would have caught this... the ice connection was up but no video was flowing... https://github.com/fippo/testbed failed so this is catch-able. Will add a test later.

KaptenJansson:
Wow that's really hard to catch just by looking at the code. Glad you found it though. Yeah, it's not really unit test testable. LGTM.

PR #488, opened 2017-04-06
edge: move RTCPeerConnection shim to separate file

fippo:
this makes it easier to reuse outside adapter.js and is the way to letting me split this out the "wicked shim" for easier releases.

(based on #487 but wait until after the patch version please)


KaptenJansson:
LGTM

PR #491, opened 2017-04-07
test: add videoElement.loadedmetadata test

fippo:
adds a test which is similar to the web-platform-test simplecall test
  https://github.com/w3c/web-platform-tests/blob/master/webrtc/simplecall.html

Would have probably caught #487 (but let me verify with the buggy version before merging)

Even though Harald is ooo I think he would like moar tests ;-)

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -1457,6 +1457,128 @@ test('Basic connection establishment with datachannel', function(t) {
   });
 });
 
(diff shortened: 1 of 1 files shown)


fippo:
hrm. The test does not catch the failure :-/
Shall we have it anyway because we like tests?

KaptenJansson:
I agree, moar tests! LGTM

Issue #492, opened 2017-04-07
webrtc-adapter is not getting last device for facingMode environment anymore on Chrome


nicolashenry:
**Browsers and versions affected**
Chrome Version 57.0.2987.132

**Description**
webrtc-adapter 3.3 is not getting last device for facingMode environment anymore when chrome which lies with facingMode is detected. This change seems to have been done here :
```
-            var back = devices.find(function(d) {
-              return d.label.toLowerCase().indexOf('back') !== -1;
-            }) || (devices.length && devices[devices.length - 1]);
-            if (back) {
-              constraints.video.deviceId = face.exact ? {exact: back.deviceId} :
-                                                        {ideal: back.deviceId};
+            var dev = devices.find(function(d) {
+              return d.label.toLowerCase().indexOf(match) !== -1;
+            });
+            if (dev) {
+              constraints.video.deviceId = face.exact ? {exact: dev.deviceId} :
+                                                        {ideal: dev.deviceId};
```
I think this part is missing in 3.3 for back camera : 
```}) || (devices.length && devices[devices.length - 1]);```

**Steps to reproduce**
Use getUserMedia with Chrome on android with ```facingMode: 'environment'```

**Expected results**
To get the back camera

**Actual results**
Got the front camera

fippo:
@jan-ivar can you take a look?

jan-ivar:
@nicolashenry What camera model do you have, if I may ask?

Yes I [removed](https://github.com/webrtc/adapter/pull/483) that part of the logic, after I learned the Surface Pro 4 defaults to the back camera, so I thought that assumption might not be good anymore.

Turns out I also messed up that patch for Surface Pro 4 (I should have tested for `"rear"` as well as `"back"`), so I need another patch. I'll put that logic back. Still curious what camera you have though.

jan-ivar:
@nicolashenry Can you confirm that https://github.com/webrtc/adapter/issues/495 fixes your problem?

nicolashenry:
@jan-ivar I had to launch grunt task to regenerate adapter.js from your branch. It works perfectly using the generated file. :+1:

PR #495, opened 2017-04-08
Fix facingMode to look for 'rear' as well, or use last cam again.

jan-ivar:
**Description**
Fix for https://github.com/webrtc/adapter/issues/492.

**Purpose**
Second attempt to make it pick the back camera on Surface Pro 4. Also bring back the "use last cam" hack removed in that patch, as it's apparently still useful (broke some other cam as reported).

### DIFF
--- a/src/js/chrome/getusermedia.js
+++ b/src/js/chrome/getusermedia.js
@@ -80,22 +80,27 @@ module.exports = function() {
             navigator.mediaDevices.getSupportedConstraints().facingMode &&
             !getSupportedFacingModeLies)) {
         delete constraints.video.facingMode;
-        var match;
+        var matches;
         if (face.exact === 'environment' || face.ideal === 'environment') {
-          match = 'back';
+          matches = ['back', 'rear'];
         } else if (face.exact === 'user' || face.ideal === 'user') {
-          match = 'front';
+          matches = ['front'];
         }
-        if (match) {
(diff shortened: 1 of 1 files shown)


fippo (approved):

src/js/chrome/getusermedia.js
```diff
@@ -80,22 +80,27 @@ module.exports = function() {
             navigator.mediaDevices.getSupportedConstraints().facingMode &&
             !getSupportedFacingModeLies)) {
         delete constraints.video.facingMode;
-        var match;
+        var matches;
         if (face.exact === 'environment' || face.ideal === 'environment') {
-          match = 'back';
+          matches = ['back', 'rear'];
```
KaptenJansson, line 85: Do we have a list of other facingModes? Looking at the spec, none of these are there. https://w3c.github.io/mediacapture-main/#idl-def-videofacingmodeenum 

Do we know why they are adding non spec facingModes? Can we spread the knowledge of the spec somehow?

fippo: this is different, it looks for "back" or "rear" in the camera label. Bad hack but... ;-)

KaptenJansson: Ow right, I missed that.

KaptenJansson (approved):
LGTM

PR #496, opened 2017-04-10
edge: throw some SLD/SRD state errors

fippo:
as described in https://tools.ietf.org/html/draft-ietf-rtcweb-jsep-18#section-5.5
and https://tools.ietf.org/html/draft-ietf-rtcweb-jsep-18#section-5.6

(merging at next minor release is sufficient)

### DIFF
--- a/src/js/edge/rtcpeerconnection_shim.js
+++ b/src/js/edge/rtcpeerconnection_shim.js
@@ -148,6 +148,19 @@ function getCommonCapabilities(localCapabilities, remoteCapabilities) {
   return commonCapabilities;
 }
 
+// is action=setLocalDescription with type allowed in signalingState
+function isActionAllowedInSignalingState(action, type, signalingState) {
+  return {
+    offer: {
+      setLocalDescription: ['stable', 'have-local-offer'],
+      setRemoteDescription: ['stable', 'have-remote-offer']
+    },
+    answer: {
+      setLocalDescription: ['have-remote-offer', 'have-local-pranswer'],
+      setRemoteDescription: ['have-local-offer', 'have-remote-pranswer']
+    }
+  }[type][action].indexOf(signalingState) !== -1;
+}
 
 module.exports = function(edgeVersion) {
   var RTCPeerConnection = function(config) {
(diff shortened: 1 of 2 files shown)


src/js/edge/rtcpeerconnection_shim.js
```diff
@@ -478,6 +478,22 @@ module.exports = function(edgeVersion) {
 
   RTCPeerConnection.prototype.setLocalDescription = function(description) {
     var self = this;
+
+    if ((description.type === 'offer' &&
```
KaptenJansson: For the sake of readability I would say the number of checks are on the limit, WDYT about putting them in a function with a descriptive name that returns a bool?

src/js/edge/rtcpeerconnection_shim.js
```diff
@@ -588,6 +604,22 @@ module.exports = function(edgeVersion) {
 
   RTCPeerConnection.prototype.setRemoteDescription = function(description) {
     var self = this;
+
+    if ((description.type === 'offer' &&
```
KaptenJansson: Same as previous comment

src/js/edge/rtcpeerconnection_shim.js
fippo: hrm. essentially that should be a function that checks if the state transition is allowed. I like that idea!

fippo: done

fippo:
@MirkoBonadei can you take a look at this on too? (give me an inch... :-))

src/js/edge/rtcpeerconnection_shim.js
```diff
@@ -478,6 +491,18 @@ module.exports = function(edgeVersion) {
 
   RTCPeerConnection.prototype.setLocalDescription = function(description) {
     var self = this;
+
+    if (!isActionAllowedInSignalingState('setLocalDescription',
+        description.type, this.signalingState)) {
+      var e = new Error('Can not set remote ' + description.type +
```
MirkoBonadei: I guess the error message should start with: 'Can not set local ', right?

MirkoBonadei (changes requested):
lgtm, just a question about the error message.

src/js/edge/rtcpeerconnection_shim.js
fippo: 🦅 👁 

fixed + squashed!

PR #497, opened 2017-04-10
edge: move getCommonCapabilities out of the RTRCPeerConnection object

fippo:
technically this is a breaking change... however that API has intentionally been marked as private by using a underscore...


fippo:
@MirkoBonadei this one is easy ;-)

MirkoBonadei:
lgtm

PR #502, opened 2017-04-13
Add nominal addTrack shim for Chrome.

jan-ivar:
**Description**
Fix for https://github.com/webrtc/samples/pull/888.

**Purpose**
Make webrtc samples set a good example.

### DIFF
--- a/src/js/chrome/chrome_shim.js
+++ b/src/js/chrome/chrome_shim.js
@@ -68,7 +68,8 @@ var chromeShim = {
       var origAddStream = RTCPeerConnection.prototype.addStream;
       var origRemoveStream = RTCPeerConnection.prototype.removeStream;
 
-      RTCPeerConnection.prototype.addTrack = function(track, ...streams) {
+      RTCPeerConnection.prototype.addTrack = function(track) {
+        var streams = [].slice.call(arguments, 1);
         var pc = this;
         if (pc.signalingState === 'closed') {
           throw new DOMException(
--- a/src/js/chrome/chrome_shim.js
+++ b/src/js/chrome/chrome_shim.js
@@ -53,8 +53,13 @@ var chromeShim = {
     }
   },
 
-  shimGetSendersWithDtmf: function() {
+  // This shim is only sufficient to substitute existing addStream use with
+  // addTrack in existing samples. It does not add the full power of addTrack,
+  // and tries to fail if used that way.
+
+  shimAddTrackAndSendersWithDtmf: function() {
     if (typeof window === 'object' && window.RTCPeerConnection &&
+        !('addTrack' in RTCPeerConnection.prototype) &&
         !('getSenders' in RTCPeerConnection.prototype) &&
         'createDTMFSender' in RTCPeerConnection.prototype) {
       RTCPeerConnection.prototype.getSenders = function() {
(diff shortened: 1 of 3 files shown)


jan-ivar:
@fippo PTAL.

src/js/chrome/chrome_shim.js
```diff
@@ -63,24 +68,62 @@ var chromeShim = {
...
+                this._dtmf = pc.createDTMFSender(track);
+              } else {
+                this._dtmf = null;
+              }
+            }
+            return this._dtmf;
+          }
+        };
+        pc._senders.push(sender);
+        // Only call addStream once all tracks of a stream have been added.
```
fippo, line 103: how about calling addStream inside a shimmed createOffer before calling the native createOffer?

jan-ivar: I worry about leaving the peer connection in a weird state for that long. Is there a particular use-case this would help with?

jan-ivar: Oh, and because this should trigger `pc.onnegotiationneeded`, that's why.

fippo:
the reason we didn't add addTrack a while back that we wanted to take care the scenario where only the  audio track of an AV stream is added.

The other alternative to do would be to create a clone of the stream that has the same id (see the edge rtcpeerconnection shim for overriding the id getter of MediaStream). That allows us much better control for this scenario but raises some issues for things like comparing the result of getLocalStreams (legacy, ugly) with the stream originally added). We might also have to do some clever sdp mangling to retain the id of the stream added in the SDP msid properties. This requires sdp mangling which is a rabbit hole i'd rather not enter.

fippo:
old discussion in https://github.com/webrtc/adapter/pull/182

jan-ivar:
FWIW I don't think anyone would expect `getLocalStreams()` to return anything sensible for `addTrack`. If they do they're definitely off-spec IMHO.

I could improve the polyfill by wrapping the subset of tracks in a new MediaStream instead of failing if you prefer, and that should work, but yeah the illusion is only so deep.

src/js/chrome/chrome_shim.js
```diff
@@ -63,25 +68,59 @@ var chromeShim = {
       var origAddStream = RTCPeerConnection.prototype.addStream;
       var origRemoveStream = RTCPeerConnection.prototype.removeStream;
 
-      RTCPeerConnection.prototype.addStream = function(stream) {
+      RTCPeerConnection.prototype.addTrack = function(track, ...streams) {
```
fippo: is ...streams ES6? unless we break es5 compat we should be looking at arguments.length here.

src/js/chrome/chrome_shim.js
```diff
@@ -63,25 +68,60 @@ var chromeShim = {
...
+        pc._senders.push(sender);
+        // Only call addStream once all tracks of a stream have been added.
+        var stream = streams[0];
+        var tracks = stream.getTracks();
+        if (tracks.every(t => pc._senders.find(sndr => sndr.track === t))) {
+          origAddStream.apply(pc, [stream]);
+        }
+        // Emit an error if this doesn't happen by the end of this run.
+        Promise.resolve().then(() => {
+          if (!tracks.every(t => pc._senders.find(sndr => sndr.track === t))) {
```
fippo, line 111: arrow functions. Yes, we should do a compilation step...

fippo:
@jan-ivar yours or mine? ;-)

jan-ivar:
@fippo Let's go with yours. Mine tried hard to keep the stream.id intact, but that illusion breaks down as soon as someone does an extra `stream.addTrack(track)`, and then it's a mess. Yours seems much cleaner.

KaptenJansson:
Close this then in favor of #503 ?

fippo:
yes

PR #503, opened 2017-04-13
chrome: shim pc.addTrack

fippo:
shims addTrack by creating a new "virtual" stream and adding the track to that.
Future calls to addTrack will add the track to the "virtual" stream and call
removeStream(); addStream() to trigger ONN. This seems to have no visible effect (and stats continue to be ok too)

alternative to #502 -- less elaborate (aka: doesn't bother to integrate with the senders object yet).
The main disadvantage of this approach is that
1) pc.getLocalStreams() returns different streams than those that were added. Using getLocalStreams and addTrack together seems unlikely. This can be solved by writing a getter for the mediastream id attribute.
2) the msid in the SDP is that of the virtual stream, not the stream added. This could be solved with marginal SDP mangling in a createOffer shim.

### DIFF
--- a/src/js/chrome/chrome_shim.js
+++ b/src/js/chrome/chrome_shim.js
@@ -80,6 +80,65 @@ var chromeShim = {
       var origAddStream = RTCPeerConnection.prototype.addStream;
       var origRemoveStream = RTCPeerConnection.prototype.removeStream;
 
+      if (!RTCPeerConnection.prototype.addTrack) {
+        RTCPeerConnection.prototype.addTrack = function(track, stream) {
+          var pc = this;
+          if (pc.signalingState === 'closed') {
+            throw new DOMException(
+              'The RTCPeerConnection\'s signalingState is \'closed\'.',
+              'InvalidStateError');
+          }
+          var streams = [].slice.call(arguments, 1);
+          if (streams.length !== 1 ||
+              !streams[0].getTracks().find(function(t) {
+                return t === track;
+              })) {
+            // this is not fully correct but all we can manage without
+            // [[associated MediaStreams]] internal slot.
+            throw new DOMException(
+              'The adapter.js addTrack polyfill only supports a single ' +
+              ' stream which is associated with the specified track.',
+              'NotSupportedError');
+          }
+
+          pc._senders = pc._senders || [];
+          var alreadyExists = pc._senders.find(function(t) {
+            return t.track === track;
+          });
+          if (alreadyExists) {
+            throw new DOMException('Track already exists.',
+                'InvalidAccessError');
+          }
+
+          pc._streams = pc._streams || {};
+          var oldStream = pc._streams[stream.id];
+          if (oldStream) {
+            oldStream.addTrack(track);
+            pc.removeStream(oldStream);
+            pc.addStream(oldStream);
+          } else {
+            var newStream = new MediaStream([track]);
+            pc._streams[stream.id] = newStream;
+            pc.addStream(newStream);
+          }
+
+          var sender = {
+            track: track,
+            get dtmf() {
+              if (this._dtmf === undefined) {
+                if (track.kind === 'audio') {
+                  this._dtmf = pc.createDTMFSender(track);
+                } else {
+                  this._dtmf = null;
+                }
+              }
+              return this._dtmf;
+            }
+          };
+          pc._senders.push(sender);
+          return sender;
+        };
+      }
       RTCPeerConnection.prototype.addStream = function(stream) {
         var pc = this;
         pc._senders = pc._senders || [];
--- a/test/test.js
+++ b/test/test.js
@@ -1465,6 +1465,415 @@ test('Basic connection establishment with bidirectional streams', function(t) {
   });
 });
 
+test('Basic connection establishment with addTrack', function(t) {
+  var driver = seleniumHelpers.buildDriver();
+
+  var testDefinition = function() {
+    var callback = arguments[arguments.length - 1];
+
+    var counter = 1;
+    window.testPassed = [];
+    window.testFailed = [];
+    var tc = {
+      ok: function(ok, msg) {
+        window[ok ? 'testPassed' : 'testFailed'].push(msg);
+      },
+      is: function(a, b, msg) {
+        this.ok((a === b), msg + ' - got ' + b);
+      },
+      pass: function(msg) {
+        this.ok(true, msg);
+      },
+      fail: function(msg) {
+        this.ok(false, msg);
+      }
+    };
+    var pc1 = new RTCPeerConnection(null);
+    var pc2 = new RTCPeerConnection(null);
+
+    pc1.oniceconnectionstatechange = function() {
+      if (pc1.iceConnectionState === 'connected' ||
+          pc1.iceConnectionState === 'completed') {
+        callback(pc1.iceConnectionState);
+      }
+    };
+
+    pc2.onaddstream = function(event) {
+      tc.pass('onaddstream called');
+      tc.ok(event.stream.getAudioTracks().length === 1, 'with an audio track');
+      tc.ok(event.stream.getVideoTracks().length === 1, 'with a video track');
+    };
+
+    var dictionary = obj => JSON.parse(JSON.stringify(obj));
+
+    var addCandidate = function(pc, event) {
+      if (event.candidate) {
+        event.candidate = dictionary(event.candidate);
+      }
+      pc.addIceCandidate(event.candidate).then(function() {
+        // TODO: Decide if we are interested in adding all candidates
+        // as passed tests.
+        tc.pass('addIceCandidate ' + counter++);
+      })
+      .catch(function(err) {
+        tc.fail('addIceCandidate ' + err.toString());
+      });
+    };
+    pc1.onicecandidate = function(event) {
+      addCandidate(pc2, event);
+    };
+    pc2.onicecandidate = function(event) {
+      addCandidate(pc1, event);
+    };
+
+    var constraints = {audio: true, video: true, fake: true};
+    navigator.mediaDevices.getUserMedia(constraints)
+    .then(function(stream) {
+      stream.getTracks().forEach(function(track) {
+        pc1.addTrack(track, stream);
+      });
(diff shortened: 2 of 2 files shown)


fippo:
I've tried SDP mangling in createOffer. However, this means that setLocalDescription will be called with a stream id which Chrome might consider unknown so I don't think this is a good idea.

@henbos @guidou can you take a look? In particular the question above and calling removeStream/addStream in rapid succession?

test/test.js
```diff
@@ -1726,7 +1726,7 @@ test('dtmf', t => {
 
     navigator.mediaDevices.getUserMedia({audio: true})
     .then(stream => {
-      pc1.addStream(stream);
+      pc1.addTrack(stream.getAudioTracks()[0], stream);
```
henbos: We should still have test coverage for addStream. Perhaps this test should be parameterized for the different ways of adding tracks.

src/js/chrome/chrome_shim.js
```diff
@@ -80,9 +109,33 @@ var chromeShim = {
               }
               return this._dtmf;
             }
+          };
+          pc._senders.push(sender);
+          return sender;
+        };
+      } else {
+        RTCPeerConnection.prototype.addStream = function(stream) {
```
henbos: Why are you not shiming addStream if there exists an addTrack? Is the assumption that if addTrack is implemented, addStream will be implemented by using addTrack? Can you add a comment?

src/js/chrome/chrome_shim.js
```diff
@@ -63,12 +63,41 @@ var chromeShim = {
...
+          if (pc.signalingState === 'closed') {
+            throw new DOMException(
+              'The RTCPeerConnection\'s signalingState is \'closed\'.',
+              'InvalidStateError');
+          }
+          var streams = [].slice.call(arguments, 1);
+          if (streams.length !== 1 ||
+              !streams[0].getTracks().find(t => t === track)) {
+            throw new DOMException(
+              'The adapter.js addTrack polyfill only supports a single stream.',
```
henbos: Add "... which is associated with the specified track." to cover the OR case of the above if.

If I'm reading the spec correctly, the association between tracks and streams is based on senders internal streams slot, not whether or not the stream contains the track - but based on current implementation, it would be broken if stream didn't. Can you add a comment in an appropriate place?

src/js/chrome/chrome_shim.js
```diff
@@ -63,12 +63,41 @@ var chromeShim = {
...
+              'NotSupportedError');
+          }
+
+          pc._senders = pc._senders || [];
+          if (pc._senders.find(t => t === track)) {
+            throw new DOMException('Track already exists.',
+                'InvalidAccessError');
```
henbos: This is correct according to spec, but doesn't this imply we can't add a track to a new stream if it has already been added to the peer connection? Even though we can add and remove it from a MediaStream, whatever that means in absence of pc.addTrack or if pc.addTrack was used instead of pc.addStream.

src/js/chrome/chrome_shim.js
```diff
@@ -63,12 +63,41 @@ var chromeShim = {
...
+          var oldStream = pc._streams[stream.id];
+          if (oldStream) {
+            oldStream.addTrack(track);
+            pc.removeStream(oldStream);
+            pc.addStream(oldStream);
+          } else {
+            var newStream = new MediaStream([track]);
+            pc._streams[stream.id] = newStream;
+            pc.addStream(newStream);
+          }
```
henbos, line 123: We should have test coverage both for adding to an existing stream and a new stream?

src/js/chrome/chrome_shim.js
```diff
@@ -63,12 +63,41 @@ var chromeShim = {
...
+
+          pc._streams = pc._streams || {};
+          var oldStream = pc._streams[stream.id];
+          if (oldStream) {
+            oldStream.addTrack(track);
+            pc.removeStream(oldStream);
+            pc.addStream(oldStream);
+          } else {
+            var newStream = new MediaStream([track]);
```
henbos, line 120: Why is the streams argument ignored in favor of creating a new MediaStream? Would this work if the streams was an empty list (which we have earlier made sure it isn't by throwing an exception)?

test/test.js
```diff
@@ -1726,7 +1726,7 @@ test('dtmf', t => {
 
     navigator.mediaDevices.getUserMedia({audio: true})
```
henbos: In general, I am worried about the lack of test coverage in this PR. addTrack and addStream seem to have a lot of special cases (how was the track added, if added/removed, etc) and RTCPeerConnection contains stuff like stream getters. We should make sure calls can be made, streams are as expected, etc.

henbos (changes requested):

src/js/chrome/chrome_shim.js
fippo: because the stream might be a stream containing an audio + video track but only the audio track can be added using addTrack and adding the stream directly would result in both the audio and video track being added

test/test.js
fippo: so far this is a strawman. But yes, moar tests are definetly needed.

fippo: (and I wonder why i changed the dtmf test...)

src/js/chrome/chrome_shim.js
fippo: odd. I suppose my thinking was "if you want to use getSenders use addTrack". But... the current code doesn't make sense.

test/test.js
fippo: it turns out i changed it because it broke but now i'm shimming addStream with senders again so this change got reverted

test/test.js
fippo: what is the minimum test coverage we need?
1) call established + ice up with addTrack, using an audio-video stream
2) call established, ice up but not sending video when using an audio-only stream
3) addTrack with an audio track from streamA and a video track from streamB should result in ontrack with two different streams at the remote end.

fippo: added tests. I'm not fully happy with them, in particular they don't assert that for case 3 above there are two remote tracks

src/js/chrome/chrome_shim.js
fippo: is this sufficiently covered by testing with an a/v stream or do we need an audio->audio/video upgrade test?

src/js/chrome/chrome_shim.js
fippo: @jan-ivar btw, this (copied from you) is es6 and should compare t.track === track

@henbos can you give an example?

src/js/chrome/chrome_shim.js
fippo: hrm. Now you made me wonder if we could have an internal slot in our senders. Might make sense to add that.

Moving forward I assume you'll expose RTCRtpSender together with addTrack?

fippo:
test failure seems to be due to #508 which is now in beta? :-/

fippo:
hrm. So github doesn't allow me to say "i have addressed that feedback, please take another look"... @henbos manual ping ;-)

KaptenJansson:
@henbos ping

fippo:
ok... so I did a test. And found behaviour I object to :-/
Checkout this branch, open test/testpage.html

Paste this:
```
var v = document.createElement('video');
v.autoplay = true;
v.id = 'remoteVideo';
document.body.appendChild(v);
var pc1 = new RTCPeerConnection();
var pc2 = new RTCPeerConnection();
var audioStream = null;
pc1.onicecandidate = e => pc2.addIceCandidate(e.candidate);
pc2.onicecandidate = e => pc1.addIceCandidate(e.candidate);
pc2.onaddstream = function(e) { document.getElementById('remoteVideo').srcObject = e.stream; }
navigator.mediaDevices.getUserMedia({audio: true})
.then(stream => {
  audioStream = stream;
  stream.getTracks().forEach(t => {
    sender = pc1.addTrack(t, stream);
    console.log(sender);
  });
  return pc1.createOffer();
})
.then((offer) => {
  return pc2.setRemoteDescription(offer);
})
.then(() => {
  return pc1.setLocalDescription(pc2.remoteDescription);
})
.then(() => {
  return pc2.createAnswer();
})
.then(answer => {
  return pc1.setRemoteDescription(answer);
})
.then(() => {
  return pc2.setLocalDescription(pc1.remoteDescription);
});
```
then paste this
```
navigator.mediaDevices.getUserMedia({video: true})
.then(stream => {
  stream.getTracks().forEach(t => {
    audioStream.addTrack(t);
    sender = pc1.addTrack(t, audioStream);
    console.log(sender);
  });
  return pc1.createOffer();
})
.then((offer) => {
  return pc2.setRemoteDescription(offer);
})
.then(() => {
  return pc1.setLocalDescription(pc2.remoteDescription);
})
.then(() => {
  return pc2.createAnswer();
})
.then(answer => {
  return pc1.setRemoteDescription(answer);
})
.then(() => {
  return pc2.setLocalDescription(pc1.remoteDescription);
});
```
Expected result: the remote video is shown. It is not. getStats say it is received and pasting this actually shows it:
```
remoteVideo.srcObject = pc2.getRemoteStreams()[0]
```

This is fairly easy to work around by unconditionally setting srcObject i ontrack but... I think this has worked in the past and might be a regression from the native srcObject implementation.

I filed https://bugs.chromium.org/p/chromium/issues/detail?id=720258  with a simpler reproduction.

src/js/chrome/chrome_shim.js
```diff
@@ -63,6 +63,65 @@ var chromeShim = {
...
+          if (alreadyExists) {
+            throw new DOMException('Track already exists.',
+                'InvalidAccessError');
+          }
+
+          pc._streams = pc._streams || {};
+          var oldStream = pc._streams[stream.id];
+          if (oldStream) {
+            oldStream.addTrack(track);
+            pc.removeStream(oldStream);
```
fippo, line 117: TODO: this is to trigger ONN. Maybe try dispatchEvent(ONN)

henbos (approved):
LGTM

fippo:
rebased. Don't merge until we have had a patch release please, this is a minor.

KaptenJansson:
LGTM

PR #505, opened 2017-04-17
edge: implement addTrack

fippo:
implements RTCPeerConnection.addTrack using ORTC


fippo:
hold on, apparently this breaks some selenium tests :-/

fippo:
looks like RTCRtpSender silently dislikes being created with a null DTLS transport. Will need to figure out a workaround.

PR #506, opened 2017-04-17
Use native RTCPeerConnection API on Edge when available

oooookk7:
**Description**
Prevent overriding the original implemented `RTCPeerConnection` API when it is available.

**Purpose**
Since WebRTC API is introduced in the latest release version of edge as noted [here](https://blogs.windows.com/msedgedev/2017/01/31/introducing-webrtc-microsoft-edge/), it would be awesome to use the native implemented `RTCPeerConnection` API when available.

### DIFF
--- a/src/js/edge/rtcpeerconnection_shim.js
+++ b/src/js/edge/rtcpeerconnection_shim.js
@@ -71,6 +71,10 @@ function filterIceServers(iceServers, edgeVersion) {
 }
 
 module.exports = function(edgeVersion) {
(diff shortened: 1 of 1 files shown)


oooookk7:
I signed it!

fippo:
As noted in the blog post, this RTCPeerConnection implements a pre-1.0 "legacy" version of the API.
Among other things it doesn't implement ontrack so tests will fail.

oooookk7:
Makes sense.

PR #507, opened 2017-04-18
chrome: return empty array from getSenders

fippo:
makes getSenders return an empty array when called before addStream/addTrack.

No test. This belongs into a wpt test imo :-)

@jan-ivar PTAL


jan-ivar (approved):
Nice catch, thanks!

KaptenJansson:
LGTM

PR #511, opened 2017-04-19
edge: fix setRemoteDescription exception

### LINKED ISSUE #510: Unable to get property 'id' of undefined or null reference in Edge
Edge 15.15063
Chrome 58.0.3029.68

**Description**
Upon setting the remote answer from Chrome, here is the error received:

> SCRIPT5007: Unable to get property 'id' of undefined or null reference

And clicking on the line points to this:
```
receiverList.forEach(function(item) {
 var track = item[0];
 var receiver = item[1];
 if (stream.id !== item[2].id) { // Points to here
      return;
   }
  var trackEvent = new Event('track');
  trackEvent.track = track;
  trackEvent.receiver = receiver;
  trackEvent.streams = [stream];
  self.dispatchEvent(trackEvent);
  if (self.ontrack !== null) {
   window.setTimeout(function() {
     self.ontrack(trackEvent);
   }, 0);
  }
});
```

**Steps to reproduce**
1. Go to [appear.in](appear.in) on Chrome and on Edge 15.15063 with video and audio enabled
2. Open the console on Edge and notice the errors. Notice that Edge receives the video stream but Chrome doesn't.

fippo:
fixes #510

### DIFF
--- a/src/js/edge/rtcpeerconnection_shim.js
+++ b/src/js/edge/rtcpeerconnection_shim.js
@@ -807,17 +807,18 @@ module.exports = function(edgeVersion) {
         if (rtpReceiver &&
             (direction === 'sendrecv' || direction === 'sendonly')) {
           track = rtpReceiver.track;
-          receiverList.push([track, rtpReceiver]);
           if (remoteMsid) {
             if (!streams[remoteMsid.stream]) {
               streams[remoteMsid.stream] = new MediaStream();
             }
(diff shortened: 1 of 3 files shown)


MirkoBonadei:
lgtm

fippo:
if you could make a patch release just from this that would be great!

MirkoBonadei:
Ok. Thanks for adding tests.

PR #515, opened 2017-04-21
Convert module to ES module

pioug:
**Description**

- Convert module to ES module
- Bundle `adapter.js` using [Rollup](https://github.com/rollup/rollup)
- Include Babel in test runner

**Purpose**

Hello, I am experimenting Rollup and ES6 modules these days and I wanted to see how it can perform on a small library. If I did correctly, `adapter.js` bundled with Rollup is ~2 KB smaller (after minification) than the Browserify bundle. It's a little win 😅
- https://github.com/pioug/adapter/blob/rollup/release/adapter.rollup.min.js 
- https://github.com/pioug/adapter/blob/rollup/release/adapter.browserify.min.js

I am wondering if people working on this repo are interested by this approach. I understand that tooling is a kind of sensitive topic and changing tools is not always better. 

On my end, I can spend time to fully transition the repo to Rollup and ES modules (updating the CLI, removing Browserify, updating the tests, adding pipelines for the other builds).


MirkoBonadei:
@fippo what do you think?

fippo:
i'm not a fan of es2015 modules personally so I don't have an opinion.

Moving to the generated version to be something compiled is something i've considered for a while now, glad to hear that it even saves some bytes!

KaptenJansson:
Thanks for the suggestion but at the moment we do not think it's worth the effort.

Issue #514, opened 2017-04-21
Chrome not receiving video stream from Edge


oooookk7:
**Browsers and versions affected**
Edge 15.16179, Windows 10.0.16179 Build 16179 (Home)
Chrome 58.0.3029.81, Mac OSX 10.11.6 (El Capitan)

**Description**
It seems like Chrome doesn't receive any video stream from Edge even though ssrc sources are provided, and direction is `a=sendrecv`, the m= line is active and not rejected.

Here are the SDPs. The SDPs seems correct but pasting here as extra information.

Native RTCPeerConnection:
- Chrome answer: [native-chrome-answer.txt](https://github.com/webrtc/adapter/files/946276/native-chrome-answer.txt)
- Edge offer: [native-edge-offer.txt](https://github.com/webrtc/adapter/files/946275/native-edge-offer.txt)

Polyfill RTCPeerConnection:
- Chrome answer: [chrome-answer.txt](https://github.com/webrtc/adapter/files/946279/chrome-answer.txt)
- Edge offer: [edge-offer.txt](https://github.com/webrtc/adapter/files/946278/edge-offer.txt)

Some notes:
- TURN ICE servers was provided.
- When testing with the native RTCPeerConnection in Edge, the video stream is received, but not with the ORTC API polyfill in v3.3.3. I'm not sure this should also be raised issue in Edge as their bug since it seems like a browser issue, but opening here to be sure since it works on the native RTCPeerConnection but not the polyfill.
- ICE was connected, although sometimes with failures.
- Once Chrome had completed gathering, Chrome sends the end-of-candidate signal to Edge. I've tried with sending the end-of-candidate signal once there is no response from `.onicecandidate` handle on Chrome's end for 2s after the previous ICE candidate generated as well, but I doubt it might be the cause of the issue as well.

**Steps to reproduce**
1. Chrome opens Peer connection.
2. Edge opens Peer connection.
3. Edge sends remote offer with `{ offerToReceiveAudio: true, offerToReceiveVideo: true }`
4. Chrome sends remote answer.
5. Noticed that in `peerConnection.getStats()`, only audio packets is received from Edge. Attached `<video>` element `.srcObject` with MediaStream is blank.

**Expected results**
Video stream should be rendered on attached `<video>` element.

**Actual results**
Video stream not appearing or rendering on `<video>` element

fippo:
[...]

PR #516, opened 2017-04-25
test: pin selenium-webdriver

fippo:
due to incompabilities between selenium 3.4.0 and geckodriver 1.4.0

also checks if firefox tests still pass :-)

PR #517, opened 2017-04-25
edge: handle addStream after setRemoteDescription

fippo:
fixes https://github.com/webrtc/adapter/issues/514

The rtpSender was not created when the client code was doing this:
```
pc.setRemoteDescription(...)
.then(() => {
  pc.addStream(...);
  return pc.createAnswer();
})
```

(patch version after this please)

### DIFF
--- a/src/js/edge/rtcpeerconnection_shim.js
+++ b/src/js/edge/rtcpeerconnection_shim.js
@@ -718,27 +718,6 @@ module.exports = function(edgeVersion) {
           }
         }
 
-        // FIXME: look at direction.
-        if (self.localStreams.length > 0 &&
-            self.localStreams[0].getTracks().length >= sdpMLineIndex) {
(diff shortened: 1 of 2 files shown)


MirkoBonadei:
It seems that it is breaking a test on firefox stable.
(Ok, after this I will patch the version).

fippo:
I think that is the effect of #516 -- let me close and reopen to trigger travis again?

MirkoBonadei:
Ok, thanks!

Lenne231:
@fippo It seems to be that this fix got lost when moving https://github.com/otalk/rtcpeerconnection-shim out.

fippo:
whoops? taking a look...

fippo:
nope, the code is in there and seems to work. Can you show a jsfiddle where it doesn't?

PR #520, opened 2017-04-27
in strict mode cant update state. This fixes that.

mrgmatthews:
in Edge when the dtls was failing also getting an error 
"Assignment to read-only properties is not allowed in strict mode"


mrgmatthews:
I signed it!

MirkoBonadei:
@fippo Can you review? Thanks!

fippo (approved):

PR #509, opened 2017-04-19
chrome: lookup RTPReceiver for ontrack

fippo:
looks up the RTPReceiver for ontrack when pc.getReceivers is available.

fixes #508

PR #521, opened 2017-05-01
Adding shims for addStream and legacy callback APIs

youennf:
**Description**
Adding legacy APIs (addStream) and callback support for setRemoteDescription, setLocalDescription, addIceCandidate, createOffer and createAnswer for Safari.

**Purpose**
Support some legacy websites.

### DIFF
--- a/src/js/safari/safari_shim.js
+++ b/src/js/safari/safari_shim.js
@@ -13,6 +13,14 @@ var safariShim = {
   // TODO: check for webkitGTK+
   // shimPeerConnection: function() { },
 
+  shimAddStream: function() {
+    if (typeof window === 'object' && window.RTCPeerConnection &&
+        !('addStream' in window.RTCPeerConnection.prototype)) {
+      RTCPeerConnection.prototype.addStream = function(stream) {
+        stream.getTracks().forEach(track => this.addTrack(track, stream));
+      }
+    }
+  },
   shimOnAddStream: function() {
     if (typeof window === 'object' && window.RTCPeerConnection &&
         !('onaddstream' in window.RTCPeerConnection.prototype)) {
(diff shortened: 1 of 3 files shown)


youennf:
Failing test seems to be selenium-based. Not sure whether this is fine or whether something should be done here.

src/js/safari/safari_shim.js
```diff
@@ -43,7 +51,65 @@ var safariShim = {
...
+    var setRemoteDescription = prototype.setRemoteDescription;
+    var addIceCandidate = prototype.addIceCandidate;
+
+    prototype.createOffer = function(successCallback, failureCallback) {
+      var promise = createOffer.apply(this, [arguments[2]]);
+      if (!failureCallback) {
+        return promise;
+      }
+      promise.then(successCallback, failureCallback);
+      return Promise.resolve();
```
fippo, line 72: that doesn't look right. Shouldn't you return the lowercase promise? (also below)

fippo:
travis failure is due to #509

src/js/safari/safari_shim.js
```diff
@@ -43,7 +51,65 @@ var safariShim = {
...
+    }
+    var prototype = RTCPeerConnection.prototype;
+    var createOffer = prototype.createOffer;
+    var createAnswer = prototype.createAnswer;
+    var setLocalDescription = prototype.setLocalDescription;
+    var setRemoteDescription = prototype.setRemoteDescription;
+    var addIceCandidate = prototype.addIceCandidate;
+
+    prototype.createOffer = function(successCallback, failureCallback) {
+      var promise = createOffer.apply(this, [arguments[2]]);
```
fippo: I think you need something along the lines of https://github.com/webrtc/adapter/blob/master/src/js/chrome/chrome_shim.js#L297 to pass through the createOffer(offerOptions) stuff here.

src/js/safari/safari_shim.js
youennf: As per https://www.w3.org/TR/webrtc/#legacy-interface-extensions, createOffer step 7says "Return a promise resolved with undefined".
I guess that we should really return void but cannot due to WebIDL limitations, hence the current design.

src/js/safari/safari_shim.js
youennf: Right, I'll fix that!

src/js/safari/safari_shim.js
fippo: ah... that makes sense!

src/js/safari/safari_shim.js
jan-ivar: `options` is the first argument in the promise case, and the third argument in the callback case. This passes up the wrong argument in the promise case.

jan-ivar (changes requested):

fippo:
wrt to unit tests what do you think about this using chai and sinon-chai (happy to make a PR when you're done):
```
  describe('createOffer shim', () => {
    describe('options passing', () => {
      let stub;
      beforeEach(() => {
        stub = sinon.stub();
        window.RTCPeerConnection.prototype.createOffer = stub;
        shim.shimCallbacksAPI();
      });

      it('with no arguments', () => {
        pc.createOffer();
        expect(stub).to.have.been.calledWith(undefined);
      });

      it('with two callbacks', () => {
        pc.createOffer(null, null);
        expect(stub).to.have.been.calledWith(undefined);
      });

      it('with a non-function first argument', () => {
        pc.createOffer(1);
        expect(stub).to.have.been.calledWith(1);
      });

      it('with two callbacks and options', () => {
        pc.createOffer(null, null, 1);
        expect(stub).to.have.been.calledWith(1);
      });

      it('with two callbacks and two additional arguments', () => {
        pc.createOffer(null, null, 1, 2);
        expect(stub).to.have.been.calledWith(1);
      });
    });
  });
```
I've been wanting to add sinon for stubs for a long time...

fippo (approved):

MirkoBonadei:
lgtm

youennf:
Don't really know chai and sinon-chai. Happy to see how this would look like.

PR #524, opened 2017-05-02
test: use sinon and sinon-chai for unit tests

fippo:
uses [sinon](http://sinonjs.org/) and [sinon-chai](https://github.com/domenic/sinon-chai)
for unit testing

@youennf what do you think? I also noticed that the legacy createAnswer technically doesn't have answer options... see issue over on webrtc-pc


youennf:
LGTM!

KaptenJansson (approved):
Much cleaner! LGTM

PR #525, opened 2017-05-02
edge: improve negotiationneeded

fippo:
improves the maybeFireNegotiationNeeded implementation to work
better with addTrack in the future.

the requirement here is that ONN does not fire too early for this:
```
// let stream have an audio and a video track...

pc.onnegotationneeded = () => {
  // should be called after both tracks have been added
};
stream.getTracks().forEach((t) => {
  pc.addTrack(t);
})
```

hrm... maybe I can get @pthatcherg to review this? 👋

### DIFF
--- a/test/unit/edge.js
+++ b/test/unit/edge.js
@@ -1295,4 +1295,27 @@ describe('Edge shim', () => {
       });
     });
   });
+
+  describe('negotationneeded', () => {
+    it('fires asynchronously after addTrack', (done) => {
+      const pc = new RTCPeerConnection();
+
+      const audioTrack = new MediaStreamTrack();
+      audioTrack.kind = 'audio';
+      const videoTrack = new MediaStreamTrack();
+      videoTrack.kind = 'video';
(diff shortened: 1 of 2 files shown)


test/unit/edge.js
```diff
@@ -1152,4 +1152,30 @@ describe('Edge shim', () => {
...
+
+      pc.onnegotiationneeded = function(e) {
+        pc.createOffer()
+        .then((offer) => {
+          const sections = SDPUtils.splitSections(offer.sdp);
+          expect(sections.length).to.equal(3);
+          done();
+        });
+      };
+      pc.addStream(audioStream);
```
KaptenJansson: Maybe could update this to use addTrack now? Same for the next line? Then remove the comment on line 1157.

KaptenJansson (changes requested):

KaptenJansson (approved):

KaptenJansson:
LGTM

PR #526, opened 2017-05-05
all: add augmenting RTCIceCandidate constructor

fippo:
adds an RTCIceCandidate constructor which parses the fields as
described in https://w3c.github.io/webrtc-pc/#rtcicecandidate-interface


fippo:
Hrm... looks like I need a slightly different approach here. bbiab...

PR #527, opened 2017-05-05
all: add augmenting RTCIceCandidate constructor

fippo:
adds an RTCIceCandidate constructor which parses the fields as
described in https://w3c.github.io/webrtc-pc/#rtcicecandidate-interface

### DIFF
--- a/src/js/common.js
+++ b/src/js/common.js
@@ -1,0 +1,108 @@
+/*
+ *  Copyright (c) 2017 The WebRTC project authors. All Rights Reserved.
+ *
+ *  Use of this source code is governed by a BSD-style license
+ *  that can be found in the LICENSE file in the root of the source
+ *  tree.
+ */
+ /* eslint-env node */
+'use strict';
+
+var SDPUtils = require('sdp');
+
+// Wraps the peerconnection event eventNameToWrap in a function
+// which returns the modified event object.
+function wrapPeerConnectionEvent(eventNameToWrap, wrapper) {
+  if (!window.RTCPeerConnection) {
+    return;
+  }
+  var proto = window.RTCPeerConnection.prototype;
+  var nativeAddEventListener = proto.addEventListener;
+  proto.addEventListener = function(nativeEventName, cb) {
+    if (nativeEventName !== eventNameToWrap) {
+      return nativeAddEventListener.apply(this, arguments);
+    }
+    var wrappedCallback = function(e) {
+      cb(wrapper(e));
+    };
+    this._eventMap = this._eventMap || {};
+    this._eventMap[cb] = wrappedCallback;
+    return nativeAddEventListener.apply(this, [nativeEventName,
+      wrappedCallback]);
+  };
+
+  var nativeRemoveEventListener = proto.removeEventListener;
+  proto.removeEventListener = function(nativeEventName, cb) {
+    if (nativeEventName !== eventNameToWrap || !this._eventMap
+        || !this._eventMap[cb]) {
+      return nativeRemoveEventListener.apply(this, arguments);
+    }
+    var unwrappedCb = this._eventMap[cb];
+    delete this._eventMap[cb];
+    return nativeRemoveEventListener.apply(this, [nativeEventName,
+      unwrappedCb]);
+  };
+
+  Object.defineProperty(proto, 'on' + eventNameToWrap, {
+    get: function() {
+      return this['_on' + eventNameToWrap];
+    },
+    set: function(cb) {
+      if (this['_on' + eventNameToWrap]) {
+        this.removeEventListener(eventNameToWrap,
+            this['_on' + eventNameToWrap]);
+      }
+      this.addEventListener(eventNameToWrap,
+          this['_on' + eventNameToWrap] = cb);
+    }
+  });
+}
+
+module.exports = {
+  shimRTCIceCandidate: function() {
+    // foundation is arbitrarily chosen as an indicator for full support for
+    // https://w3c.github.io/webrtc-pc/#rtcicecandidate-interface
+    if (window.RTCIceCandidate && 'foundation' in
+        window.RTCIceCandidate.prototype) {
+      return;
+    }
+
+    var NativeRTCIceCandidate = window.RTCIceCandidate;
+    window.RTCIceCandidate = function(args) {
+      // Remove the a= which shouldn't be part of the candidate string.
+      if (typeof args === 'object' && args.candidate &&
+          args.candidate.indexOf('a=') === 0) {
+        args = JSON.parse(JSON.stringify(args));
+        args.candidate = args.candidate.substr(2);
+      }
+
+      // Augment the native candidate with the parsed fields.
+      var nativeCandidate = new NativeRTCIceCandidate(args);
+      var parsedCandidate = SDPUtils.parseCandidate(args.candidate);
+      var augmentedCandidate = Object.assign(nativeCandidate,
+          parsedCandidate);
+
+      // Add a serializer that does not serialize the extra attributes.
+      augmentedCandidate.toJSON = function() {
+        return {
+          candidate: augmentedCandidate.candidate,
+          sdpMid: augmentedCandidate.sdpMid,
+          sdpMLineIndex: augmentedCandidate.sdpMLineIndex,
+          ufrag: augmentedCandidate.ufrag
+        };
+      };
+      return augmentedCandidate;
+    };
+
+    // Hook up the augmented candidate in onicecandidate and
+    // addEventListener('icecandidate', ...)
+    wrapPeerConnectionEvent('icecandidate', function(e) {
+      if (e.candidate) {
+        Object.defineProperty(e, 'candidate', {
+          value: new RTCIceCandidate(e.candidate), writable: 'false'
+        });
+      }
+      return e;
+    });
+  }
+};
--- a/test/unit/rtcicecandidate.js
+++ b/test/unit/rtcicecandidate.js
@@ -1,0 +1,80 @@
+/*
+ *  Copyright (c) 2017 The WebRTC project authors. All Rights Reserved.
+ *
+ *  Use of this source code is governed by a BSD-style license
+ *  that can be found in the LICENSE file in the root of the source
+ *  tree.
+ */
+/* eslint-env node */
+const chai = require('chai');
+const expect = chai.expect;
+
+describe('RTCIceCandidate', () => {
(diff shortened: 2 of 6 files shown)


fippo:
hrm... @jan-ivar can you take a look?

My biggest gripe here is that this is a lot less useful unless there is a way to hook up both onicecandidate (easy) and addEventListener('icecandidate')

fippo:
hrm. The toJSON stuff doesn't work in Edge :-/

jan-ivar:
@fippo Hooking up RTCPeerConnection.prototype.addEventListener seems doable.

src/js/common.js
```diff
@@ -0,0 +1,50 @@
...
+ */
+ /* eslint-env node */
+'use strict';
+
+var SDPUtils = require('sdp');
+
+module.exports = {
+  shimRTCIceCandidate: function() {
+    if (!window.RTCIceCandidate) {
+      window.RTCIceCandidate = function(args) {
```
jan-ivar: If there's no window.RTCIceCandidate we're probably on a non-webrtc browser and shouldn't shim anything.

I see you moved this from the edge shim. By the comment there, it seems like a flaw specific to Edge. Maybe it should move back?

src/js/common.js
```diff
@@ -0,0 +1,50 @@
...
+        return args;
+      };
+    }
+    if (!window.RTCIceCandidate.prototype.foundation) {
+      var NativeRTCIceCandidate = window.RTCIceCandidate;
+      window.RTCIceCandidate = function(args) {
+        if (args.candidate.indexOf('a=') === 0) {
```
jan-ivar: I think we should defer input validation to the native class as much as possible. i.e. get out of the way of invalid args, e.g.
```js
if (typeof args === 'object' && typeof args.candidate === 'string' &&
    args.candidate.indexOf('a=') === 0)
```

src/js/common.js
```diff
@@ -0,0 +1,50 @@
...
+          args.candidate = args.candidate.substr(2);
+        }
+        var nativeCandidate = new NativeRTCIceCandidate(args);
+        var parsedCandidate = SDPUtils.parseCandidate(args.candidate);
+        Object.assign(nativeCandidate, parsedCandidate);
```
jan-ivar: This seems to overwrite all props in `nativeCandidate`, making it redundant. Did you mean:
```js
var candidate = Object.assign({}, parsedCandidate, new NativeRTCIceCandidate(args));
```
?

src/js/common.js
```diff
@@ -0,0 +1,50 @@
...
+var SDPUtils = require('sdp');
+
+module.exports = {
+  shimRTCIceCandidate: function() {
+    if (!window.RTCIceCandidate) {
+      window.RTCIceCandidate = function(args) {
+        return args;
+      };
+    }
+    if (!window.RTCIceCandidate.prototype.foundation) {
```
jan-ivar: This will give a `TypeError` I think if `.foundation` is ever defined. I.e. `if (!window.RTCIceCandidate.prototype.candidate)` gives me a `TypeError` today.

test/unit/rtcicecandidate.js
```diff
@@ -0,0 +1,74 @@
...
+        candidate: candidateString,
+        sdpMid: 'audio',
+        sdpMLineIndex: 0
+      });
+      expect(candidate.candidate).to.equal(candidateString);
+      expect(candidate.sdpMid).to.equal('audio');
+      expect(candidate.sdpMLineIndex).to.equal(0);
+    });
+
+    it('drops the a= part of the candidate if present', () => {
```
jan-ivar, line 40: Is this "drop the `a=` part" feature documented anywhere?

jan-ivar (changes requested):
Cool stuff!

src/js/common.js
fippo: I considered so but that would mean I have to call this after the Edge shim. Doable...

src/js/common.js
fippo: right, I need a 'in' here!

test/unit/rtcicecandidate.js
fippo: both chrome and firefox used to emit ice candidates with a=candidate in the ice candidate event. This has been removed a while back.

Both chrome and firefox do this which is wrong imo:
```
var pc = new RTCIceCandidate({candidate: 'a=candidate:foo'})
pc.candidate; // returns a=candidate:foo
```

src/js/common.js
```diff
@@ -12,38 +12,36 @@ var SDPUtils = require('sdp');
...
+    // https://w3c.github.io/webrtc-pc/#rtcicecandidate-interface
+    if (window.RTCIceCandidate && !('foundation' in
+        window.RTCIceCandidate.prototype)) {
       var NativeRTCIceCandidate = window.RTCIceCandidate;
       window.RTCIceCandidate = function(args) {
-        if (args.candidate.indexOf('a=') === 0) {
+        // remove the a= which is still accepted by implementations
+        // for legacy reasons.
+        if (typeof args === 'object' && args.candidate &&
+            args.candidate.indexOf('a=') === 0) {
```
jan-ivar: As long as you're cool with `new RTCIceCandidate(1)` failing with `TypeError: args.candidate.indexOf is not a function`  in adapter js instead of the built in `TypeError: Argument 1 of RTCIceCandidate.constructor can't be converted to a dictionary.`

src/js/common.js
```diff
@@ -12,38 +12,36 @@ var SDPUtils = require('sdp');
...
+          args = JSON.parse(JSON.stringify(args));
           args.candidate = args.candidate.substr(2);
         }
         var nativeCandidate = new NativeRTCIceCandidate(args);
+
+        // augment the native candidate with the parsed fields.
         var parsedCandidate = SDPUtils.parseCandidate(args.candidate);
-        Object.assign(nativeCandidate, parsedCandidate);
+        var augmentedCandidate = Object.assign(nativeCandidate,
+            parsedCandidate);
```
jan-ivar: Works, though `augmentedCandidate` is still just `nativeCandidate` by a different name.

jan-ivar (approved):
lgtm.

fippo:
hooked up things in the addEventListener and onicecandidate which makes them useful. Do I need to care about the removeEventListener stuff?

Now I just need to write browser tests to prove this works...

fippo:
gee. This actually worked out better than I expected!

src/js/common.js
```diff
@@ -0,0 +1,97 @@
...
+ /* eslint-env node */
+'use strict';
+
+var SDPUtils = require('sdp');
+
+module.exports = {
+  shimRTCIceCandidate: function() {
+    // foundation is arbitrarily chosen as an indicator for full support for
+    // https://w3c.github.io/webrtc-pc/#rtcicecandidate-interface
+    if (window.RTCIceCandіdidate && 'foundation' in
```
henbos: Typo!

src/js/common.js
```diff
@@ -0,0 +1,97 @@
...
+        window.RTCIceCandidate.prototype) {
+      return;
+    }
+
+    var NativeRTCIceCandidate = window.RTCIceCandidate;
+    window.RTCIceCandidate = function(args) {
+      // remove the a= which is still accepted by implementations
+      // for legacy reasons.
```
henbos: Problem here is not that it accepts "a=" but that it returns ice candidates with "a=", this is a bug regardless of if we want to accept "a=" or not in the SDP.

Can you change the comment to "Remove the a= which shouldn't be part of the candidate string."

src/js/common.js
```diff
@@ -0,0 +1,97 @@
...
+ */
+ /* eslint-env node */
+'use strict';
+
+var SDPUtils = require('sdp');
+
+module.exports = {
+  shimRTCIceCandidate: function() {
+    // foundation is arbitrarily chosen as an indicator for full support for
+    // https://w3c.github.io/webrtc-pc/#rtcicecandidate-interface
```
henbos, line 64: nit: Capital first letter in all comments.

src/js/common.js
```diff
@@ -0,0 +1,97 @@
...
+          Object.defineProperty(e, 'candidate', {
+            value: new RTCIceCandidate(e.candidate), writable: 'false'
+          });
+        }
+        cb(e);
+      };
+      this._eventMap = this._eventMap || {};
+      this._eventMap[cb] = wrappedCallback;
+      return nativeAddEventListener.apply(this, [eventName, wrappedCallback]);
+    };
```
henbos: Can you move add/removeEventListener to somewhere more generic, outside of shimRTCIceCandidate? I'm imagining if we need to shim non-RTCIceCandidate related things in the future and there being conflicts between the different overrides of addEventListener. (Although these should be chainable so maybe not too bad...)

henbos (changes requested):

src/js/common.js
fippo: in this case i'd argue that "foundation" is a technical term. All other comments fixed though.

fippo:
@henbos done -- remains to be seen whether we will need to wrap more events but this should work.

fippo:
It looks like the change of component from string to numeric was something the Edge shim was not ready for yet so i've reverted the changes in the sdp package for now. Will reapply in the future, the fix there is trivial (done in #533)

test/unit/rtcicecandidate.js
```diff
@@ -0,0 +1,78 @@
...
+  });
+
+  it('does not serialize the extra attributes', () => {
+    const candidate = new RTCIceCandidate({
+      candidate: candidateString,
+      sdpMid: 'audio',
+      sdpMLineIndex: 0
+    });
+    const serialized = JSON.stringify(candidate);
+    expect(Object.keys(JSON.parse(serialized)).length).to.equal(4);
```
henbos: What's the fourth member? Could you add a comment?

henbos (approved):
lgtm

test/unit/rtcicecandidate.js
fippo: it should have four members. Will add a commment and then squash!

fippo: ufrag. Surprised? ;-)

fippo:
lint failure is because of unrelated merge and new rules. Will fix elsewhere

fippo:
oh. From "cool but unexpected side-effects": this will make Chrome signal ufrag as part of the top-level object. I think it would be safer to do a major version bump.

KaptenJansson:
LGTM

PR #531, opened 2017-05-12
chrome: complete spec GUM error mapping

fippo:
@guidou can you please take a look?

chrome bug: https://bugs.chromium.org/p/chromium/issues/detail?id=721412

Note that this is a breaking change so requires a major version bump.
I'd like to test this in adapter first and see if we missed anything before we do the same change in Chrome itself.

### DIFF
--- a/src/js/chrome/getusermedia.js
+++ b/src/js/chrome/getusermedia.js
@@ -120,9 +120,13 @@ module.exports = function() {
   var shimError_ = function(e) {
     return {
       name: {
-        ConstraintNotSatisfiedError: 'OverconstrainedError',
         PermissionDeniedError: 'NotAllowedError',
-        TrackStartError: 'NotReadableError'
+        InvalidStateError: 'NotReadableError',
+        DevicesNotFoundError: 'NotFoundError',
+        ConstraintNotSatisfiedError: 'OverconstrainedError',
+        TrackStartError: 'NotReadableError',
+        MediaDeviceFailedDueToShutdown: 'NotReadableError',
+        MediaDeviceKillSwitchOn: 'NotReadableError'
       }[e.name] || e.name,
       message: e.message,
       constraint: e.constraintName,


src/js/chrome/getusermedia.js
```diff
@@ -120,9 +120,13 @@ module.exports = function() {
   var shimError_ = function(e) {
     return {
       name: {
-        ConstraintNotSatisfiedError: 'OverconstrainedError',
         PermissionDeniedError: 'NotAllowedError',
-        TrackStartError: 'NotReadableError'
+        InvalidStateError: 'NotReadableError',
```
guidou, line 124: InvalidStateError depending on the case could be mapped to NotAllowedError, NotReadableError or AbortError. If we have to choose one, I guess NotReadableError is OK.

src/js/chrome/getusermedia.js
```diff
@@ -120,9 +120,13 @@ module.exports = function() {
...
+        DevicesNotFoundError: 'NotFoundError',
+        ConstraintNotSatisfiedError: 'OverconstrainedError',
+        TrackStartError: 'NotReadableError',
```
guidou, line 127: This one can map to NotReadableError or AbortError. According to spec, it should be NotReadableError if "a hardware error such as an OS/program/webpage lock prevents access", and AbortError for all other cases.
I'm slightly leaning towards AbortError here, although the part of Chrome that throws this cannot does not have enough context to know if the cause matches (or not) NotReadableError's description.

guidou:
The mapping is OK, although there are a couple of cases that are not obvious. 
@fippo  LGTM assuming you choose a mapping in the less clear cases based on your practical experience in the wild.

src/js/chrome/getusermedia.js
fippo: sounds good to me but lets double-check with @jan-ivar ;-)

jan-ivar: @guidou Is this the error one gets in Windows when another app is hogging the camera? If so, then it would be unfortunate if the two browsers reported a different error for that.

fippo: @jan-ivar yes, Chrome 59 will (finally!) report an error for that instead of just returning with a track that fails later. Bug @ https://bugs.chromium.org/p/chromium/issues/detail?id=651910

(I wanted to rename TrackStartError there but... it turned out to be this rabbit hole :-))

jan-ivar: @fippo Well that sounds like "OS/program/webpage lock prevents access" to me, so I'd pick `NotReadableError` for that, at least for Firefox.

KaptenJansson:
LGTM

KaptenJansson:
Anything left to discuss here?

fippo:
we also need something for PermissionDismissedError.
spec-question here: https://github.com/w3c/mediacapture-main/issues/451

fippo:
so looks like we're not going to map PermissionDismissedError and go for NotReadableError? @jan-ivar @guidou please use emoji reactions on this comment ;-)

jan-ivar:
@fippo Strictly speaking we should map `PermissionDismissedError` to `NotAllowedError`, since the request was denied, and the latter is an allowable error according to the spec, while the former is not.

However, if this'll cause pain, I'm ok holding off till `navigator.permissions.query({name: "camera"})` is implemented.

👍 on `NotReadableError`

fippo:
we should probably also come up with a way to opt-in on this.

jan-ivar:
@fippo Major version bump?

PR #533, opened 2017-05-15
edge: deal with numeric component id

fippo:
deals with either a string or numeric component id from
parseIceCandidate.


KaptenJansson:
LGTM

PR #542, opened 2017-05-16
Integrate Babel in toolchain

pioug:
**Description**

- Add Babel plugin to Browserify pipeline (with `babel-preset-env` and its configuration)
- Add Babel plugin to Mocha runner

**Purpose**

Integrate Babel will allow new JS syntax (including `import`/`export`) in the repo and can avoid this kind of modification https://github.com/webrtc/adapter/pull/537. It is a better approach than my previous PR https://github.com/webrtc/adapter/pull/515 as it keeps the same tools and gives time to transition (or not).

Last one after I stop pushing 😅

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,9 @@
+{
+  "presets": [
+    ["env", {
+      "targets": {
+        "browsers": ["last 1 version"]
+      }
+    }]
+  ]
+}
--- a/.eslintrc
+++ b/.eslintrc
@@ -57,5 +57,8 @@
     "process": true,
     "Promise": true,
     "Map": true
+  },
+  "parserOptions": {
+      "sourceType": "module"
   }
 }
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -46,6 +46,9 @@ module.exports = function(grunt) {
             './src/js/edge/edge_shim.js'
           ]
         }
+      },
+      options: {
+        transform: ['babelify']
       }
     },
     githooks: {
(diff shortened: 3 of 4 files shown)


src/js/chrome/getusermedia.js
```diff
@@ -11,7 +11,7 @@ var logging = require('../utils.js').log;
 var browserDetails = require('../utils.js').browserDetails;
 
 // Expose public methods.
-module.exports = function() {
```
fippo: we could go without that (for now at least), right?

.babelrc
```diff
@@ -0,0 +1,9 @@
+{
+  "presets": [
+    ["env", {
+      "targets": {
+        "browsers": ["last 1 version"]
```
fippo, line 5: adapter currently works in very old browsers, probably even Chrome 3x. This would break (for the generated code; which is used in the webrtc samples and bower)

It would be nice if we could stop support some of those super old versions (and kill the associated code) but some of the value of adapter is in making things work for old browser versions so you don't have to worry about this.

fippo (commented):
Much easier to digest, thanks!

src/js/chrome/getusermedia.js
pioug: Yes sure, it was for testing. I will remove it.

.babelrc
pioug: Yes this field is up to you. 

There is more explicit way to _query_ browsers https://github.com/ai/browserslist#queries. Maybe something like `Chrome > 30` is more suitable. 

You can test the query here http://browserl.ist/?q=last+1+version.

KaptenJansson:
I think I prefer #574 over this. What are your thoughts?

.babelrc
zoonman: The whole idea of adapter to cover the gap between browser versions and vendors.
As soon as you start dropping support of old/rare browsers you are loosing the existence point of adapter.

KaptenJansson:
Closing for now since there is not enough time to do this big change and all the follow up. Can be reopened in the future if there is time and commitment to do so.

PR #528, opened 2017-05-08
edge: implement addTrack

fippo:
implements addTrack for Edge. Also makes addStream use addTrack.

Pretty big refactor but I think this is better.

Since Edge doesn't allow creating an RTPSender with a null transport the transports need to be created early.

PR #543, opened 2017-05-16
edge: fix bundle parsing

fippo:
fixes a regression where the second transport was
created even though BUNDLE was used.

Introduced in #528 :-/


fippo:
(note to self: need tests to prevent this in the future)

KaptenJansson:
LGTM

PR #544, opened 2017-05-16
Bump version to 3.4.2

KaptenJansson:
**Description**

**Purpose**


fippo (approved):

fippo:
ah... this would include #527 which i'd rather have in a major version. Revert that before the release?

Easier than reverting: deactivate commonShim.shimRTCIceCandidate()

KaptenJansson:
Like so?

Issue #8, opened 2015-02-04
adapter.js should not pollute the global namespace


juberti:
From https://github.com/webrtc/samples/issues/297

ibc commented on Dec 13, 2014
No one in the world should define a global function called trace() (and so many other window globals defined in adapter.js).

adapter.js seems to ignore any new design pattern and JS module system (AMD, CommonJS, etc).

Would you accept a PR removing all the global stuff and providing a library suitable to be loaded by any existing JS module loader without polluting the window global namespace?
Sam Dutton
 Owner
samdutton commented on Dec 13, 2014
You have a point :).

adapter.js worked OK in the context of the demos it was initially built for but, of course, will cause namespace pollution in larger projects.

At the very least it would be great to refactor it from an object-oriented approach with no globals.

@juberti @KaptenJansson @phoglund WDYT?
Iñaki Baz Castillo

ibc commented on Dec 13, 2014
I mean something that can be loaded with npm install webrtc-adapter (and borwserified with browserify), and Bower, and so on.
Iñaki Baz Castillo

ibc commented on Dec 13, 2014
Let me know if you are interested and I will provide a PR.
Justin Uberti
 Owner
juberti commented on Dec 14, 2014
yeah, I think it's time for adapter.js to grow up. Would you mind changing the title of this issue to be more descriptive?

And please provide an initial PR for the adapter - we'll then have to do a subsequent PR for changing all of our sample code.
Iñaki Baz Castilloibc changed the title from Please build modern JS to Make adapter.js a modern JS library (module systems aware) on Dec 14, 2014
Iñaki Baz Castillo

ibc commented on Dec 14, 2014
I will start in my own repo by creating a JS "project" (not a single adapter.js file), which will include Grunt/gulp tasks for linting, browserifying, etc.

Let me some time please, I will notify it here when done.
Sam Dutton
 Owner
 samdutton commented on Dec 14, 2014
Thanks for doing this!

adapter.js will be happy to get some attention :).
…
Justin Uberti
 Owner
juberti commented on Dec 14, 2014
Before spending too much time, it would be good to get a high-level overview of what you expect it to look like in the end. I am all for avoiding global namespace pollution but I want to make sure we're not biting off too much here.
Christoffer Jansson
 Owner
KaptenJansson commented on Dec 14, 2014
SGTM
Will

willscott commented on Dec 15, 2014
FYI - I've begun maintaining a CommonJS library to do something similar: https://www.npmjs.org/package/webrtc-adapter
[...]

PR #547, opened 2017-05-17
Make polluting global scope optional

aiham:
Hi,

We'd like to use adapter.js in our project but we don't agree with the decision to pollute the global scope. It's been [mentioned in the past](https://github.com/webrtc/adapter/issues/8#issuecomment-124082448) but not fixed. The current non-global approach only avoids exposing the `adapter` object, but still modifies the prototypes of native objects like `RTCPeerConnection`.

Our SDK is included on many different webpages so we do our best to avoid polluting the global scope because our clients are likely to be making decisions based on its state. Including adapter.js in its current form could invalidate their assumptions. It would be great if we could find a way to resolve this. I have a possible solution:

I propose that adapter and its shims be converted into factories that have the `window` object injected into them rather than always modifying the global scope. The existing adapter release files will still behave as they currently do by invoking the main factory with the global `window` object. An additional release file exposing the underlying factory will enable us to provide our own custom window-like object for the shims to modify.

The diff view of this PR is quite large but that's mostly indentation so please view it by [ignoring whitespace changes](https://github.com/webrtc/adapter/pull/547/files?w=1).

Taking a look at [just the changes to the safari shim](https://github.com/webrtc/adapter/pull/547/commits/5cf3586a57be808e585d6aea9dc4b10591532076?w=1) (with whitespace ignored) should demonstrate the general pattern.

I tried to stick to the following when implementing this solution:

- Don't break backwards compatibility with the existing adapter.js API
- Don't change the general structure of the project (excluding whitespace changes)
- Prefix global objects with `window.` where possible
  - If a global object is used often (eg. `RTCPeerConnection`) then declare it as a local variable and ensure its always pointing to the `window` property of the same name.

This PR unfortunately touches almost every line in the src folder so I made an alternative PR that makes as little change as possible in terms of lines, unfortunately the negative is that it's not backwards compatible with the existing adapter.js API: [#548](https://github.com/webrtc/adapter/pull/548) It works in a similar way but instead of converting the shims into factories, an additional `window` argument is added to each shim function.

The most ideal solution is for the project to be restructured to accomodate these requirements (and others, such as properly modularising the shims) but I assume that's not likely to happen, mostly because it would break backwards compatibility.

The unit tests and integration tests all pass and it should be pretty seamless for both users and developers of adapter.js. Let me know if you have any questions, concerns or alternative solutions.

Note: There are quite a few conflicts because of all the PRs that have been merged in the past week but it should still illustrate the solution. If you agree with this approach then I can quickly work on resolving the conflicts.

Aiham


aiham:
I signed it!

fippo:
I have a slight preference for #548 but could live with either style

aiham:
No problem, I'll close this and focus on #548

fippo:
wait, lets ask for @KaptenJansson for an opinion too ;-)

KaptenJansson:
I also prefer #548 since I think adapter has most value as a complete shim hence there is not really a need to make it more modular now. And I assume people do not want to inject different window objects in each shim?

I also wonder if it then makes sense to remove the extra build artifacts in https://github.com/webrtc/adapter/blob/master/Gruntfile.js#L18-L49 ? I think the MS edge should be standard now, we can do this in a follow up PR.

aiham:
@KaptenJansson Sounds good. I'll stick with #548 then.

PR #548, opened 2017-05-17
Make polluting global scope optional (alternative)

aiham:
Alternative solution to #547


aiham:
I signed it!

fippo (approved):
👍 (we should do a major version bump after this)

aiham:
Thanks. I'll start working on resolving the conflicts.

Gruntfile.js
```diff
@@ -46,6 +46,12 @@ module.exports = function(grunt) {
             './src/js/edge/edge_shim.js'
           ]
         }
+      },
+      // Use this if want to invoke the factory to construct adapter yourself
+      // with a custom window object.
+      adapterFactory: {
+        src: ['./src/js/adapter_factory.js'],
+        dest: './out/adapter_factory.js'
```
fippo: is it necessary to provide this? If you want the factory you'd need to require/import the module, i wouldn't want to make our life harder with this ;-)

fippo (commented):
bonus points for making the mock unit tests easier! :-)

Gruntfile.js
aiham: Good point. We can access it with `const adapterFactory = require('webrtc-adapter/src/js/adapter_factory.js');` we just have to be careful in case the file structure changes in the future.

I'll remove this.

aiham:
@fippo I've updated the branch. There was something I wanted to change that might be a little controversial so I split it into its own commit: 509a50fc3d64842b7d3f0b3628b9e626f5ebbaa1

Removes `utils.browserDetails` and runs `utils.detectBrowser()` for every shim. This is repetitive but I think its a cheap operation and less disruptive than converting `utils` into a factory.

Without this commit the browser detection always runs against the global window. With this commit almost all unit tests are completely isolated from each other.

WDYT?

aiham:
@fippo @KaptenJansson I'm going to wait on getting the final green light on these 2 commits before resolving the conflicts again otherwise I'll probably end up repeating myself a bunch of times. Just let me know once you're happy with it, I'll resolve the conflicts and we can get it merged in before anything else causes another conflict. Thanks!

fippo:
:+1: -- even if the only reason was to make testing more isolated!

fippo:
the last commit will conflict with #550 -- but i can rebase,

aiham:
@fippo Oh you're converting them to unit tests, that's great. I just spent some time trying to fake the Firefox version in the integration tests but I think going with unit tests is the better approach. It would be great if you could rebase. Should I disable these tests so my PR passes?

fippo:
Might be easier if you revert your last commit and then merge mine. Then its no longer your problem ;-)

KaptenJansson:
This looks great! LGTM once the test passes, will hold of reviewing/merging other PR's until this is in so you can solve all conflicts.

aiham:
@fippo Done!

fippo:
argh... so much for "that plan is easier".

consenting to inclusion obviously.

aiham:
I think the chrome beta test will pass if restarted. It timed out.

fippo:
\o/

@KaptenJansson so lets merge this and i'll try to re-land #527 with a native implementation for edge

KaptenJansson:
LGTM! Thanks for doing this!

fippo:
I wonder if we should remove https://github.com/fippo/eslint-config-webrtc/blob/master/index.js now. That would prevent us from accidentally using those objects, right?

fippo:
pity, it looks like eslint knows most of these things nowadays.

aiham:
@fippo @KaptenJansson :tada: Thanks for helping to get this through!

PR #550, opened 2017-05-17
test: use unit tests for getUserMedia constraints converter

fippo:
uses sinon.stub to make assertions about how the constraints
converter works in different browsers and versions.

@jan-ivar PTAL -- this was one of the more scary tests, I find this a bit easier to reason about.
IIRC this test was one of the reasons we are still using webkitGetUserMedia.


test/unit/getusermedia-constraints.js
```diff
@@ -0,0 +1,175 @@
...
+        }
+      };
+      window.navigator.getUserMedia(legacy);
+      expect(window.navigator.mozGetUserMedia).to.have.been.calledWith(legacy);
+    });
+  });
+
+  describe('in Firefox 38+', () => {
+    beforeEach(() => {
+      utils.browserDetails.version = 53;
```
jan-ivar: The old version had 38 here.

jan-ivar (approved):
Nice! sinon seems a good fit for this sort of thing.

test/unit/getusermedia-constraints.js
fippo: changed

fippo:
sinon is a lot less hassle than doing this yourself for sure ;-)

Issue #554, opened 2017-05-17
peerconnection/pc1/ Demo Does Not Work on Edge


dcreey:
**Browsers and versions affected**
Edge

**Description**
peerconnection/pc1 demo does not work on Edge. Error is thrown: "SCRIPT16387: Could not complete the operation due to error c0044003.". 

Candidates are manually added to the paired peer connection via addIceCandidate. Before sending the 'end-of-candidates' (null) the remote candidates list contains duplicate candidates (foundation numbers 1,2,3 appear twice for a total of 6 candidates).

Perhaps either duplicate candidates (same foundation number) should be filtered out, or an error should be thrown if attempting to add a candidate with the same foundation number to the peer connection.

**Steps to reproduce**
Click Start, then Call in the peerconnection/pc1 demo on Edge.

**Expected results**
Peer connections successfully establish a connection and video displays in the remote video element.

**Actual results**
Error is thrown in the console and the remote video element remains black.

fippo:
looks like a regression between 3.4.1 and 3.4.2.
Can you downgrade to 3.3.x? (not for the samples obviously)

fippo:
alternatively deactivate https://github.com/webrtc/adapter/blob/master/src/js/adapter_core.js#L104
-- this shouldn't have ended in this release and it seems to cause onicecandidate to fire twice

dcreey:
Version 3.4.0 works.

PR #555, opened 2017-05-17
Revert "Revert "Revert "Merge pull request #527 from fippo/rtcicecandidate"""

fippo:
Reverts webrtc/adapter#549

This caused #554 and we need a release without it (and 3.4.2 does include it unfortunately)

This is rather odd, In particular, this code calls
```
pc.onicecandidate = (e) => {console.log(e.candidate.candidate);};
pc.createOffer({offerToReceiveAudio: true})
.then(offer => pc.setLocalDescription(offer));
```
onicecandidate twice for every candidate. Also the wrapper does not even work in Edge since it does not define addEventListener on the prototype.

I'll probably just implement it natively in Edge/ORTC.


fippo:
@KaptenJansson @MirkoBonadei ping

KaptenJansson:
LGTM. New patch release?

fippo:
yes please

Issue #552, opened 2017-05-17
Chrome implements getSenders


fippo:
Chrome 60+ implements getSenders behind the experimental web platform features flag.
Currently the only attribute on the sender is its track.

This breaks shimGetSendersWithDtmf in the chrome shim as well as addTrack.
@jan-ivar @henbos can we find a way make this work?

henbos:
If getSenders exists can we update adapter.js to shim adding dtmf to each sender?

fippo:
only 50% resolved, addTrack still needs to be fixed.

KaptenJansson:
Auto close :)

PR #558, opened 2017-05-19
Fix Chrome dtmf shim to work with native sender.

jan-ivar:
**Description**
Should fix https://github.com/webrtc/adapter/issues/552.

**Purpose**
Keep spec dtmf working.

### DIFF
--- a/src/js/chrome/chrome_shim.js
+++ b/src/js/chrome/chrome_shim.js
@@ -173,6 +173,32 @@ var chromeShim = {
           }
         });
       };
+    } else if (typeof window === 'object' && window.RTCPeerConnection &&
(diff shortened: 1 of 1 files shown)


fippo:
LGTM. But we also need to make addTrack work again, i'll take a look

aiham (changes requested):
Since #548 has been merged, could we prefix the globals with `window.` in this PR?

KaptenJansson:
LGTM after the window prefixes has been added.

jan-ivar:
Done.

fippo:
the dtmf test seems a bit flaky?

KaptenJansson:
Kicking tests to see if they want to play along today.

Issue #557, opened 2017-05-18
Chrome DOMException: Track already exists for fake devices


adamweeks:
When running automation tests, we open two chrome instances and call each other. We use the following options to use fake devices in Chrome:
* `--use-fake-device-for-media-stream`,
* `--use-fake-ui-for-media-stream`

Since version 3.4.1, we are getting `DOMException: Track already exists` error on our second instance when attempting to answer a call. 

This error did not happen in prior versions and test calls were able to be established properly.

Chrome version 58 on mac.

fippo:
hrm. Do those tests use addTrack when available?

This is probably the exception [thrown here](https://github.com/webrtc/adapter/blob/master/src/js/chrome/chrome_shim.js#L108) so i wonder why it is thrown. What is the sequence of events your test calls?

adamweeks:
Our tests places a call from browser A to browser B. The error is thrown when browser B attempts to answer the call. This is where we are adding tracks: https://github.com/ciscospark/spark-js-sdk/blob/master/packages/node_modules/@ciscospark/plugin-phone/src/webrtc.js#L39

fippo:
Hrm. That code looks like you're trying to solve exactly the problem the addTrack shim is supposed to solve.

I've come close to weird behaviour in https://jsfiddle.net/9awhwp03/
 -- if you use pc.getLocalStreams() instead of the saved stream things go wrong. 
But not the same error you get.

I think the root cause is using pc.getLocalStreams() which returns a different stream than what you expect. Using getLocalStreams() (a stream-based webrtc api) together with addTrack (track-based) is somewhat underspecified.

jan-ivar:
@fippo Looks like they're keeping track of all local tracks in the same local stream. That should work in Firefox. They probably do it for symmetry with addStream in Chrome, which makes sense to me.

This would trip up your Chrome shim which does the same thing. Since the track is already where it needs to be, might it suffice to suppress the error?

fippo:
Still trying to replicate. Even copying the code into a fiddle https://jsfiddle.net/ofssnscp/ I get a different error (the track is now part of multiple streams) so i'm missing a step.

fippo:
I made the code work with two modifications:
1) get rid of pc.getLocalStreams
2) conditionally remove the "silly hack" if pc.addTrack exists

I think the first can be avoided by modifying pc.getLocalStreams together with shimming pc.addTrack.

jan-ivar:
[...]

PR #559, opened 2017-05-19
chrome: make addTrack + getLocalStreams work

fippo:
fix for #557 (hopefully)

### DIFF
--- a/src/js/chrome/chrome_shim.js
+++ b/src/js/chrome/chrome_shim.js
@@ -74,96 +74,54 @@ var chromeShim = {
     if (typeof window === 'object' && window.RTCPeerConnection &&
         !('getSenders' in window.RTCPeerConnection.prototype) &&
         'createDTMFSender' in window.RTCPeerConnection.prototype) {
-      window.RTCPeerConnection.prototype.getSenders = function() {
-        return this._senders || [];
+      var shimSenderWithDtmf = function(pc, track) {
+        return {
+          track: track,
+          get dtmf() {
+            if (this._dtmf === undefined) {
+              if (track.kind === 'audio') {
+                this._dtmf = pc.createDTMFSender(track);
+              } else {
+                this._dtmf = null;
+              }
+            }
+            return this._dtmf;
+          }
+        };
       };
-      var origAddStream = window.RTCPeerConnection.prototype.addStream;
-      var origRemoveStream = window.RTCPeerConnection.prototype.removeStream;
 
-      if (!window.RTCPeerConnection.prototype.addTrack) {
+      // shim addTrack when getSenders is not available.
+      if (!window.RTCPeerConnection.prototype.getSenders) {
+        window.RTCPeerConnection.prototype.getSenders = function() {
+          return this._senders || [];
+        };
+        var origAddTrack = window.RTCPeerConnection.prototype.addTrack;
         window.RTCPeerConnection.prototype.addTrack = function(track, stream) {
           var pc = this;
-          if (pc.signalingState === 'closed') {
-            throw new DOMException(
-              'The RTCPeerConnection\'s signalingState is \'closed\'.',
-              'InvalidStateError');
+          var sender = origAddTrack.apply(pc, arguments);
+          if (!sender) {
+            sender = shimSenderWithDtmf(pc, track);
+            pc._senders.push(sender);
           }
-          var streams = [].slice.call(arguments, 1);
-          if (streams.length !== 1 ||
-              !streams[0].getTracks().find(function(t) {
-                return t === track;
-              })) {
-            // this is not fully correct but all we can manage without
-            // [[associated MediaStreams]] internal slot.
-            throw new DOMException(
-              'The adapter.js addTrack polyfill only supports a single ' +
-              ' stream which is associated with the specified track.',
-              'NotSupportedError');
-          }
-
-          pc._senders = pc._senders || [];
-          var alreadyExists = pc._senders.find(function(t) {
-            return t.track === track;
-          });
-          if (alreadyExists) {
-            throw new DOMException('Track already exists.',
-                'InvalidAccessError');
-          }
-
-          pc._streams = pc._streams || {};
-          var oldStream = pc._streams[stream.id];
-          if (oldStream) {
-            oldStream.addTrack(track);
-            pc.removeStream(oldStream);
-            pc.addStream(oldStream);
-          } else {
-            var newStream = new window.MediaStream([track]);
-            pc._streams[stream.id] = newStream;
-            pc.addStream(newStream);
-          }
-
-          var sender = {
-            track: track,
-            get dtmf() {
-              if (this._dtmf === undefined) {
-                if (track.kind === 'audio') {
-                  this._dtmf = pc.createDTMFSender(track);
-                } else {
-                  this._dtmf = null;
-                }
-              }
-              return this._dtmf;
-            }
-          };
-          pc._senders.push(sender);
           return sender;
         };
       }
+      var origAddStream = window.RTCPeerConnection.prototype.addStream;
       window.RTCPeerConnection.prototype.addStream = function(stream) {
         var pc = this;
         pc._senders = pc._senders || [];
         origAddStream.apply(pc, [stream]);
         stream.getTracks().forEach(function(track) {
-          pc._senders.push({
-            track: track,
-            get dtmf() {
-              if (this._dtmf === undefined) {
-                if (track.kind === 'audio') {
-                  this._dtmf = pc.createDTMFSender(track);
-                } else {
-                  this._dtmf = null;
-                }
-              }
-              return this._dtmf;
-            }
-          });
+          pc._senders.push(shimSenderWithDtmf(pc, track));
         });
       };
 
+      var origRemoveStream = window.RTCPeerConnection.prototype.removeStream;
       window.RTCPeerConnection.prototype.removeStream = function(stream) {
         var pc = this;
         pc._senders = pc._senders || [];
-        origRemoveStream.apply(pc, [stream]);
+        origRemoveStream.apply(pc, [(pc._streams[stream.id] || stream)]);
+
         stream.getTracks().forEach(function(track) {
           var sender = pc._senders.find(function(s) {
             return s.track === track;
@@ -245,6 +203,100 @@ var chromeShim = {
         });
       }
     }
+  },
+
+  shimAddTrack: function(window) {
+    // shim addTrack (when getSenders is available)
+    if (window.RTCPeerConnection.prototype.addTrack) {
+      return;
+    }
+
+    // also shim pc.getLocalStreams when addTrack is shimmed
+    // to return the original streams.
+    var origGetLocalStreams = window.RTCPeerConnection.prototype
+        .getLocalStreams;
+    window.RTCPeerConnection.prototype.getLocalStreams = function() {
+      var self = this;
+      var nativeStreams = origGetLocalStreams.apply(this);
+      self._reverseStreams = self._reverseStreams || {};
+      return nativeStreams.map(function(stream) {
+        return self._reverseStreams[stream.id];
+      });
+    };
+
+    var origAddStream = window.RTCPeerConnection.prototype.addStream;
+    window.RTCPeerConnection.prototype.addStream = function(stream) {
+      var pc = this;
+      pc._streams = pc._streams || {};
+      pc._reverseStreams = pc._reverseStreams || {};
+
+      // Add identity mapping for consistency with addTrack.
+      // Unless this is being used with a stream from addTrack.
+      if (!pc._reverseStreams[stream.id]) {
+        pc._streams[stream.id] = stream;
+        pc._reverseStreams[stream.id] = stream;
+      }
+      origAddStream.apply(pc, [stream]);
+    };
+
+    var origRemoveStream = window.RTCPeerConnection.prototype.removeStream;
+    window.RTCPeerConnection.prototype.removeStream = function(stream) {
+      var pc = this;
+      pc._streams = pc._streams || {};
+      pc._reverseStreams = pc._reverseStreams || {};
+
+      origRemoveStream.apply(pc, [(pc._streams[stream.id] || stream)]);
+      delete pc._reverseStreams[(pc._streams[stream.id] ?
+          pc._streams[stream.id].id : stream.id)];
+      delete pc._streams[stream.id];
+    };
+
+    window.RTCPeerConnection.prototype.addTrack = function(track, stream) {
+      var pc = this;
+      if (pc.signalingState === 'closed') {
+        throw new DOMException(
+          'The RTCPeerConnection\'s signalingState is \'closed\'.',
+          'InvalidStateError');
+      }
+      var streams = [].slice.call(arguments, 1);
+      if (streams.length !== 1 ||
+          !streams[0].getTracks().find(function(t) {
+            return t === track;
+          })) {
+        // this is not fully correct but all we can manage without
+        // [[associated MediaStreams]] internal slot.
+        throw new DOMException(
+          'The adapter.js addTrack polyfill only supports a single ' +
+          ' stream which is associated with the specified track.',
+          'NotSupportedError');
+      }
+
+      var alreadyExists = pc.getSenders().find(function(s) {
+        return s.track === track;
+      });
+      if (alreadyExists) {
+        throw new DOMException('Track already exists.',
+            'InvalidAccessError');
+      }
+
+      pc._streams = pc._streams || {};
+      pc._reverseStreams = pc._reverseStreams || {};
+      var oldStream = pc._streams[stream.id];
+      if (oldStream) {
+        // this is using odd Chrome behaviour, use with caution:
+        // https://bugs.chromium.org/p/webrtc/issues/detail?id=7815
+        oldStream.addTrack(track);
+        pc.dispatchEvent(new Event('negotiationneeded'));
+      } else {
+        var newStream = new window.MediaStream([track]);
+        pc._streams[stream.id] = newStream;
+        pc._reverseStreams[newStream.id] = stream;
+        pc.addStream(newStream);
+      }
+      return pc.getSenders().find(function(s) {
+        return s.track === track;
+      });
+    };
   },
 
   shimPeerConnection: function(window) {
@@ -446,6 +498,7 @@ var chromeShim = {
 module.exports = {
   shimMediaStream: chromeShim.shimMediaStream,
   shimOnTrack: chromeShim.shimOnTrack,
+  shimAddTrack: chromeShim.shimAddTrack,
   shimGetSendersWithDtmf: chromeShim.shimGetSendersWithDtmf,
   shimSourceObject: chromeShim.shimSourceObject,
   shimPeerConnection: chromeShim.shimPeerConnection,
--- a/test/test.js
+++ b/test/test.js
@@ -811,6 +811,29 @@ test('Call getUserMedia with impossible constraints',
       });
     });
 
+test('getSenders exists', function(t) {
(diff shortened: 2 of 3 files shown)


henbos:
I wonder what is the right behavior for getLocalStreams() when streams have been added with addTrack(), see [related discussion about remote streams](https://github.com/w3c/webrtc-pc/issues/1182).

Should a stream return every track that has a sender with that stream? If not, it makes little sense to return streams in getLocalStreams() that have not been added with addStream() or we could return streams that appear to be empty even though a sender has associated that stream with a track. If yes, do we do this only for senders that are currently in use or all senders that exist? What happens if you mix pc.addStream, stream.addTrack, pc.addTrack and sender.replaceTrack behavior...

fippo:
well, this is the classical example...
```
navigator.mediaDevices.getUserMedia({audio: true, video: true})
.then((stream) => {
  pc.addTrack(stream.getAudioTracks()[0], stream);
  // what tracks does pc.getLocalStreams()[0] have?
})
```
With this change the answer is consistent with what Firefox returns, a stream with two tracks.

I'm quite worried about mixed use of addStream and addTrack...

henbos:
Interesting.
```
navigator.mediaDevices.getUserMedia({audio: true, video: true})
.then((stream) => {
  let audioTrack = stream.getAudioTracks()[0];
  stream.removeTrack(audioTrack);
  pc.addTrack(audioTrack, stream);
  // 1. Is stream.getAudioTracks() empty?
  // 2. What tracks does pc.getLocalStreams()[0] have?
})
```
Does that work? :)

fippo:
That was throwing (and still does):
```
DOMException: The adapter.js addTrack polyfill only supports a single  stream which is associated with the specified track.
```
Arguably that is a pretty arbitrary restriction which is not covered by the spec.

fippo:
we are digressing but...
```
navigator.mediaDevices.getUserMedia({audio: true, video: true})
.then((stream) => {
  console.log(stream.id);
  let audioTrack = stream.getAudioTracks()[0];
  stream.removeTrack(audioTrack);
  pc.addTrack(audioTrack, stream);
  console.log(stream.getTracks());
})
```
In Firefox stream.getTracks() returns a stream with only a video track. pc.getLocalStreams()[0] also returns a stream with just a video track.

```
navigator.mediaDevices.getUserMedia({audio: true})
.then((stream) => {
  console.log(stream.id);
  pc.addStream(stream);
  return navigator.mediaDevices.getUserMedia({video: true})
})
.then(stream => {
  var t = stream.getTracks()[0];
  pc.addTrack(t, pc.getLocalStreams()[0]);
})
```
works in Firefox but pc.getLocalStreams()[0] only has an audio track. This shim blows up.

KaptenJansson:
@henbos any other thoughts?

fippo:
(needs minor updates to use window. now)

fippo:
updated but kept as separate commits in case we want to cherry-pick the first one for a 3.x patch release.

KaptenJansson:
Ping @henbos

src/js/chrome/chrome_shim.js
```diff
@@ -138,10 +140,28 @@ var chromeShim = {
...
+        };
       }
       RTCPeerConnection.prototype.addStream = function(stream) {
         var pc = this;
         pc._senders = pc._senders || [];
+        pc._streams = pc._streams || {};
+        pc._reverseStreams = pc._reverseStreams || {};
+
+        pc._streams[stream.id] = stream; // Add identity mapping for
+        pc._reverseStreams[stream.id] = stream; // consistency with addTrack.
```
henbos: nit: Don't split end-of-line comments into multiple comments, move to above the two lines instead.

fippo:
hrm. Something is phishy here. Probably because this uses pc.addStream internally instead of the non-shimmed version.

test/test.js
```diff
@@ -2963,6 +2963,87 @@ test('ontrack', function(t) {
...
+      tc.ok('got audio stream');
+      pc.addTrack(stream.getAudioTracks()[0], stream);
+      return navigator.mediaDevices.getUserMedia({video: true});
+    })
+    .then(function(stream) {
+      tc.ok('got video stream');
+      var track = stream.getVideoTracks()[0];
+      var localStream = pc.getLocalStreams()[0];
+      localStream.addTrack(track);
+      pc.addTrack(track, localStream);
```
henbos: Can you verify that pc.getLocalStream still returns a single stream that is localStream after this, and that it has an audio and video track?

henbos (dismissed):

henbos:
Is something not working?

fippo:
pc._streams contains two members after adding two tracks from the same stream. this is rather unexpected...

henbos:
Perhaps you should verify _streams etc. in the test as well.

fippo:
internal apis should only be tested in unit tests. Now there should be unit tests for this, yes... ;-)

fippo:
@KaptenJansson merging the PRs in that area is going to be a mess. Lets plan to do it on Wednesday

fippo:
from "weird things that work as expected" for some value of "expected":
```
var localStream;
var senders1;
var senders2;
navigator.mediaDevices.getUserMedia({audio:true, video: true})
.then(stream => {
  localStream = stream;
  pc.addTrack(stream.getTracks()[0], stream);
  console.log(pc.getLocalStreams()[0].getTracks());
})
```
show an audio and a video track

fippo:
ugh. So the test failure in chrome beta (with native senders) is highly confusing. It looks like once the track is added to the stream that has been added to the peerconnection, a sender is created:
https://jsfiddle.net/bebhg6ya/
which makes the addTrack shim bail out early since it sees the sender already existing. @henbos ^^

I think I can work around it somehow

fippo:
filed a bug about that fiddle because chrome and firefox return different results from createOffer: https://bugs.chromium.org/p/webrtc/issues/detail?id=7815

KaptenJansson:
@henbos could you PTAL again? Thanks!

src/js/chrome/chrome_shim.js
```diff
@@ -102,23 +103,24 @@ var chromeShim = {
...
           if (alreadyExists) {
             throw new DOMException('Track already exists.',
                 'InvalidAccessError');
           }
 
           pc._streams = pc._streams || {};
+          pc._reverseStreams = pc._reverseStreams || {};
           var oldStream = pc._streams[stream.id];
           if (oldStream) {
             oldStream.addTrack(track);
```
fippo: ironically I am using the behaviour that I find so "bad" here. cc @taylor-b

src/js/chrome/chrome_shim.js
```diff
@@ -102,23 +103,26 @@ var chromeShim = {
...
 
           pc._senders = pc._senders || [];
-          var alreadyExists = pc._senders.find(function(t) {
-            return t.track === track;
+          var alreadyExists = pc._senders.find(function(sender) {
+            return sender.track === track;
           });
           if (alreadyExists) {
             throw new DOMException('Track already exists.',
                 'InvalidAccessError');
```
henbos: 'A sender already exists for this track.'

src/js/chrome/chrome_shim.js
```diff
@@ -138,32 +142,58 @@ var chromeShim = {
...
-                if (track.kind === 'audio') {
-                  this._dtmf = pc.createDTMFSender(track);
-                } else {
-                  this._dtmf = null;
+        pc._streams = pc._streams || {};
+        pc._reverseStreams = pc._reverseStreams || {};
+
+        // Add identity mapping for consistency with addTrack.
+        // Unless this is being used with a stream from addTrack.
+        if (!pc._reverseStreams[stream.id]) {
```
henbos: Can this happen? The internally created streams are as far as I can tell completely hidden from the external caller (unless the caller is an internal, another part of the shimming).

src/js/chrome/chrome_shim.js
```diff
@@ -138,32 +142,58 @@ var chromeShim = {
...
+          stream.getTracks().forEach(function(track) {
+            pc._senders.push({
+              track: track,
+              get dtmf() {
+                if (this._dtmf === undefined) {
+                  if (track.kind === 'audio') {
+                    this._dtmf = pc.createDTMFSender(track);
+                  } else {
+                    this._dtmf = null;
+                  }
```
henbos: track/dtmf shimming: Can the repeated code be avoided with a helper function?

src/js/chrome/chrome_shim.js
```diff
@@ -220,6 +254,104 @@ var chromeShim = {
...
+        pc.dispatchEvent(new Event('negotiationneeded'));
+      } else {
+        var newStream = new window.MediaStream([track]);
+        pc._streams[stream.id] = newStream;
+        pc._reverseStreams[newStream.id] = stream;
+        pc.addStream(newStream);
+      }
+      return pc.getSenders().find(function(sender) {
+        return sender.track === track;
+      });
```
henbos: Do we really need to shim addTrack etc twice? It's complicated and lots of repeated code. Can we not have a single shim and do if for what is different: "if we have getSenders, return the sender with that track, otherwise return our own sender object", not repeating the code that is shared?

src/js/chrome/chrome_shim.js
fippo: the shimmed addTrack uses the shimmed addStream. Pigs fly :-)

src/js/chrome/chrome_shim.js
fippo: it is a lot of duplication but the alternative was to have a very complicated do-everything :-/

src/js/chrome/chrome_shim.js
fippo: the spec says "If an RTCRtpSender for track already exists in senders" -- i'd rather keep this close to the spec language

src/js/chrome/chrome_shim.js
fippo: done. I also noticed that with native senders we do not support dtmf. Can you consider not making them available until that is hooked up to reduce the complexity? Might be easy to backfill now though...

src/js/chrome/chrome_shim.js
```diff
@@ -90,6 +74,22 @@ var chromeShim = {
...
         'createDTMFSender' in window.RTCPeerConnection.prototype) {
+      var shimSenderWithDtmf = function(pc, track) {
+        return {
+          track: track,
+          get dtmf() {
+            if (pc._dtmf === undefined) {
+              if (track.kind === 'audio') {
+                pc._dtmf = pc.createDTMFSender(track);
+              } else {
+                pc._dtmf = null;
```
fippo: @jan-ivar what was the purpose of setting pc._dtmf to null? or did I do the wrong thing here by assuming this === pc?

jan-ivar: @fippo [dtmf](https://w3c.github.io/webrtc-pc/#dom-rtcrtpsender-dtmf) says: *"A null value indicates that this RTCRtpSender cannot send DTMF."* - We can't send DTMF signals on a video track can we?

fippo: either way i think the else is not necessary

fippo: ah... and this is sender not pc. Thanks for explaining!

fippo:
I think i'm done. Only thing left before a big squash is making the dtmf test unskip-able -- i need to check the status in Edge and Safari for that.

fippo:
so Safari doesn't seem to do dtmf and  Edge needs a shim for sender.dtmf -- easy but not in this PR.

It might be possible to make this stuff simpler by polyfilling it on RTCRtpSender.prototype globally. But again, not this PR

src/js/chrome/chrome_shim.js
jan-ivar: I guess that works. i.e. I guess `pc.createDTMFSender()` may be called more than once on the same `pc` with different tracks in Chrome? Might want to test that manually at least.

fippo: i already reverted that, it was stupid and your stuff was better :-)

src/js/chrome/chrome_shim.js
```diff
@@ -247,6 +273,126 @@ var chromeShim = {
...
+        pc._reverseStreams[newStream.id] = stream;
+        pc.addStream(newStream);
+      }
+      var sender = pc.getSenders().find(function(s) {
+        return s.track === track;
+      });
+      if (!('dtmf' in window.RTCRtpSender.prototype)) {
+        shimDtmfOnNativeSender(pc, sender);
+      }
+      return sender;
```
henbos: prepareSender(track) for the has getSender case could be lines 344-350 and here do return prepareSender(track);

src/js/chrome/chrome_shim.js
```diff
@@ -247,6 +273,126 @@ var chromeShim = {
...
+        throw new DOMException(
+          'The adapter.js addTrack polyfill only supports a single ' +
+          ' stream which is associated with the specified track.',
+          'NotSupportedError');
+      }
+
+      var alreadyExists = pc.getSenders().find(function(s) {
+        return s.track === track;
+      });
+      if (alreadyExists) {
```
henbos, line 277: findSender(track) for the has getSender case could be lines 322-324 and here if (findSender(track)).

src/js/chrome/chrome_shim.js
```diff
@@ -102,68 +119,73 @@ var chromeShim = {
...
-                } else {
-                  this._dtmf = null;
-                }
-              }
-              return this._dtmf;
-            }
-          };
+          var sender = shimSenderWithDtmf(pc, track);
           pc._senders.push(sender);
           return sender;
```
henbos: prepareSender(track) for the getSender missing case could be lines 145-147 and here return prepareSender(track).

src/js/chrome/chrome_shim.js
```diff
@@ -102,68 +119,73 @@ var chromeShim = {
           }
 
           pc._senders = pc._senders || [];
-          var alreadyExists = pc._senders.find(function(t) {
-            return t.track === track;
+          var alreadyExists = pc._senders.find(function(sender) {
+            return sender.track === track;
           });
           if (alreadyExists) {
```
henbos: findSender(track) for the getSender missing case could be lines 122-124 and here if (findSender(track)).

henbos (changes requested):
I still have a feeling shimAddTrack doesn't have to be defined twice without complicated code, see other comments. Also consider the addStream/removeStream/getLocalStreams cases.

src/js/chrome/chrome_shim.js
henbos: That is, if we define shimAddTrack once and use helper functions findSender and prepareSender that are defined differently whether or not we have getSenders.

fippo:
I can see where you want to go... basically shimming getSenders and then making the decisions in addStream depend on whether getSenders is available etc. This could work and eliminate the duplicated code. It will take another two weeks to get right at the current pace and the basic functionality that this PR is trying to fix has been reported as broken over a month ago.

src/js/chrome/chrome_shim.js
fippo: i'd rather do that with an arrow function once we can use them in ES6

src/js/chrome/chrome_shim.js
```diff
@@ -247,6 +205,99 @@ var chromeShim = {
...
+          })) {
+        // this is not fully correct but all we can manage without
+        // [[associated MediaStreams]] internal slot.
+        throw new DOMException(
+          'The adapter.js addTrack polyfill only supports a single ' +
+          ' stream which is associated with the specified track.',
+          'NotSupportedError');
+      }
+
+      var alreadyExists = pc.getSenders().find(function(s) {
```
fippo, line 274: note that now this shim relies on the shimGetSendersWithDtmf to be called later in order to provide getSenders(). Kind of weird but the tests prove it is working.

fippo: and I added an extra test just for getSenders

src/js/chrome/chrome_shim.js
```diff
@@ -247,6 +205,99 @@ var chromeShim = {
...
+
+    // also shim pc.getLocalStreams when addTrack is shimmed
+    // to return the original streams.
+    var origGetLocalStreams = window.RTCPeerConnection.prototype
+        .getLocalStreams;
+    window.RTCPeerConnection.prototype.getLocalStreams = function() {
+      var self = this;
+      var nativeStreams = origGetLocalStreams.apply(this);
+      return nativeStreams.map(function(stream) {
+        return self._reverseStreams[stream.id];
```
henbos: Do you need to do pc._reverseStreams = pc._reverseStreams || {}; before the map() in case getLocalStreams() is the first function called and _reverseStreams is undefined?

src/js/chrome/chrome_shim.js
```diff
@@ -247,6 +205,99 @@ var chromeShim = {
...
+      if (oldStream) {
+        // this is using odd Chrome behaviour, use with caution:
+        // https://bugs.chromium.org/p/webrtc/issues/detail?id=7815
+        oldStream.addTrack(track);
+        pc.dispatchEvent(new Event('negotiationneeded'));
+      } else {
+        var newStream = new window.MediaStream([track]);
+        pc._streams[stream.id] = newStream;
+        pc._reverseStreams[newStream.id] = stream;
+        pc.addStream(newStream);
```
henbos, line 294: I was going to suggest marking newStream as an "addTrack-stream" so that in getLocalStreams() we don't include it, yielding:
pc.addTrack(track, stream);
pc.getLocalStreams().includes(stream); // false

But it looks like Firefox does add addTrack-streams to local streams which would make this correct, and it should be tested.

test/test.js
```diff
@@ -2070,6 +2093,85 @@ test('dtmf', t => {
...
+      })
+      .then(e => {
+        // Test getSenders Chrome polyfill
+        try {
+          // FF51+ doesn't have removeStream
+          if (!('removeStream' in pc1)) {
+            throw new DOMException('', 'NotSupportedError');
+          }
+          // Avoid <FF51 throwing NotSupportedError - https://bugzil.la/1213441
+          pc1.removeStream(stream);
```
henbos, line 2135: Here and in other tests: Can you add a todo with a reference to #577 saying we should use removeTrack if available.

test/test.js
```diff
@@ -2707,6 +2809,105 @@ test('ontrack', function(t) {
...
+      },
+      pass: function(msg) {
+        this.ok(true, msg);
+      },
+      fail: function(msg) {
+        this.ok(false, msg);
+      }
+    };
+    var pc = new RTCPeerConnection();
+    navigator.mediaDevices.getUserMedia({audio: true})
```
henbos, line 2835: Can you add a test that makes sure a stream of an audio and video track and addTrack with one of them creates a single sender for that track alone? So that both are not added.

And a test that makes sure that multiple addTrack calls results in an exception (sender already exists).

src/js/chrome/chrome_shim.js
fippo: why would anyone... but 👍

src/js/chrome/chrome_shim.js
fippo: yes, we need more tests for this behaviour. In the new karma tests from #597 this will be much easier.

test/test.js
fippo: test('Basic connection establishment with addTrack ' +
    'and only the audio track of an av stream', function(t) {
-- this does a e2e test which ensure no stream is received at the remote end.

Multiple addTrack: yes, but in karma tests 'cause boilerplate ;-)

test/test.js
fippo: added a note to the issue instead ;-)

KaptenJansson:
@henbos do you approve of @fippo's comments?

henbos (approved):

KaptenJansson:
LGTM

PR #564, opened 2017-05-23
Add eslint-config-google and fix eslint errors

KaptenJansson:
**Description**
Use eslint-config-google but disable jsdoc and es6 rules. Tests are exempted and es6 rules are explicit enabled since we use es6 there.

**Purpose**
Adhere to the google style guide.

Note: This is not rebased with the latest changes yet, will do once it's ready before submitting this.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -4,6 +4,7 @@
     "block-spacing": [2, "never"],
     "brace-style": [2, "1tbs", {"allowSingleLine": false}],
     "camelcase": [2, {"properties": "always"}],
+    "comma-style": 2,
     "curly": 2,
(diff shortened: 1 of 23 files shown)


fippo:
instead of using eslint to fix var -> let i'd rather use https://github.com/cpojer/js-codemod#no-vars in test/

KaptenJansson:
What's the benefit of codemod over eslint? Smarter logic like being context aware?

fippo:
it uses const for things which (in its opinion) are not modified -- like the module requires. And its fairly good with that

KaptenJansson:
Closing for now.

PR #566, opened 2017-05-24
edge: add disconnected state

### LINKED ISSUE #545: Disconnected Event Appears as 'new' on _updateConnectionState
Edge shim.

**Description**
RTCPeerConnection.prototype._updateConnectionState states object does not define a disconnected property. When the transceiver.iceTransport.state or transceiver.dtlsTransport.state is 'disconnected' it results in a value of NaN which therefore emits the default state of 'new'.

**Expected results**
'disconnected' event should be emitted.

**Actual results**
'new' event is emitted.

fippo:
adds a disconnected state to the updateConnectionState algorithm.

Fixes #545


KaptenJansson:
LGTM

PR #567, opened 2017-05-25
build: allow publishing patch versions

fippo:
allows creating patch versions from a past git tag and adds a description how to do that.

@KaptenJansson wdyt? I think that is better than what I came up with earlier.
It also struck me that we do the copy to github on npm version and not npm publish.


KaptenJansson:
LGTM (note to self, make sure to be on the correct branch and right commit:))

PR #568, opened 2017-05-31
Adding stream management API

youennf:
Adding shiming for some legacy API methods related to stream management: getLocalStreams(), getRemoteStreams(), getStreamById() and removeStream().

### DIFF
--- a/src/js/safari/safari_shim.js
+++ b/src/js/safari/safari_shim.js
@@ -13,20 +13,94 @@ var safariShim = {
   // TODO: check for webkitGTK+
   // shimPeerConnection: function() { },
 
-  shimAddStream: function(window) {
-    if (typeof window === 'object' && window.RTCPeerConnection &&
-        !('addStream' in window.RTCPeerConnection.prototype)) {
+  shimLocalStreamsAPI: function(window) {
+    if (typeof window !== 'object' || !window.RTCPeerConnection) {
+      return;
+    }
+    if (!('getLocalStreams' in window.RTCPeerConnection.prototype)) {
+      window.RTCPeerConnection.prototype.getLocalStreams = function() {
+        if (!this._localStreams) {
+          this._localStreams = [];
+        }
+        return this._localStreams;
+      };
+    }
+    if (!('getStreamById' in window.RTCPeerConnection.prototype)) {
+      window.RTCPeerConnection.prototype.getStreamById = function(id) {
+        var result = null;
+        if (this._localStreams) {
+          this._localStreams.forEach(function(stream) {
+            if (stream.id === id) {
+              result = stream;
+            }
+          });
+        }
+        if (this._remoteStreams) {
+          this._remoteStreams.forEach(function(stream) {
+            if (stream.id === id) {
+              result = stream;
+            }
+          });
+        }
+        return result;
+      };
+    }
+    if (!('addStream' in window.RTCPeerConnection.prototype)) {
+      var _addTrack = window.RTCPeerConnection.prototype.addTrack;
       window.RTCPeerConnection.prototype.addStream = function(stream) {
+        if (!this._localStreams) {
+          this._localStreams = [];
+        }
+        if (this._localStreams.indexOf(stream) === -1) {
+          this._localStreams.push(stream);
+        }
         var self = this;
         stream.getTracks().forEach(function(track) {
-          self.addTrack(track, stream);
+          _addTrack.call(self, track, stream);
+        });
+      };
+
+      window.RTCPeerConnection.prototype.addTrack = function(track, stream) {
+        if (stream) {
+          if (!this._localStreams) {
+            this._localStreams = [stream];
+          } else if (this._localStreams.indexOf(stream) === -1) {
+            this._localStreams.push(stream);
+          }
+        }
+        _addTrack.call(this, track, stream);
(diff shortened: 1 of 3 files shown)


src/js/safari/safari_shim.js
```diff
@@ -13,20 +13,94 @@ var safariShim = {
...
+    }
+    if (!('getLocalStreams' in window.RTCPeerConnection.prototype)) {
+      window.RTCPeerConnection.prototype.getLocalStreams = function() {
+        if (!this._localStreams) {
+          this._localStreams = [];
+        }
+        return this._localStreams;
+      };
+    }
+    if (!('getStreamById' in window.RTCPeerConnection.prototype)) {
```
fippo, line 28: @foolip killed that in Chrome recently (or planned to due to non-spec and negligible usage) so I don't think this one should be shimmed.

test/unit/safari.js
```diff
@@ -22,6 +22,62 @@ describe('Safari shim', () => {
     };
   });
 
+  describe('shimStreamsAPI', () => {
+    beforeEach(() => {
+      window.RTCPeerConnection.prototype.addTrack = () => {};
```
fippo: = sinon.stub() (even though I don't think you need to make a .called assertion below...)

test/unit/safari.js
```diff
@@ -22,6 +22,62 @@ describe('Safari shim', () => {
...
+      const prototype = window.RTCPeerConnection.prototype;
+      expect(prototype.addTrack.length).to.equal(2);
+      expect(prototype.addStream.length).to.equal(1);
+      expect(prototype.removeStream.length).to.equal(1);
+      expect(prototype.getLocalStreams.length).to.equal(0);
+      expect(prototype.getStreamById.length).to.equal(1);
+      expect(prototype.getRemoteStreams.length).to.equal(0);
+    });
+    it('local streams API', () => {
+      const pc = new window.RTCPeerConnection();
```
fippo, line 42: move to beforeEach, including getSenders and the artifical streams

test/unit/safari.js
```diff
@@ -22,6 +22,62 @@ describe('Safari shim', () => {
...
+      pc.getSenders = () => {
+        return [];
+      };
+      var stream = {id: 'id1', getTracks: () => {
+        return [];
+      }};
+      expect(pc.getStreamById(stream.id)).to.equal(null);
```
fippo, line 49: each of these can be a separate it() -- requires a bit of effort to recreate the state in later stages. I can make a PR after this is merged if you want :-)

fippo (approved):
LGTM with style-nits on the tests.
I'll let @foolip decide on getStreamById

src/js/safari/safari_shim.js
youennf: It does not really hurt to have it here either. I saw it on some WebRTC code base but maybe this is very infrequently used.

test/unit/safari.js
youennf: In which scope pc and streams would be defined then?
Might be better to do it at test separation time.

test/unit/safari.js
youennf: Done

test/unit/safari.js
youennf: OK, a follow-up PR sounds good to me.

test/unit/safari.js
KaptenJansson: You could cleanup by using afterEach as well.

KaptenJansson (approved):
LGTM with potential improvements for tests but that can be done in a followup PR as suggested by @fippo.

test/unit/safari.js
fippo: will do when splitting up tests. For the mock'd peerconnection it does not matter but when you make the karma tests run in the browser weird things happen otherwise :-)

aisnote:
good to know

test/unit/safari.js
fippo: ah... doesn't work since pc is just a very minimal mock here which doesn't have a stop method even.

jan-ivar:
We should add deprecation warnings to all these, or this won't drive convergence toward the spec. Adapter's main mission for a while now has been to eventually disappear.

jan-ivar:
Is there enough usage of `getStreamById()` to warrant polyfilling?

foolip:
getStreamById will be gone in Chrome 62, and that removal is possible because the usage is super low: https://www.chromestatus.com/metrics/feature/timeline/popularity/1645

youennf:
> We should add deprecation warnings to all these, or this won't drive convergence toward the spec. Adapter's main mission for a while now has been to eventually disappear.

Sounds good, I would prefer doing it as a follow-up PR to also add warnings to navigator.getUserMedia and callback-based APIs.

> getStreamById will be gone in Chrome 62, and that removal is possible because the usage is super low: https://www.chromestatus.com/metrics/feature/timeline/popularity/1645

I am fine killing it. I would prefer doing it sometimes in this summer.

youennf:
Added some warning messages  at https://github.com/youennf/adapter/commit/bdf41a26a76fd5e46734bc1762560989cf998c21

youennf:
Can the PR be merged as is?

PR #571, opened 2017-06-04
chrome: change addTrack call order

fippo:
moves removeTrack a bit earlier so that removeTrack is
called with the same track that was added. Without this, the following happens:
```
pc.addStream(stream with just audio track)
pc.removeStream(stream with audio and video track)
pc.addStream(stream with audio and video track)
```
which makes me a bit uneasy.

Still pondering if I should intercept ONN to avoid https://bugs.chromium.org/p/webrtc/issues/detail?id=7693 ...


fippo:
#580 is a conflicting PR. There can be only one

PR #572, opened 2017-06-06
utils: add utils.deprecated helper

fippo:
adds a utils.deprecated helper that can be used to suggest a
spec-compliant alternative to a legacy behaviour.

Warnings can be turned off by calling disableWarnings but are
enabled by default.

@youennf @jan-ivar @KaptenJansson PTAL. I would very much prefer to call this "nag" but... ;-)

### DIFF
--- a/src/js/adapter_factory.js
+++ b/src/js/adapter_factory.js
@@ -22,7 +22,8 @@ module.exports = function(dependencies) {
   var adapter = {
     browserDetails: browserDetails,
     extractVersion: utils.extractVersion,
-    disableLog: utils.disableLog
+    disableLog: utils.disableLog,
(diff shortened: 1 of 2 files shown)


youennf:
That sounds good to me.
Might be good to add support for legacy warnings to callback-based APIs as well.

fippo:
test failure in chrome-stable seems to be related to https://travis-ci.org/webrtc/adapter/jobs/239827362#L794 where we get the wrong chrome version. Should go away, result from http://browsers.contralis.info/linux/chrome/stable shows 59 now.

jan-ivar (approved):
lgtm

KaptenJansson:
LGTM

PR #574, opened 2017-06-07
Transition adapter.js to ES modules

Victorystick:
**Description**
Prefer static `import`/`export` declarations to the dynamic CommonJS `require` calls.

**Purpose**
We can transpile `import`/`export` into whatever module format users want to consume. It's not as easy to work with CommonJS, nor is it as easy to optimize. Using Rollup to produce the pre-built release bundles also results in slightly smaller output than Browserify.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -1,51 +1,55 @@
 'use strict';
+var commonjs = require('rollup-plugin-commonjs');
+var ignore = require('rollup-plugin-ignore');
+var nodeResolve = require('rollup-plugin-node-resolve');
+
+var ignoreEdge = [ignore('./edge/edge_shim.js')];
 
 module.exports = function(grunt) {
   grunt.initConfig({
     pkg: grunt.file.readJSON('package.json'),
-    browserify: {
+    rollup: {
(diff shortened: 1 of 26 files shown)


Gruntfile.js
```diff
@@ -1,51 +1,55 @@
...
     pkg: grunt.file.readJSON('package.json'),
-    browserify: {
+    rollup: {
+      options: {
+        format: 'umd',
+        moduleName: 'adapter',
+        indent: false,
+        plugins: [
+          nodeResolve(),
+          commonjs({ include: 'node_modules/sdp/**/*.js' })
```
fippo, line 18: won't that include the tests from https://github.com/fippo/sdp/tree/master/test ?

Victorystick: The commonjs plugin only processes files that are actually imported, so it should be fine as long as you don't import `sdp`'s tests. :stuck_out_tongue:

KaptenJansson:
Closing for now since there is not enough time to do this big change and all the follow up. Can be reopened in the future if there is time and commitment to do so.

PR #576, opened 2017-06-07
test: only use getBrowserVersion for chrome and firefox

fippo:
(un)breaks browsers that do not follow the travis multirunner layout


KaptenJansson:
LGTM

PR #578, opened 2017-06-08
test: add safari webdriver support

fippo:
tests are still not running but this adds more infrastructure to make them run


fippo:
@youennf any preferences on "TechPreview" vs "TechnologyPreview"?

youennf:
Either is good, TechnologyPreview might be slightly more consistent

fippo:
I squeezed it in :-)

KaptenJansson:
LGTM

PR #579, opened 2017-06-08
Update Safari shim to handle window.URL.createObjectURL

neilkinnish:
**Description**

Adds the shim for Safari (technical preview) to handle: window.URL.createObjectURL

### DIFF
--- a/src/js/adapter_factory.js
+++ b/src/js/adapter_factory.js
@@ -93,7 +93,8 @@ module.exports = function(dependencies) {
       logging('adapter.js shimming safari.');
       // Export to the adapter global object visible in the browser.
       adapter.browserShim = safariShim;
-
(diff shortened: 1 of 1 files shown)


neilkinnish:
I signed it!

fippo:
To minimize impact (i've been hit by this but am using it as an excuse to get rid of the last createObjectURL bits) I think we should do this.

@youennf your call. @jan-ivar might have an opinion too

youennf:
That looks fine to me.
This can be disabled here all at once for all browsers whenever feasible.

fippo:
I am looking forward to the "goodbye legacy stuff" major version bump very much :-)

KaptenJansson:
LGTM

PR #580, opened 2017-06-08
addTrack: synthetic ONN

fippo:
fires ONN in JS instead of trying to make Chrome fire ONN (which it does too often)

This has a disadvantage though. In chrome://webrtc-internals addStream will only be called with an audio stream while before createOffer will be called in a state where the stream has both audio and video tracks.


fippo:
the other variant which calls removeStream might have really bad consequences for getSenders()... cc @henbos

fippo:
so removeStream-addStream creates a new RTCRtpSender which makes this variant "better" than #571. Will incorporate this into #559

PR #583, opened 2017-06-09
test: make getBrowserVersion safe to use

fippo:
makes getBrowserVersion safe to use on non-linux.
Currently breaks for both Edge and Safari.

It is amazing how long the issue has existed and we didn't notice because `()` was missing :-/


KaptenJansson:
Hmm interesting indeed. LGTM

PR #584, opened 2017-06-10
safari: shim RTCIceServer.url and nag about it

fippo:
(depends on #572)

Safari does not natively support RTCIceServer.url.
url

Shim it and nag users to update. Code is copied from the Chrome shim that does the same job but deletes RTCIceServer.url

@youennf please take a look. Also 👏 for enforcing the spec


fippo:
flaky dtmf test, @KaptenJansson needs to kick it :-/

KaptenJansson:
LGTM! Sinon is really neat!

PR #585, opened 2017-06-12
chrome: make addTrack work with native senders

### LINKED ISSUE #552: Chrome implements getSenders

fippo:
makes addTrack work with native senders and fixes #552

Probably needs to be adapted depending on the outcome of #580 vs #571


fippo:
@jan-ivar do you think this is going to work out with the dtmf shim? Should we have a test?

fippo:
and this needs the getLocalStreams shim from #559 as well

fippo:
incorporated into #559

PR #587, opened 2017-06-14
Use session ID for each peer connection on Edge

armouroflight:
We have seen that the adapter.js currently uses the same hardcoded session ID for all SDP.  This makes it a random session ID for each RTCPeerConnection which is like other browsers have been seen to do.

This change relies on changes in sdp.js which I have put in a PR for.  Let me know if you need any more details.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "test": "grunt && mocha test/unit && node test/run-tests.js"
   },
   "dependencies": {
-    "sdp": "git://github.com/armouroflight/sdp.git#cc26d22"
+    "sdp": "^2.1.0"
   },
   "engines": {
     "npm": ">=3.10.0",
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "test": "grunt && mocha test/unit && node test/run-tests.js"
   },
   "dependencies": {
-    "sdp": "^1.5.0"
+    "sdp": "git://github.com/armouroflight/sdp.git#cc26d22"
   },
   "engines": {
     "npm": ">=3.10.0",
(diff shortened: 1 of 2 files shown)


armouroflight:
I signed it!

package.json
```diff
@@ -19,7 +19,7 @@
     "test": "grunt && mocha test/unit && node test/run-tests.js"
   },
   "dependencies": {
-    "sdp": "^1.5.0"
+    "sdp": "git://github.com/armouroflight/sdp.git#cc26d22"
```
fippo: i can publish a new minor version (in the 2.x line) for this. Will come back with an exact version.
This will be a major bump of the sdp package but its ok, the other parts are prepared for this.

src/js/edge/rtcpeerconnection_shim.js
```diff
@@ -1153,7 +1155,7 @@ module.exports = function(window, edgeVersion) {
     // reorder tracks
     var transceivers = sortTracks(this.transceivers);
 
-    var sdp = SDPUtils.writeSessionBoilerplate();
+    var sdp = SDPUtils.writeSessionBoilerplate(this._sdpSessionId);
```
fippo, line 1158: this also needs to be done at the beginning of createAnswer.

fippo (approved):
(wrong comment)

fippo (changes requested):
LGTM but needs a different approach for the SDP package

src/js/edge/rtcpeerconnection_shim.js
armouroflight: Done!

armouroflight:
Let me know if you still have issues!   Thanks!

fippo (approved):
LGTM, over to @KaptenJansson -- this is patch imo

KaptenJansson:
LGTM

PR #589, opened 2017-06-15
travis: add chrome stable/beta without experiments

fippo:
Since users do not run chrome stable or beta with the
experimental web platform features flag on we should
also reflect this in our chrome test environment.


fippo:
quite unexpected that tests pass on beta with experimental web platform flag even, I thought that was broken?! Well... no breakage we can test this on.

KaptenJansson:
LGTM

PR #591, opened 2017-06-17
adapter_factory: allow disabling shim from options

### LINKED ISSUE #512: Save the original native RTCPeerConnection in Edge 15.15019+
Edge 15.15019

**Description**
Request to allow webrtc/adapter to save the original RTCPeerConnection object to msRTCPeerConnection or any prefix name so that users can decide to use the ORTC shim or the native legacy WebRTC implementation.

fippo:
adds an extra argument to the adapter factory which allows
disabling certain shims.

Fixes #512 

@letchoo holler if this doesn't solve the problem for you

### DIFF
--- a/src/js/adapter_factory.js
+++ b/src/js/adapter_factory.js
@@ -10,8 +10,15 @@
 'use strict';
 
 // Shimming starts here.
-module.exports = function(dependencies) {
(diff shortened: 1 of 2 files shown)


KaptenJansson:
LGTM

PR #597, opened 2017-06-25
test: use karma

fippo:
uses karma + travis-multirunner for tests and ports the first test to a mocha/chai one run by karma.

(lets see if I got the travis bits right, in particular using the travis-multirunner versions...)


fippo:
@KaptenJansson what is the best practice for a one-line copyright header? The full one seems onerous on the small test/e2e/ files

fippo:
I got carried away 😊

aullman:
@fippo Ooh, I like it. Is the idea to replace WebDriver with Karma? Makes sense to me. WebDriver is nice for testing UI but clunky for testing a library I have found with all the `driver.executeScript()` calls.

This will save us having to figure out how to turn on "Allow Remote Automation" in Safari as well right?

fippo:
@aullman yeah, this is also moving away from tape gradually. I got carried away a bit after pondering about https://github.com/DamonOehlman/travis-multirunner/issues/17 ;-)
There is much less boilerplate involved which makes writing tests easier. Easier tests -> more tests.

Oddly, karma-firefox-launcher seems to work in FF-unstable where tests have been broken for a while
We still need a solution for "Allow Remote Automation" for travis + safari but at least we won't have to deal with weird behaviour on file://

KaptenJansson (changes requested):
Wow you got carried away :). I agree that we should phase out tape as we go along.

LGTM with a nit: add the copyright header to all files regardless of the size/line count of the file, I think that's fine.

fippo:
I considered porting the DTMF test too but that looks like it is currently intertwined with getSenders et al stuff so i'd like to keep that separate.

Skipping a test seems quite ugly (from connection.js):
```
    beforeEach(function() {
      if (window.adapter.browserDetails.browser === 'edge') {
        this.skip();
      }
    });
```
@thepatrick do you folks happen to know a better way?

fippo:
feature of the day: make karma do its own browserify instead of using the generated bundle. Now I am somewhat keen to say if that means istanbul works...

KaptenJansson (approved):

KaptenJansson:
Yeah we can explore more options with karma, it can do lots of things like running the linters on save/file change etc ;)

LGTM and thanks a lot for doing this!

PR #600, opened 2017-06-27
Edge: only set remote candidates once for answer with a=group:BUNDLE

### LINKED ISSUE #599: Edge: InvalidStateError when remote answer has a=group:BUNDLE

Edge 15.15063

**Description**

If `setRemoteDescription` is called with a remote answer containing `a=group:BUNDLE` and multiple media sections, the second call to `iceTransport.setRemoteCandidates` will throw InvalidStateError, probably because `iceTransport.start` is called during the first section.

**Steps to reproduce**

- Create peer connection
- Add audio and video tracks
- Create an offer from Edge and send it to the remote endpoint (in my case, an SFU)
- Set local description
- Receive an answer from remote endpoint with `a=group:BUNDLE` in the session part of the SDP answer
- Set remote description

**Expected results**

- Successfully establish connnection

**Actual results**

- `iceTransport.setRemoteCandidates` throws `InvalidStateError`

aiham:
**Description**

- In the Edge shim for `setRemoteDescription`, only call `iceTransport.setRemoteCandidates` once  for a remote answer if usingBundle is true
- Also only call `iceTransport.setRemoteCandidates` if remote offer has 1 or more candidates

**Purpose**

- Resolves #599

### DIFF
--- a/src/js/edge/rtcpeerconnection_shim.js
+++ b/src/js/edge/rtcpeerconnection_shim.js
@@ -798,7 +798,8 @@ module.exports = function(window, edgeVersion) {
               self._createIceGatherer(mid, sdpMLineIndex);
         }
 
-        if (isComplete && (!usingBundle || sdpMLineIndex === 0)) {
+        if (isComplete && cands.length &&
+            (!usingBundle || sdpMLineIndex === 0)) {
           transceiver.iceTransport.setRemoteCandidates(cands);
         }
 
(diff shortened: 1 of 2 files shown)


aiham:
I've made an assumption that only candidates from the first media section are necessary. I'm basing that on https://github.com/webrtc/adapter/blob/cf321e9db865cedc8c7140c62c6cfd699f4ae602/src/js/edge/rtcpeerconnection_shim.js#L801

aiham:
@fippo Looks like your repo has diverged a little: https://github.com/fippo/rtcpeerconnection-shim/blob/107435efa21ee2093f69de4a4e6ee73d05a1a4b4/rtcpeerconnection.js#L886

Btw, the check for `iceTransport.state === 'new'` doesn't resolve #599 for me as I've observed the state to be `new` when InvalidStateError is thrown, even though https://msdn.microsoft.com/en-us/library/mt433113(v=vs.85).aspx only mentions it happening if state is `closed`.

Any suggestions?

fippo:
this issue sounds very familiar... for offers my version takes bundle into account [here](https://github.com/fippo/rtcpeerconnection-shim/blob/master/rtcpeerconnection.js#L800).
For answers it does not (because such SDP is not valid technically in bundle) but I don't see much harm in having a similar check for the sdpMLineIndex = 0 in answers too.

fippo (approved):

KaptenJansson:
LGTM

PR #601, opened 2017-06-28
test: support travis multirunner BVER in selenium lib

fippo:
supports both variants of BVER for safari technology preview -- unstable coming from https://github.com/DamonOehlman/travis-multirunner/pull/23


KaptenJansson:
LGTM with a note to add this to utilities.

PR #604, opened 2017-06-29
test: set SAFARI_BIN in karma.conf

fippo:
sets SAFARI_BIN to point to Safari Technology Preview
when BVER is set to experimental. This makes karma tests run in STP.

### DIFF
--- a/test/karma.conf.js
+++ b/test/karma.conf.js
@@ -27,6 +27,13 @@ if (process.env.BROWSER) {
   browsers = ['chrome', 'firefox'];
 }
 
+// uses Safari Technology Preview.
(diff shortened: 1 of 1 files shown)


fippo:
should this go to .travis.yml instead?

KaptenJansson:
If travis specific, yes.

fippo:
its quite useful for local usage too... will update with setting CHROME_BIN and FIREFOX_BIN as well (when on linux)

aullman:
Another option is to use https://www.npmjs.com/package/karma-safaritechpreview-launcher

fippo:
yeah but using a npm module for a single env var... ;-)

KaptenJansson:
Yeah that is true. LGTM.

PR #609, opened 2017-07-03
test: provide expected karma run results as files

fippo:
this allows comparing the results of a testsuite run to a previous run.
Also re-enables the "no dtmf sender on video rtpsenders" test which was broken in Firefox but that is now a "known failure".

We'll see how that goes moving forward. Might be a bit too much effort updating those files all the time.
To update them, run the karma test for all browsers that need updating with stabilityReporter.update set to true.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
     "karma-mocha": "^1.3.0",
     "karma-mocha-reporter": "^2.2.3",
     "karma-safari-launcher": "^1.0.0",
+    "karma-stability-reporter": "git+https://github.com/fippo/karma-stability-reporter.git#v1.0.0",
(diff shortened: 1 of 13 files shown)


package.json
```diff
@@ -48,6 +48,7 @@
     "karma-mocha": "^1.3.0",
     "karma-mocha-reporter": "^2.2.3",
     "karma-safari-launcher": "^1.0.0",
+    "karma-stability-reporter": "git+https://github.com/fippo/karma-stability-reporter.git#v1.0.0",
```
KaptenJansson, line 51: Would prefer this to be a NPM package for consistency but we can do that later.

KaptenJansson (approved):
LGTM

PR #610, opened 2017-07-03
test: add tests for addTrack

fippo:
adds tests for addTrack which were promised during the review of #559

@henbos PTAL, do you want more tests?
@jan-ivar looks like you throw InvalidParameterError instead. Filed a [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1377884) for you. We can soon mark that test as expected failure


test/e2e/addTrack.js
```diff
@@ -0,0 +1,110 @@
...
+  });
+
+  describe('and getLocalStreams', () => {
+    it('returns a stream with audio and video even if just an ' +
+        'audio track was added', () => {
+      return navigator.mediaDevices.getUserMedia({audio: true, video: true})
+      .then(stream => {
+        pc.addTrack(stream.getTracks()[0], stream);
+        const localStreams = pc.getLocalStreams();
+        expect(localStreams).to.have.length(1);
```
henbos, line 92: To confirm, is this what Firefox does?
I would have expected getLocalStreams() to be unaffected by addTrack(). But if this is the case I might have to modify getLocalStreams(), perhaps to return every stream of every sender instead of a set of streams modified on add/removeStream() before shipping addTrack() in Chrome.

test/e2e/addTrack.js
```diff
@@ -0,0 +1,110 @@
...
+        const localStreams = pc.getLocalStreams();
+        const localStream = localStreams[0];
+        const track = stream.getTracks()[0];
+        localStream.addTrack(track);
+        pc.addTrack(track, localStream);
+        expect(localStreams).to.have.length(1);
+        expect(localStreams[0].getTracks()).to.have.length(2);
+        expect(pc.getSenders()).to.have.length(2);
+      });
+    });
```
henbos, line 133: Can you add a test that does...?
stream = audio and video
addTrack(stream.audio);
expect getLocalStreams() == 1, getSenders() == 1
addStream(stream);
expect getLocalStreams() == 1, getSenders() == 2

test/e2e/addTrack.js
```diff
@@ -0,0 +1,110 @@
...
+    it('if the track has already been added', () => {
+      return navigator.mediaDevices.getUserMedia({audio: true})
+      .then(stream => {
+        pc.addTrack(stream.getTracks()[0], stream);
+        const again = () => {
+          pc.addTrack(stream.getTracks()[0], stream);
+        };
+        expect(again).to.throw(/already/)
+          .that.has.property('name').that.equals('InvalidAccessError');
+      });
```
henbos, line 32: Can you add a test that makes sure it also throws an exception on addTrack if a sender was already created with addStream?

henbos (changes requested):
Loving these new tests!

test/e2e/addTrack.js
fippo: clone the repo, checkout [the branch for this PR](https://help.github.com/articles/checking-out-pull-requests-locally/), setenv or export BROWSER=firefox and run npm test. Add console.log as desired, this is now super easy ;-)

jan-ivar: @henbos Yeah that's what Firefox does atm. I should double-check with @docfaraday what will happen with `getLocalStreams()` once we're done with transceivers.

test/e2e/addTrack.js
fippo: Nice one! This currently fails because I am not creating an internal stream for addStream streams (required by and done in #611)

Firefox throws here since it has already been added:
```
  describe('and addStream', () => {
    it('adds unadded tracks of a stream', () => {
      return navigator.mediaDevices.getUserMedia({audio: true, video: true})
      .then(stream => {
        pc.addTrack(stream.getTracks()[0], stream);
        const localStreams = pc.getLocalStreams();
        expect(localStreams).to.have.length(1);
        expect(pc.getSenders()).to.have.length(1);

        pc.addStream(stream);
        expect(pc.getLocalStreams()).to.have.length(1);
        expect(pc.getSenders()).to.have.length(2);
      });
    });
  });
```
I have a slight preference for the Firefox behaviour since we throw when calling addTrack with the same track twice in the spec

henbos: Throwing because one of the tracks have been added when you addStream() makes sense too. We'll have to make sure it adds all or none of the tracks though so it doesn't fail halfway through adding tracks of a stream. This is achieved if getLocalStream() behaves as above (how Firefox works) and if addStream() checks if getLocalStreams() has the stream. That's a straight-forward implementation that makes sense.

henbos: Just expect it to throw on addStream and we're good.

fippo: so that should throw BUT pc.getSenders() should still be 1. Hrm... I think that can be achieved by calling addTrack with the [1] track and then calling addStream.

test/e2e/addTrack.js
henbos: I filed https://crbug.com/738918

test/e2e/addTrack.js
henbos: Firefox polyfills, so it does not do an all-or-nothing, it just forEach track does addTrack(track, stream). Considering this is legacy, I'm ok with adopting that behavior.

fippo: yes. We should should not do it if we ever decide to actually specify the behaviour (-:

fippo: added a test which tests for the exception

test/e2e/addTrack.js
fippo: done

fippo:
squash and we're good? Need to add expectations for #609

henbos (approved):
LGTM

jan-ivar:
> @jan-ivar looks like you throw InvalidParameterError instead. Filed a bug for you. We can soon mark that test as expected failure

@fippo Thanks! BTW, shouldn't this be a WPT test as well?

fippo:
made a minor change to the chrome shim to make this not work:
```
pc.addTrack(stream.getTracks()[0], stream);
pc.addStream(stream)
```
moar tests ;-)

PR #611, opened 2017-07-03
chrome: shim removeTrack

fippo:
shims pc.removeTrack.

@henbos if you're bored... ;-)

### DIFF
--- a/src/js/chrome/chrome_shim.js
+++ b/src/js/chrome/chrome_shim.js
@@ -71,6 +71,7 @@ var chromeShim = {
   },
 
   shimGetSendersWithDtmf: function(window) {
+    // Overrides addTrack/removeTrack, depends on shimAddTrackRemoveTrack.
     if (typeof window === 'object' && window.RTCPeerConnection &&
         !('getSenders' in window.RTCPeerConnection.prototype) &&
         'createDTMFSender' in window.RTCPeerConnection.prototype) {
@@ -86,14 +87,16 @@ var chromeShim = {
               }
             }
             return this._dtmf;
-          }
+          },
+          _pc: pc
         };
       };
 
-      // shim addTrack when getSenders is not available.
+      // augment addTrack when getSenders is not available.
       if (!window.RTCPeerConnection.prototype.getSenders) {
         window.RTCPeerConnection.prototype.getSenders = function() {
-          return this._senders || [];
+          this._senders = this._senders || [];
+          return this._senders.slice(); // return a copy of the internal state.
         };
         var origAddTrack = window.RTCPeerConnection.prototype.addTrack;
         window.RTCPeerConnection.prototype.addTrack = function(track, stream) {
@@ -104,6 +107,16 @@ var chromeShim = {
             pc._senders.push(sender);
           }
           return sender;
+        };
+
+        var origRemoveTrack = window.RTCPeerConnection.prototype.removeTrack;
+        window.RTCPeerConnection.prototype.removeTrack = function(sender) {
+          var pc = this;
+          origRemoveTrack.apply(pc, arguments);
+          var idx = pc._senders.indexOf(sender);
+          if (idx !== -1) {
+            pc._senders.splice(idx, 1);
+          }
         };
       }
       var origAddStream = window.RTCPeerConnection.prototype.addStream;
@@ -156,7 +169,7 @@ var chromeShim = {
             }
           }
           return this._dtmf;
-        },
+        }
       });
     }
   },
@@ -205,8 +218,8 @@ var chromeShim = {
     }
   },
 
-  shimAddTrack: function(window) {
-    // shim addTrack (when getSenders is available)
+  shimAddTrackRemoveTrack: function(window) {
+    // shim addTrack and removeTrack.
     if (window.RTCPeerConnection.prototype.addTrack) {
       return;
     }
@@ -242,8 +255,10 @@ var chromeShim = {
       // Add identity mapping for consistency with addTrack.
       // Unless this is being used with a stream from addTrack.
       if (!pc._reverseStreams[stream.id]) {
-        pc._streams[stream.id] = stream;
-        pc._reverseStreams[stream.id] = stream;
+        var newStream = new window.MediaStream(stream.getTracks());
+        pc._streams[stream.id] = newStream;
+        pc._reverseStreams[newStream.id] = stream;
+        stream = newStream;
       }
       origAddStream.apply(pc, [stream]);
     };
@@ -294,6 +309,8 @@ var chromeShim = {
       if (oldStream) {
         // this is using odd Chrome behaviour, use with caution:
         // https://bugs.chromium.org/p/webrtc/issues/detail?id=7815
+        // Note: we rely on the high-level addTrack/dtmf shim to
+        // create the sender with a dtmf sender.
         oldStream.addTrack(track);
         pc.dispatchEvent(new Event('negotiationneeded'));
       } else {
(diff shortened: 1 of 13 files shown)


fippo:
Firefox does not throw an error when removing a track that does not belong to the pc. Filed [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1377888) for @jan-ivar ;-)

src/js/chrome/chrome_shim.js
```diff
@@ -205,7 +206,7 @@ var chromeShim = {
     }
   },
 
-  shimAddTrack: function(window) {
+  shimAddTrackRemoveTrack: function(window) {
     // shim addTrack (when getSenders is available)
     if (window.RTCPeerConnection.prototype.addTrack) {
       return;
```
henbos: This will also skip shimming removeTrack if addTrack already exists (is supported or has already been shimmed). The "shim addTrack when getSenders is is not available" case will shim addTrack but not removeTrack, see previous comment.

src/js/chrome/chrome_shim.js
```diff
@@ -93,7 +93,8 @@ var chromeShim = {
       // shim addTrack when getSenders is not available.
```
henbos: What about shim removeTrack when getSenders is not available? If addTrack is shimmed here and it already exists in the later code, it will skip shimming removeTrack because it skips shimming both addTrack and removeTrack

src/js/chrome/chrome_shim.js
```diff
@@ -297,6 +300,47 @@ var chromeShim = {
         return s.track === track;
       });
     };
+
+    window.RTCPeerConnection.prototype.removeTrack = function(sender) {
```
henbos, line 327: Can you add a comment to say which case this applies too, i.e. "when getSenders is available". If this should apply to both cases, make sure this doesn't abort in that case (see previous comments) and say in the comment that it applies to both.

src/js/chrome/chrome_shim.js
```diff
@@ -297,6 +300,47 @@ var chromeShim = {
...
+        return s === sender;
+      });
+      if (!isLocal) {
+        throw new DOMException('Sender was not created by this connection.',
+            'InvalidAccessError');
+      }
+
+      // Search for the native stream the senders track belongs to.
+      pc._streams = pc._streams || {};
+      var stream;
```
henbos, line 342: let stream = null;

src/js/chrome/chrome_shim.js
```diff
@@ -297,6 +300,47 @@ var chromeShim = {
...
+      Object.keys(pc._streams).forEach(function(streamid) {
+        var hasTrack = pc._streams[streamid].getTracks().find(function(track) {
+          return sender.track === track;
+        });
+        if (hasTrack) {
+          stream = pc._streams[streamid];
+        }
+      });
```
henbos, line 350: if (stream == null) skip the removeStream/removeTrack step, but still dispatch a negotiationneeded event...
Hmh, It would seem silly to fire an negotiationneeded event if nothing changed, but that seems to be what the spec suggests. Perhaps do skip it and file a spec bug not to fire the event if the sender was already inactive? I'm guessing that's a spec bug.

src/js/chrome/chrome_shim.js
```diff
@@ -297,6 +300,47 @@ var chromeShim = {
...
+        throw new DOMException(
+          'The RTCPeerConnection\'s signalingState is \'closed\'.',
+          'InvalidStateError');
+      }
+      var isLocal = pc.getSenders().find(function(s) {
+        return s === sender;
+      });
+      if (!isLocal) {
+        throw new DOMException('Sender was not created by this connection.',
+            'InvalidAccessError');
```
henbos: The correct behavior when removing a track is to make the sender inactive, but not to remove it. We (both adapter and chrome) remove senders that are no longer used, and this I think we should continue to do until we have transceiver support.

But a side-effect is that this is not a reliable way to check if the sender was created by this connection. If you remove the same sender twice the second will throw an exception instead of be a NO-OP. I would rather skip throwing an exception if you use another PC's sender than to have it throw an exception if you remove an already removed sender since this should be a more common use case.

test/e2e/removeTrack.js
```diff
@@ -0,0 +1,99 @@
...
+      .then(stream => {
+        const pc2 = new RTCPeerConnection();
+        const sender = pc2.addTrack(stream.getTracks()[0], stream);
+        const invalidSender = () => {
+          pc.removeTrack(sender);
+        };
+        expect(invalidSender).to.throw(/not created/)
+            .that.has.property('name').that.equals('InvalidAccessError');
+        pc2.close();
+      });
```
henbos, line 34: Add a test that makes sure it is OK to do removeTrack twice.

test/e2e/removeTrack.js
```diff
@@ -0,0 +1,99 @@
...
+        };
+        expect(afterClose).to.throw(/closed/)
+          .that.has.property('name').that.equals('InvalidStateError');
+      });
+    });
+  });
+
+  ['addStream', 'addTrack'].forEach(variant => {
+    describe('after ' + variant + ' for an audio/video track', () => {
+      beforeEach(() => {
```
henbos, line 65: beforeEach... Awesome testing tools 👍

henbos (changes requested):
Looking good!

src/js/chrome/chrome_shim.js
fippo: shimGetSendersWithDtmf is called after this. The comment is a bit misleading now, thanks for catching that!

src/js/chrome/chrome_shim.js
fippo: see above. Do you intend to ship addTrack without removeTrack (you can probably ship but i would not do so unflagged)
The behaviour of both shims is heavily intertwined, i don't think it makes much sense to untangle them.

src/js/chrome/chrome_shim.js
fippo: uh... I think you caught a missing removeTrack shim that cleans up the _senders stuff!

src/js/chrome/chrome_shim.js
fippo: I discussed with @jan-ivar a few days ago and... https://github.com/w3c/webrtc-pc/pull/1434 ;-)
Now need to dogfood too

src/js/chrome/chrome_shim.js
fippo: Right. Ideally we would set sender.track = null but I don't think we can control the behaviour enough to be spec-compliant.

fippo: What is the use-case here -- holding on to the result of pc.getSenders()? I hope people don't do that

jan-ivar:
> Firefox does not throw an error when removing a track that does not belong to the pc. 

@fippo Thanks! Again it'd be nice if this were a WPT tests as well I think...

src/js/chrome/chrome_shim.js
fippo: it wasn't missing but _senders was called in the addTrackRemoveTrack shim -- and it should have been called in the getSendersWithDtmf shim. Moved there

src/js/chrome/chrome_shim.js
henbos: No that's fine, as long as that assumption is clear with a comment. addTrack and removeTrack will be shipped at the same time.

src/js/chrome/chrome_shim.js
henbos: Yes, as of now you would only hit this problem if you held on to senders (getSenders() or addTrack() return values, maybe the app has its own map for some reason) after being removed. It's also possible that a rollback happens in which case a sender is no longer used.

If getSenders() is updated you could get inactive senders too, someone might write "remove all tracks" code that just loops through all senders and does remove, including already inactive senders. That could work now but break later if this throws and exception and transceivers are implemented.

fippo: right. But you plan to ship the real thing way ahead of that (and also really soon now?) and then cruft won't be necessary anymore?

test/e2e/removeTrack.js
fippo: done. Changed the logic a bit so that each sender has a ._pc

src/js/chrome/chrome_shim.js
```diff
@@ -105,6 +107,13 @@ var chromeShim = {
           }
           return sender;
         };
+
+        var origRemoveTrack = window.RTCPeerConnection.prototype.removeTrack;
+        window.RTCPeerConnection.prototype.removeTrack = function(sender) {
+          var pc = this;
+          origRemoveTrack.apply(pc, arguments);
+          pc._senders.splice(pc._senders.indexOf(sender), 1);
```
henbos: If you removeTrack with a sender that does not exist this will remove the last item of _senders. Make sure indexOf(sender) is not -1.

src/js/chrome/chrome_shim.js
```diff
@@ -105,6 +107,13 @@ var chromeShim = {
           }
           return sender;
         };
+
+        var origRemoveTrack = window.RTCPeerConnection.prototype.removeTrack;
```
henbos, line 112: You're relying on addTrack and removeTrack existing. This is only guaranteed if chromeShim.shimAddTrackRemoveTrack has been called prior. Can you make this assumption explicit with a comment and throwing an exception that shim failed if that isn't the case?

henbos (approved):
LGTM % comments

src/js/chrome/chrome_shim.js
fippo: added a comment. I don't think we need to explicitly throw but should test once you unflag the native variants

PR #613, opened 2017-07-05
Use simple object assignment instead of Object.assign 

xdumaine:
which crashes shimming in phantom

**Description**

This PR switches the initial options assignment from `Object.assign` to simple property assignment.

**Purpose**

Prior to version 4.x, the script could be loaded in phantom.js/IE/Safari without crashing. As of 4.x, the use of `Object.assign` (which does not exist in these browsers) causes an exception to be thrown immediately upon loading adapter.js.

### DIFF
--- a/src/js/adapter_factory.js
+++ b/src/js/adapter_factory.js
@@ -13,12 +13,18 @@
 module.exports = function(dependencies, opts) {
   var window = dependencies && dependencies.window;
 
-  var options = Object.assign({
+  var options = {
     shimChrome: true,
     shimFirefox: true,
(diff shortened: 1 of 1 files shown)


fippo (approved):

fippo:
ugh!

KaptenJansson:
LGTM, maybe worth adding phantom into the mix of browsers since it's still used heavily for automation?

fippo:
at least some kind of check that the output can be parsed and things don't throw up. I do hope folks replace  phantomjs with chrome headless though :-)

PR #617, opened 2017-07-10
chrome_shim: polyfill track event in SRD

### LINKED ISSUE #575: [Chrome] Please don't force users to just set pc.ontrack but also allow pc.addEventListener('track')
The Chrome shim provides [pc.ontrack](https://github.com/webrtc/adapter/blob/master/src/js/chrome/chrome_shim.js#L19), but if the app uses a proper `pc.addEventListener('track', fn)` it's ignored (because the shim just react on `pc.ontrack` being set).

Well, I do know that many people just rely on `pc.onxxxx` setters, but probably most of them don't even know what a [EventHandler](https://html.spec.whatwg.org/multipage/webappapis.html#eventhandler) is.

So my proposal is that the Chrome PeerConnection shim calls to `dispatchEvent(event)` (as in lines 46 and 61) without requiring that the user sets `pc.ontrack` before.

fippo:
fixes #575, this turned out to be easier

@jan-ivar can you please take a look?

### DIFF
--- a/src/js/chrome/chrome_shim.js
+++ b/src/js/chrome/chrome_shim.js
@@ -24,19 +24,24 @@ var chromeShim = {
           return this._ontrack;
         },
         set: function(f) {
-          var self = this;
           if (this._ontrack) {
             this.removeEventListener('track', this._ontrack);
-            this.removeEventListener('addstream', this._ontrackpoly);
           }
           this.addEventListener('track', this._ontrack = f);
-          this.addEventListener('addstream', this._ontrackpoly = function(e) {
+        }
+      });
+      var origSetRemoteDescription =
+          window.RTCPeerConnection.prototype.setRemoteDescription;
+      window.RTCPeerConnection.prototype.setRemoteDescription = function() {
+        var pc = this;
+        if (!pc._ontrackpoly) {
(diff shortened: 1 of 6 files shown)


fippo:
I _think_ the addstream event listeners need to be removed from the ontrack setter, right? (done now)

src/js/chrome/chrome_shim.js
```diff
@@ -64,9 +69,11 @@ var chromeShim = {
               event.streams = [e.stream];
               this.dispatchEvent(event);
             }.bind(this));
-          }.bind(this));
+          }.bind(this);
+          this.addEventListener('addstream', this._ontrackpoly);
```
jan-ivar: Nit: since you changed `self` to `pc` you might as well change the last two occurrences of `this` to `pc` here as well and drop the redundant `bind(this)` stuff.

jan-ivar (commented):
Lgtm. This never removes the _ontrackpoly handler but I guess that's ok.

src/js/chrome/chrome_shim.js
fippo: done. I prefer the "pc = this" style ;-)

KaptenJansson:
LGTM

PR #620, opened 2017-07-11
Add BROWSER, BVER and CI to preversion step. Also remove Faucet.

KaptenJansson:
**Description**
- Add BROWSER, BVER and CI to preversion step.
- Remove faucet
**Purpose**
Make sure the npm version step can be run as before without having to provide env vars.
Also faucet does not provide any value at the moment hence why I removed it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "The WebRTC project authors (https://www.webrtc.org/)"
   ],
   "scripts": {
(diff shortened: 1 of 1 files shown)


fippo:
we would need to move faucet to the selenium test step. But that is going away soon

fippo (approved):

PR #625, opened 2017-07-13
chrome_shim: reject removeTrack(track)

### LINKED ISSUE #624: removeTrack fails with 'Sender was not created by this connection.'
Chrome: Version 59.0.3071.115 (Official Build) (64-bit)
adapter.js: 4.2.0

**Description**
Calling removeTrack causes the exception 'Sender was not created by this connection.'.
The code that's failing is: 
      var isLocal = sender._pc === pc;
      if (!isLocal) {
        throw new DOMException('Sender was not created by this connection.',
            'InvalidAccessError');
      }

sender._pc is undefined. The only place _pc appears to be set is in the routine shimGetSendersWithDtmf.

**Steps to reproduce**
Call removeTrack

**Expected results**
Track should be removed without throwing an exception

**Actual results**
An exception is thrown.

fippo:
rejects removeTrack(track) explicitly, fixes #624.

### DIFF
--- a/src/js/chrome/chrome_shim.js
+++ b/src/js/chrome/chrome_shim.js
@@ -338,6 +338,12 @@ var chromeShim = {
           'The RTCPeerConnection\'s signalingState is \'closed\'.',
           'InvalidStateError');
       }
+      // We can not yet check for sender instanceof RTCRtpSender
+      // since we shim RTPSender. So we check if sender._pc is set.
+      if (!sender._pc) {
+        throw new DOMException('Argument 1 of RTCPeerConnection.removeTrack ' +
+            'does not implement interface RTCRtpSender.', 'TypeError');
+      }
       var isLocal = sender._pc === pc;
       if (!isLocal) {
         throw new DOMException('Sender was not created by this connection.',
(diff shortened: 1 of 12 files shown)


test/e2e/expectations/MicrosoftEdge
```diff
@@ -35,6 +35,7 @@ PASS track event the event has a track
 PASS track event the event has a set of streams
 PASS track event the event has a receiver that is contained in the set of receivers
 ERR removeTrack allows removeTrack twice timeout
+ERR removeTrack throws an exception if the argument is a track, not a sender AssertionError
```
fippo, line 38: I learned to cheat by using vs code :-)

src/js/chrome/chrome_shim.js
```diff
@@ -338,6 +338,12 @@ var chromeShim = {
           'The RTCPeerConnection\'s signalingState is \'closed\'.',
           'InvalidStateError');
       }
+      // We can not yet check for sender instanceOf RTCRtpSender
```
jan-ivar: instanceof

jan-ivar (approved):

KaptenJansson (approved):

PR #628, opened 2017-07-18
karma: make developer life easier

fippo:
by setting CHROME_BIN and FIREFOX_BIN when not set.
Also allows changing the stability reporters update setting
via an environment variable.


KaptenJansson (approved):
Needed those variables after all :)

PR #630, opened 2017-07-19
chrome: make ONN async

fippo:
makes the negotiationneeded fired from addTrack async

@henbos can you take a look? Not sure what a good test is...

### DIFF
--- a/src/js/chrome/chrome_shim.js
+++ b/src/js/chrome/chrome_shim.js
@@ -319,7 +319,11 @@ var chromeShim = {
         // Note: we rely on the high-level addTrack/dtmf shim to
         // create the sender with a dtmf sender.
         oldStream.addTrack(track);
-        pc.dispatchEvent(new Event('negotiationneeded'));
+
+        // Trigger ONN async.
+        Promise.resolve().then(function() {
+          pc.dispatchEvent(new Event('negotiationneeded'));
+        });
       } else {
         var newStream = new window.MediaStream([track]);
         pc._streams[stream.id] = newStream;


src/js/chrome/chrome_shim.js
```diff
@@ -319,7 +319,9 @@ var chromeShim = {
         // Note: we rely on the high-level addTrack/dtmf shim to
         // create the sender with a dtmf sender.
         oldStream.addTrack(track);
-        pc.dispatchEvent(new Event('negotiationneeded'));
+        window.setTimeout(function() {
+          pc.dispatchEvent(new Event('negotiationneeded'));
+        }, 0);
```
henbos: Another way to do this is to do `Promise.resolve().then(function() { ... });`.
Not sure which is preferred though, in my testing I found that setTimeout fired after the promise even if the promise was called first (but async in both cases). I guess we should either always do setTimeout or always do Promise.resolve() for async actions as to not get surprising order of events.

In this file I see no uses of setTimeout and one usage of "return Promise.resolve();".
I think we should use `Promise.resolve()` everywhere, or else we might get weird results.

henbos (approved):
LGTM but do Promise.resolve() instead of setTimeout()

src/js/chrome/chrome_shim.js
fippo: those are different timers. I think @jan-ivar explained the difference to me a while ago... not sure which is better, since addTrack does not return a promise i think it doesn't matter.

jan-ivar: @fippo It's ... complicated. TL;DR: [Apparently](http://logs.glob.uno/?c=mozilla%23content&s=19+Jul+2017&e=22+Jul+2017#c453594) there's no official way to [queue a task](http://w3c.github.io/html/webappapis.html#queuing) from JS.

I was discussing this two days ago with @docfaraday, who's moving our ONN implementation from c++ to JS.

When he used `setTimeout` in gecko, the event fired way too late. You could literally do `addTrack` followed by `await setLocalDescription(await createOffer())` followed by `await setLocalDescription({type: "rollback"})` before the ONN event fired. Not good.

On the other hand `Promise.resolve().then(() => { /* here */ })` is technically [too soon](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_promises#Timing), and might race with other things, at least in theory. We ended up using an [internal method](http://searchfox.org/mozilla-central/rev/3a3af33f513071ea829debdfbc628caebcdf6996/xpcom/threads/nsIThreadManager.idl#88-95).

There are [hacks](https://stackoverflow.com/q/27788175/918910) out there, but when I tried [the postMessage hack](https://jsfiddle.net/jib1/j2xfweLo/) it wasn't reliable in FF. Might be good enough for Chrome though.

fippo: Will go for Promise.resolve(), its better than setTimeout at least. Thanks!

KaptenJansson:
Merging, test failures are unrelated.

PR #631, opened 2017-07-19
test: fix karma config

fippo:
cwd() is a function of the process, not the operating system


KaptenJansson (approved):
My bad, I should have caught this.

PR #637, opened 2017-07-24
chrome: munge sdp to keep stream ids

### LINKED ISSUE #634: Chrome: MSID does not match MediaStream passed to addStream or addTrack

Google Chrome 59, 60

**Description**

adapter.js appears to be changing MSID values when `addStream` or `addTrack` is used.

**Steps to reproduce**

* [JSFiddle using `addStream` without adapter.js](https://jsfiddle.net/7sh3oc28/) (Chrome, Firefox)
* [JSFiddle using `addStream` with adapter.js (latest)](https://jsfiddle.net/cs22nvjL/) (Chrome, Firefox)
* [JSFiddle using `addTrack` without adapter.js](https://jsfiddle.net/3e2t3wzv/1/) (Firefox)
* [JSFiddle using `addTrack` with adapter.js (latest)](https://jsfiddle.net/1nbr1u7m/1/) (Chrome, Firefox)

**Expected results**

I get these expected results when adapter.js is not used in both Chrome and Firefox.

```
stream.id $STREAM_ID
track.id $TRACK_ID
$OFFER
msid $STREAM_ID $TRACK_ID
stream.id === msid[0] true
track.id === msid[1] true
```

**Actual results**

I get these results when adapter.js is used in Chrome.

```
stream.id $STREAM_ID
track.id $TRACK_ID
$OFFER
msid $RANDOM_ID $TRACK_ID
stream.id === msid[0] false
track.id === msid[1] true
```

fippo:
munges the SDP in createOffer, createAnswer and setLocalDescription as well
as the pc.localDescription accessor in order to retain the stream ids that
are modifed by the addTrack shim.

Fixes #634 

@henbos can you take a look? I tried to mangle the getStats result too but that didn't work :-/

### DIFF
--- a/src/js/chrome/chrome_shim.js
+++ b/src/js/chrome/chrome_shim.js
@@ -337,6 +337,88 @@ var chromeShim = {
       });
     };
 
+    // replace the internal stream id with the external one and
+    // vice versa.
+    function replaceInternalStreamId(pc, description) {
+      var sdp = description.sdp;
+      Object.keys(pc._reverseStreams || []).forEach(function(internalId) {
+        var externalStream = pc._reverseStreams[internalId];
+        var internalStream = pc._streams[externalStream.id];
+        sdp = sdp.replace(new RegExp(internalStream.id, 'g'),
+            externalStream.id);
+      });
+      return new RTCSessionDescription({
+        type: description.type,
+        sdp: sdp
+      });
+    }
+    function replaceExternalStreamId(pc, description) {
+      var sdp = description.sdp;
+      Object.keys(pc._reverseStreams || []).forEach(function(internalId) {
+        var externalStream = pc._reverseStreams[internalId];
+        var internalStream = pc._streams[externalStream.id];
+        sdp = sdp.replace(new RegExp(externalStream.id, 'g'),
+            internalStream.id);
+      });
+      return new RTCSessionDescription({
+        type: description.type,
+        sdp: sdp
+      });
+    }
+    ['createOffer', 'createAnswer'].forEach(function(method) {
+      var nativeMethod = window.RTCPeerConnection.prototype[method];
+      window.RTCPeerConnection.prototype[method] = function() {
+        var pc = this;
+        var args = arguments;
+        var isLegacyCall = arguments.length &&
+            typeof arguments[0] === 'function';
+        if (isLegacyCall) {
+          return nativeMethod.apply(pc, [
+            function(description) {
+              var desc = replaceInternalStreamId(pc, description);
+              args[0].apply(null, [desc]);
+            },
+            function(err) {
+              if (args[1]) {
+                args[1].apply(null, err);
+              }
+            }, arguments[2]
+          ]);
+        }
+        return nativeMethod.apply(pc, arguments)
+        .then(function(description) {
+          return replaceInternalStreamId(pc, description);
+        });
+      };
+    });
+
+    var origSetLocalDescription =
+        window.RTCPeerConnection.prototype.setLocalDescription;
+    window.RTCPeerConnection.prototype.setLocalDescription = function() {
+      var pc = this;
+      if (!arguments.length || !arguments[0].type) {
+        return origSetLocalDescription.apply(pc, arguments);
+      }
+      arguments[0] = replaceExternalStreamId(pc, arguments[0]);
+      return origSetLocalDescription.apply(pc, arguments);
+    };
+
+    // TODO: mangle getStats: https://w3c.github.io/webrtc-stats/#dom-rtcmediastreamstats-streamidentifier
+
+    var origLocalDescription = Object.getOwnPropertyDescriptor(
+        window.RTCPeerConnection.prototype, 'localDescription');
+    Object.defineProperty(window.RTCPeerConnection.prototype,
+        'localDescription', {
+          get: function() {
+            var pc = this;
+            var description = origLocalDescription.get.apply(this);
+            if (description.type === '') {
+              return description;
+            }
+            return replaceInternalStreamId(pc, description);
+          }
+        });
+
     window.RTCPeerConnection.prototype.removeTrack = function(sender) {
       var pc = this;
       if (pc.signalingState === 'closed') {
--- a/test/e2e/msid.js
+++ b/test/e2e/msid.js
@@ -1,0 +1,56 @@
+/*
+ *  Copyright (c) 2017 The WebRTC project authors. All Rights Reserved.
+ *
+ *  Use of this source code is governed by a BSD-style license
+ *  that can be found in the LICENSE file in the root of the source
+ *  tree.
+ */
+ /* eslint-env node */
+'use strict';
+
+describe('MSID', () => {
+  let pc1;
+  let pc2;
+
+  function negotiate(pc, otherPc) {
+    return pc.createOffer()
+    .then(function(offer) {
+      return pc.setLocalDescription(offer);
+    }).then(function() {
+      return otherPc.setRemoteDescription(pc.localDescription);
+    }).then(function() {
+      return otherPc.createAnswer();
+    }).then(function(answer) {
+      return otherPc.setLocalDescription(answer);
+    }).then(function() {
+      return pc.setRemoteDescription(otherPc.localDescription);
+    });
+  }
+
+  beforeEach(() => {
+    pc1 = new RTCPeerConnection(null);
+    pc2 = new RTCPeerConnection(null);
+
+    pc1.onicecandidate = event => pc2.addIceCandidate(event.candidate);
+    pc2.onicecandidate = event => pc1.addIceCandidate(event.candidate);
+  });
+  afterEach(() => {
+    pc1.close();
+    pc2.close();
+  });
+
+  it('signals track and stream ids', (done) => {
+    let localStream;
+    pc2.ontrack = (e) => {
+      expect(e.track.id).to.equal(localStream.getTracks()[0].id);
+      expect(e.streams[0].id).to.equal(localStream.id);
+      done();
+    };
+    navigator.mediaDevices.getUserMedia({video: true})
+    .then((stream) => {
+      localStream = stream;
+      pc1.addTrack(stream.getTracks()[0], stream);
+      return negotiate(pc1, pc2);
+    });
+  });
+});
--- a/test/e2e/expectations/chrome-beta
+++ b/test/e2e/expectations/chrome-beta
@@ -44,6 +44,7 @@ PASS navigator.mediaDevices enumerateDevices returns some audioinput devices
 PASS navigator.mediaDevices enumerateDevices returns some videoinput devices
 PASS navigator.mediaDevices enumerateDevices returns some audiooutput devices
 PASS MediaStream window.MediaStream exists
+PASS MSID signals track and stream ids
 PASS track event RTCPeerConnection.prototype.ontrack exists
 PASS track event is called by setRemoteDescription during renegotiation
 PASS track event is called by setRemoteDescription track event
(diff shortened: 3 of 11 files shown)


henbos:
Shimming getStats didn't work? Isn't the only thing that needs to be updated RTCMediaStreamStats.streamIdentifiers and RTCMediaStreamTrackStats.trackIdentifier?

fippo:
tracks are kept the same ;-)
I had a patch that did this:
```
-    // TODO: mangle getStats: https://w3c.github.io/webrtc-stats/#dom-rtcmediastreamstats-streamidentifier
+    var origGetStats = window.RTCPeerConnection.prototype.getStats;
+    window.RTCPeerConnection.prototype.getStats = function() {
+      var pc = this;
+      return origGetStats.apply(pc, arguments)
+      .then(function(result) {
+        result.forEach(function(stat) {
+          if (stat.type === 'stream' && pc._reverseStreams) {
+            var externalStream = pc._reverseStreams[stat.streamIdentifier];
+            if (externalStream) {
+              stat.streamIdentifier = externalStream.id;
+            }
+          }
+        });
+        return result;
+      });
+    };
```
It seemed to work but magically in the next function the streamIdentifier was the old string again.

henbos:
Maybe another shimming of getStats overwrites that shimming?

fippo:
nah, I iterated again with another forEach before the "return result". Check here:
    https://jsfiddle.net/jfoe1L7g/
Really really odd...

henbos:
Oh, the stats objects are dictionaries, so they are returned as copies. Once returned you can modify them, but because the RTCStatsReport is a Maplike interface, every time you retrieve an element you get a copy of the original. In order to mangle this you need to mangle `result`. I think `Map` is a `Maplike`? If so you can just add all the modified members to a `new Map()` and return that.

henbos:
https://jsfiddle.net/s9goLhbp/1/

markandrus:
Any ETA/pre-reqs for landing this? The only workaround I could see was pulling copies of `addStream` and `removeStream` off a "fresh" RTCPeerConnection `prototype` from an &lt;iframe&gt;, but I really don't want to land such a workaround in my code...

KaptenJansson:
@henbos could you take a look at the latest commit?

fippo:
track-based test failures in beta with experimental flag is real now... @henbos we need to have a fig.. err friendly chat ;-)

fippo:
@jan-ivar can you take a look at https://github.com/webrtc/adapter/pull/637/commits/466e3f651d306dbb59734421887eba99920f2396 please? I noticed that createOffer(cb, err, opts) resolved the promise with no arguments which lead to a very very confusing error.

jan-ivar:
@fippo I'd say leave legacy code alone, so we A) don't break it and B) leave the pearls for migraters.

The legacy callback methods return `Promise<void>` [per spec](https://w3c.github.io/webrtc-pc/#legacy-interface-extensions) due to webidl overloading limitations.

src/js/chrome/chrome_shim.js
```diff
@@ -335,6 +335,82 @@ var chromeShim = {
...
+      window.RTCPeerConnection.prototype[method] = function() {
+        var pc = this;
+        var opts;
+        var args = arguments;
+        if (arguments.length === 1 && typeof arguments[0] === 'object') {
+          opts = arguments[0];
+        } else if (arguments.length === 3 && typeof arguments[2] === 'object') {
+          opts = arguments[2];
+        }
+        return nativeMethod.apply(pc, [opts])
```
jan-ivar: So this diverts calls of the legacy callback API to the promise API, which I don't like, as it will suppress native deprecation warnings (like in Firefox), and may also skew any telemetry of how many still call the legacy API. How about:
```js
var p, isLegacyCall = arguments.length && typeof arguments[0] === 'function';
if (isLegacyCall) {
  p = new Promise((r, e) => nativeMethod.apply(pc, [r, e, arguments[2]]));
} else {
  p = nativeMethod.apply(pc, arguments);
}
```
?

src/js/chrome/chrome_shim.js
```diff
@@ -335,6 +335,82 @@ var chromeShim = {
...
+          var desc = new window.RTCSessionDescription({
+            type: description.type,
+            sdp: sdp
+          });
+          if (args.length > 0 && typeof args[0] === 'function') {
+            args[0].apply(null, [desc]);
+            return Promise.resolve();
+          }
+          return desc;
+        }, function(err) {
```
jan-ivar: You're not catching any potential programming errors in this code. I'd change `.then(a, b)` to `.then(a).catch(b)`.

src/js/chrome/chrome_shim.js
```diff
@@ -335,6 +335,82 @@ var chromeShim = {
...
+            sdp = sdp.replace(new RegExp(internalStream.id, 'g'),
+                externalStream.id);
+          });
+          var desc = new window.RTCSessionDescription({
+            type: description.type,
+            sdp: sdp
+          });
+          if (args.length > 0 && typeof args[0] === 'function') {
+            args[0].apply(null, [desc]);
+            return Promise.resolve();
```
jan-ivar: redundant. Just `return;` will do

src/js/chrome/chrome_shim.js
```diff
@@ -335,6 +335,82 @@ var chromeShim = {
...
+          }
+          return desc;
+        }, function(err) {
+          if (args.length > 1 && typeof args[1] === 'function') {
+            args[1].apply(null, [err]);
+          }
```
jan-ivar: `} else { throw err; }` otherwise you're removing errors for promise-users.

jan-ivar (changes requested):

src/js/chrome/chrome_shim.js
```diff
@@ -337,34 +337,54 @@ var chromeShim = {
...
           var sdp = description.sdp;
-          Object.keys(pc._reverseStreams || []).forEach(function(internalId) {
-            var externalStream = pc._reverseStreams[internalId];
-            var internalStream = pc._streams[externalStream.id];
-            sdp = sdp.replace(new RegExp(internalStream.id, 'g'),
-                externalStream.id);
-          });
           var desc = new window.RTCSessionDescription({
             type: description.type,
             sdp: sdp
```
jan-ivar: I think you meant to call `replaceInternalStreamId()` here.

src/js/chrome/chrome_shim.js
```diff
@@ -378,30 +398,17 @@ var chromeShim = {
...
-      .then(function() {
-        if (args.length > 1 && typeof args[1] === 'function') {
-          args[1].apply(null, []);
-        }
-      })
-      .catch(function(err) {
-        if (args.length > 2 && typeof args[2] === 'function') {
-          args[2].apply(null, [err]);
-        }
+        sdp: replaceExternalStreamId(pc, arguments[0].sdp)
```
jan-ivar: We might want to wrap this in
```js
if (arguments[0].type || typeof arguments[0].sdp === 'string')
```
to protect this polyfill from bogus input without having to mimic native error checking.

jan-ivar (changes requested):

src/js/chrome/chrome_shim.js
fippo: no, we take the external stream id in the SDP (which comes from createOffer/createAnswer usually) and replace it with the internal one before SLD. I agree that the naming is suboptimal, but its two different operations

src/js/chrome/chrome_shim.js
fippo: 👍 -- I didn't check for .sdp because rollbacks

KaptenJansson:
@henbos could you take another look?

src/js/chrome/chrome_shim.js
```diff
@@ -335,6 +335,97 @@ var chromeShim = {
...
+      var desc = new window.RTCSessionDescription({
+        type: arguments[0].type,
+        sdp: replaceExternalStreamId(pc, arguments[0].sdp)
+      });
+      var isLegacyCall = arguments.length > 1;
+      if (isLegacyCall) {
+        return origSetLocalDescription.apply(pc,
+            [desc, arguments[1], arguments[2]]);
+      }
+      return origSetLocalDescription.apply(pc, [desc]);
```
henbos: Can you instead of "var desc = ..." do "arguments[0] = ..." and then do "origSetLocalDescription.apply(pc, arguments)" and get rid of the isLegacyCall check?

src/js/chrome/chrome_shim.js
```diff
@@ -335,6 +335,97 @@ var chromeShim = {
...
+        sdp = sdp.replace(new RegExp(externalStream.id, 'g'),
+            internalStream.id);
+      });
+      return sdp;
+    }
+    ['createOffer', 'createAnswer'].forEach(function(method) {
+      var nativeMethod = window.RTCPeerConnection.prototype[method];
+      window.RTCPeerConnection.prototype[method] = function() {
+        var pc = this;
+        var args = arguments;
```
henbos, line 372: No need for "args", simply use "arguments".

src/js/chrome/chrome_shim.js
```diff
@@ -335,6 +335,97 @@ var chromeShim = {
...
+    }
+    function replaceExternalStreamId(pc, sdp) {
+      Object.keys(pc._reverseStreams || []).forEach(function(internalId) {
+        var externalStream = pc._reverseStreams[internalId];
+        var internalStream = pc._streams[externalStream.id];
+        sdp = sdp.replace(new RegExp(externalStream.id, 'g'),
+            internalStream.id);
+      });
+      return sdp;
+    }
```
henbos: Every call to replaceInternal/ExternalStreamId is in a "new window.RTCSessionDescription". If you make these functions do that bit too you reduce a few repeated lines of code in the rest of the file.

src/js/chrome/chrome_shim.js
```diff
@@ -335,6 +335,97 @@ var chromeShim = {
...
+      });
+      var isLegacyCall = arguments.length > 1;
+      if (isLegacyCall) {
+        return origSetLocalDescription.apply(pc,
+            [desc, arguments[1], arguments[2]]);
+      }
+      return origSetLocalDescription.apply(pc, [desc]);
+    };
+
+    // TODO: mangle getStats: https://w3c.github.io/webrtc-stats/#dom-rtcmediastreamstats-streamidentifier
```
henbos: Do you want to do this separately? https://jsfiddle.net/s9goLhbp/1/

test/e2e/msid.js
```diff
@@ -0,0 +1,56 @@
...
+      expect(e.streams[0].id).to.equal(localStream.id);
+      done();
+    };
+    navigator.mediaDevices.getUserMedia({video: true})
+    .then((stream) => {
+      localStream = stream;
+      pc1.addTrack(stream.getTracks()[0], stream);
+      return negotiate(pc1, pc2);
+    });
+  });
```
henbos, line 55: Add a test that verifies "localDescription" is what you expect.

henbos (changes requested):
Looking good.

src/js/chrome/chrome_shim.js
fippo: no, args is used in the legacy case. Can't use arguments there since that is a different function scope

src/js/chrome/chrome_shim.js
fippo: sneaky...

src/js/chrome/chrome_shim.js
fippo: yeah, different PR

test/e2e/msid.js
fippo: A test for accessing .localDescription is indeed missing. Can I add this later? I promise ;-)

src/js/chrome/chrome_shim.js
henbos: Acknowledged.

test/e2e/msid.js
henbos: If so can you create a bug or add a TODO?

henbos (approved):
LGTM

KaptenJansson (approved):

PR #641, opened 2017-08-02
chrome: workaround for #640

### LINKED ISSUE #640: Chrome: receiver.track can be null, causing issue with "track" event shim

Chrome Canary 62

**Description**

In Chrome, RTCRtpReciever's `track` can be `null`. This causes an error in the "track" event shim.

1. Offer with `offerToReceiveAudio: true`
2. Answer with `a=msid:stream track1`
3. Offer with `offerToReceiveAudio: true`
4. Answer, replacing `a=msid:stream track1` with `a=msid:stream track2`

**Steps to reproduce**

See [JSFiddle](https://jsfiddle.net/kbg8149n/)

**Expected results**

Two "track" events for `track1` and `track2` (`track1` should end)

**Actual results**

```
adapter-latest.js:827 Uncaught TypeError: Cannot read property 'id' of null
    at https://webrtc.github.io/adapter/adapter-latest.js:827:34
    at Array.find (native)
    at MediaStream.<anonymous> (https://webrtc.github.io/adapter/adapter-latest.js:826:46)
```

fippo:
fixes #640 (kind of)

upstream bug: https://bugs.chromium.org/p/chromium/issues/detail?id=751785

@KaptenJansson lets see how good the "patch release" instructions are ;-)

### DIFF
--- a/src/js/chrome/chrome_shim.js
+++ b/src/js/chrome/chrome_shim.js
@@ -42,7 +42,7 @@ var chromeShim = {
               var receiver;
               if (window.RTCPeerConnection.prototype.getReceivers) {
                 receiver = pc.getReceivers().find(function(r) {
-                  return r.track.id === te.track.id;
+                  return r.track && r.track.id === te.track.id;
                 });
               } else {
                 receiver = {track: te.track};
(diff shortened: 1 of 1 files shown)


KaptenJansson:
LGTM

fippo:
the implication is that we dont get a receiver in this case. But that will break in user code, not adapter

KaptenJansson:
Well I think this was general code health fix (i.e. make sure the object exists before accessing it's properties).

fippo:
the IDL said it always exists ,-)

PR #642, opened 2017-08-03
travis: allow firefox beta to fail

fippo:
firefox 55+ currently times out when trying to fetch file:// urls
which causes the selenium tests to fail. Allow travis failure.


fippo:
https://bugzilla.mozilla.org/show_bug.cgi?id=1332122 -- will be fixed in FF56

KaptenJansson:
LGTM

PR #644, opened 2017-08-09
common: make URL.createObjectURL shim depend on native support

### LINKED ISSUE #569: Conflict with external JS librairies
Tested on Chrome with latest adapter.js

**Description**
When deploying adapter.js on customer site, it happen that there is conflicts between existing JS librairies on the web site and adapter.js
I have two example with requireJs and browser.js

**Steps to reproduce**
I have prepared some simple page to show the issues :

Issue 1 :
http://www.apizee.com/Demo/testAdapter/testAdapterRequireJS.html
Error : Uncaught Error: Mismatched anonymous define() module: function (){var define,module,exports;return (function e(t,n,r){function s(o,u){if(!n[o]){if(!t[o]){var a=typeof require=="function" [...]

Issue 2 : http://www.apizee.com/Demo/testAdapter/testAdapterBrowser.html
Error : Uncaught TypeError: Cannot read property 'bind' of undefined
    at Object.shimCreateObjectURL (adapter-latest.js:3445)
    at adapter-latest.js:657 [...]

**Expected results**
Would expect adapter.js to have no conflict / impact with other librairies

**Actual results**
Conflict cause that adapter.js script cannot be executed

fippo:
fixes #569

### DIFF
--- a/src/js/utils.js
+++ b/src/js/utils.js
@@ -138,7 +138,8 @@ var utils = {
     var URL = window && window.URL;
 
     if (!(typeof window === 'object' && window.HTMLMediaElement &&
-          'srcObject' in window.HTMLMediaElement.prototype)) {
+          'srcObject' in window.HTMLMediaElement.prototype &&
+        URL.createObjectURL && URL.revokeObjectURL)) {
       // Only shim CreateObjectURL using srcObject if srcObject exists.
       return undefined;
     }


src/js/utils.js
```diff
@@ -145,6 +145,9 @@ var utils = {
 
     var nativeCreateObjectURL = URL.createObjectURL.bind(URL);
     var nativeRevokeObjectURL = URL.revokeObjectURL.bind(URL);
```
jan-ivar: This still calls `undefined.bind()`. Better to test for 
```js
if (!URL.createObjectURL || !URL.revokeObjectURL)
```

jan-ivar (changes requested):

src/js/utils.js
fippo: moved up a bit, thanks for suggesting this!

KaptenJansson (approved):

PR #646, opened 2017-08-14
chrome: fix GUM error callback

fippo:
only calls the chrome error callback if it exists

@jan-ivar can you take a quick look?

### DIFF
--- a/src/js/chrome/getusermedia.js
+++ b/src/js/chrome/getusermedia.js
@@ -151,7 +151,9 @@ module.exports = function(window) {
   var getUserMedia_ = function(constraints, onSuccess, onError) {
     shimConstraints_(constraints, function(c) {
       navigator.webkitGetUserMedia(c, onSuccess, function(e) {
-        onError(shimError_(e));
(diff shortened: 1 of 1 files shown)


jan-ivar (approved):
lgtm

src/js/chrome/getusermedia.js
```diff
@@ -151,7 +151,9 @@ module.exports = function(window) {
   var getUserMedia_ = function(constraints, onSuccess, onError) {
     shimConstraints_(constraints, function(c) {
       navigator.webkitGetUserMedia(c, onSuccess, function(e) {
-        onError(shimError_(e));
+        if (onError) {
```
tgabi333, line 154: `typeof onError === 'function'` wouldn't be a better solution?

fippo: i think throwing up on garbage input is better than silently ignoring ;-)

KaptenJansson (approved):

KaptenJansson (approved):

PR #647, opened 2017-08-15
Shim event.transceiver.

jan-ivar:
**Description**
Shim `event.transceiver` on older browsers (i.e. all of them at this point).

**Purpose**
Let's us consider removing redundant `event.receiver`. in favor of `event.transceiver.receiver`.

### DIFF
--- a/src/js/adapter_factory.js
+++ b/src/js/adapter_factory.js
@@ -116,6 +116,7 @@ module.exports = function(dependencies, opts) {
       safariShim.shimCallbacksAPI(window);
       safariShim.shimLocalStreamsAPI(window);
       safariShim.shimRemoteStreamsAPI(window);
+      safariShim.shimTrackEventTransceiver(window);
       safariShim.shimGetUserMedia(window);
       break;
     default:
--- a/src/js/chrome/chrome_shim.js
+++ b/src/js/chrome/chrome_shim.js
@@ -51,6 +51,7 @@ var chromeShim = {
               var event = new Event('track');
               event.track = te.track;
               event.receiver = receiver;
+              event.transceiver = {receiver: receiver};
               event.streams = [e.stream];
               pc.dispatchEvent(event);
             });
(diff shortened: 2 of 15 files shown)


src/js/edge/edge_shim.js
```diff
@@ -66,8 +66,26 @@ module.exports = {
...
+    var BaseRTCPeerConnection =
         shimRTCPeerConnection(window, browserDetails.version);
+
+    // Shim event.transceiver member ontop of base shim
+    window.RTCPeerConnection = function(config) {
+      var pc = new BaseRTCPeerConnection(config);
+
+      var baseDispatchEvent = pc.dispatchEvent;
+      pc.dispatchEvent = function(event) {
+        console.log('JIB HERE HERE!');
```
fippo: good job ;-)

src/js/edge/edge_shim.js
```diff
@@ -66,8 +66,26 @@ module.exports = {
       });
```
fippo: I think i'd rather like to do a native shim for this one... will see when I find time!

jan-ivar: Should I leave it here until then (it'd become a no-op at that point)?

jan-ivar:
Note that this shim might confuse JS expecting to find the full transceiver if this member is present. I don't know that anyone is yet, and since adoption of adapter.js is voluntary, I'm hoping this won't surprise anyone.

jan-ivar:
The travis errors seem unrelated. @fippo any thoughts on how to proceed?

fippo:
https://github.com/webrtc/adapter/pull/647/files#diff-93bbbc7918507cda99384ea8b2ce2fd4R54
-- this line contains an extra "s" at the end. The other failures are because of Chrome or Firefox/selenium issues and can be ignored

fippo:
tests failures look sad but ok, squash once more and we're good.

KaptenJansson (approved):

PR #653, opened 2017-08-28
test: remove selenium tests

fippo:
removes the selenium tests which have largely been superseeded by the karma ones.

I don't see much value in the remaining tests and they are currently the biggest issue getting things to go green (mainly because of https://bugzilla.mozilla.org/show_bug.cgi?id=1332122)

Still need to update test/README.md.

selenium will stay with us for the samples at least ;-)


KaptenJansson (approved):

fippo:
@jan-ivar do you see much value in the tests that were removed?

jan-ivar:
@fippo Lgtm. I suppose one area where "impossible constraints" still mean the same on all browsers would be `{width: {min: 9999}}`, but I guess by unit-testing constraints conversion with before/after tests, we site-step relying on browser behavior which tends to be hard to test in these environments.

KaptenJansson:
I'm thinking it's probably good to update the README in this PR as well, WDYT?

KaptenJansson (approved):
Thanks for doing this!

PR #654, opened 2017-08-28
test: update chrome expectations

fippo:
updates the chrome expecations to match the addTrack behaviour which is behind a flag
in chrome beta and unstable

@henbos 😭 😢 😿


henbos (commented):
I would expect most of these to pass, is this the fault of adapter shimming when it shouldn't or chrome being broken?

fippo:
I suspected chrome being broken. Easiest way to reproduce is to extract them into a fiddle [using this as a template](https://jsfiddle.net/gurdiga/c5z5w0zy/) and then run without adapter

See [here](https://jsfiddle.net/ypqLmpp8/3/) for the first failure

KaptenJansson:
@fippo what about the FF failures?

fippo:
they were due to file:// selenium timeouts. We're pretty green, I fixed a copy-paste mistake in FF nightly expectations

KaptenJansson (approved):

PR #658, opened 2017-08-30
Bump version to 5.0.1

KaptenJansson:
**Description**

**Purpose**


fippo (approved):

fippo:
uh. Can you remove the package-lock please? With this the ortc shim would be locked even for patch versions

PR #659, opened 2017-08-30
test: use headless firefox

fippo:
uses headless firefox which is described in
    https://developer.mozilla.org/en-US/Firefox/Headless_mode

We still need xvfb for the time being due to firefox ESR


KaptenJansson (approved):

KaptenJansson:
Great stuff, love headless!

fippo:
cc'ing @jan-ivar who might be surprised otherwise ;-)

PR #663, opened 2017-09-01
package: remove package-lock, prevent new one from being created

fippo:
with a package lockfile dependencies would be pinned which is undesirable for a library.


fippo:
@MirkoBonadei @KaptenJansson PTAL (and i'd like a new patch release after that, I fixed a nasty bug in the peerconnection shim)

KaptenJansson:
I notices this new NPM wizardry last week and was wondering what it was, now I know :)

KaptenJansson (approved):

PR #656, opened 2017-08-30
Download browser in npm test step, chrome stable if BROWSER and/or BVER are not specified

KaptenJansson:
**Description**
Download browser in npm test step
Chrome stable if BROWSER and/or BVER are not specified

**Purpose**
Download browser automatically locally and on travis.

@fippo PTAL

PR #665, opened 2017-09-01
Move grunt-shell to devDependencies

ratson:
**Description**
`grunt-shell` should be a `devDependencies`

**Purpose**
User should not require to install `grunt-shell`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,6 @@
     "test": "grunt && grunt downloadBrowser && mocha test/unit && karma start test/karma.conf.js"
   },
   "dependencies": {
-    "grunt-shell": "^2.1.0",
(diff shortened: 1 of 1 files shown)


KaptenJansson:
@ratson Could you please the sign the CLA? Thanks!

@fippo Did this happen by accident or do you know if there is a reason for this to be under deps rather than dev-deps?

fippo:
missed that while reviewing #656 -- thanks for the fix!

KaptenJansson:
Doh I was the one that introduced this. I used grunt install save-dev-deps I thought......

ratson:
@googlebot I signed it!

KaptenJansson (approved):

PR #667, opened 2017-09-02
all: move createObjectURL shim to common shims

fippo:
moves the common shim out of utils. It is a shim, not a util


KaptenJansson (approved):
Well spotted :)

PR #675, opened 2017-09-11
chrome: override addtrack because of bugs

fippo:
the native addTrack (which is in chrome stable behind a flag) seems to have
some issues with video freezes:
    https://bugs.chromium.org/p/chromium/issues/detail?id=762527
overrides with the shim until this can be fixed.

@henbos PTAL (once CI is green)

### DIFF
--- a/src/js/chrome/chrome_shim.js
+++ b/src/js/chrome/chrome_shim.js
@@ -228,8 +228,10 @@ var chromeShim = {
   },
 
   shimAddTrackRemoveTrack: function(window) {
+    var browserDetails = utils.detectBrowser(window);
     // shim addTrack and removeTrack.
-    if (window.RTCPeerConnection.prototype.addTrack) {
+    if (window.RTCPeerConnection.prototype.addTrack &&
+        browserDetails.version >= 62) {
       return;
     }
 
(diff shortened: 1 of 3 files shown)


src/js/chrome/chrome_shim.js
```diff
@@ -228,8 +228,10 @@ var chromeShim = {
   },
 
   shimAddTrackRemoveTrack: function(window) {
+    var browserDetails = utils.detectBrowser(window);
     // shim addTrack and removeTrack.
-    if (window.RTCPeerConnection.prototype.addTrack) {
+    if (window.RTCPeerConnection.prototype.addTrack &&
+        browserDetails.version >= 63) {
```
henbos: Shouldn't this be >= 62 if you couldn't repro on 62.0.3202.9?

henbos (approved):
Approve % comment (is the version number correct?)

src/js/chrome/chrome_shim.js
fippo: yeah, that was written before i knew that 62.0.3202.9 fixed the issue :-)

KaptenJansson (approved):

PR #676, opened 2017-09-12
safari: shim legacy offerToReceive*

fippo:
shims offerToReceive in Safari to behave similar to how it behaves in other
browsers. Most importantly this means creating a transceiver on peerconnections
that have no local streams but call `createOffer({offerToReceiveVideo: true})`

Roughly https://github.com/fippo/adapter/tree/offerToReceive but without the e2e tests. The behaviour of createOffer({offerToReceive*: false}) is discussed [here](https://github.com/w3c/webrtc-pc/issues/1586) (and in linked PRs) but the tests would fail due to [this bug](https://bugs.webkit.org/show_bug.cgi?id=176665)

@youennf can you take a look? 
@jan-ivar can you check if the code is in line with your latest spec PRs?

### DIFF
--- a/src/js/safari/safari_shim.js
+++ b/src/js/safari/safari_shim.js
@@ -252,6 +252,45 @@ var safariShim = {
         }
       });
     }
+  },
+
+  shimCreateOfferLegacy: function(window) {
+    var origCreateOffer = window.RTCPeerConnection.prototype.createOffer;
+    window.RTCPeerConnection.prototype.createOffer = function(offerOptions) {
+      var pc = this;
+      if (offerOptions) {
+        var audioTransceiver = pc.getTransceivers().find(function(transceiver) {
+          return transceiver.sender.track &&
+              transceiver.sender.track.kind === 'audio';
+        });
+        if (offerOptions.offerToReceiveAudio === false && audioTransceiver) {
+          if (audioTransceiver.getDirection() === 'sendrecv') {
+            audioTransceiver.setDirection('sendonly');
+          } else if (audioTransceiver.getDirection() === 'recvonly') {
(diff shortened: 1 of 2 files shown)


youennf:
Patch looks good to me.
This will mostly work when the RTCPeerConnection is fresh (no transceivers basically).
Maybe a warning would help there.
setDirection will probably not be implemented until libwebrtc has a better support for it.

mizoRC:
Works for me too

fippo:
updated a bit do handle non-sendrecv transceivers a bit better (in theory)

KaptenJansson:
LGTM

src/js/safari/safari_shim.js
```diff
@@ -252,6 +252,45 @@ var safariShim = {
...
+        }
+
+        var videoTransceiver = pc.getTransceivers().find(function(transceiver) {
+          return transceiver.sender.track &&
+              transceiver.sender.track.kind === 'video';
+        });
+        if (offerOptions.offerToReceiveVideo === false && videoTransceiver) {
+          if (videoTransceiver.getDirection() === 'sendrecv') {
+            videoTransceiver.setDirection('sendonly');
+          } else if (audioTransceiver.getDirection() === 'recvonly') {
```
neilyoung, line 284: Wouldn't be "videoTransceiver.getDirection() == 'recvonly'" be correct?

fippo: this was fixed in #688 almost four years ago...

neilyoung: Well, I'm not the fastest :) Found this snippet somewhere else as wrong as it was. Thanks

PR #684, opened 2017-09-26
fix #683 RTCPeerConnection.getStats(mediaStreamTrack, callback) does not work correct

### LINKED ISSUE #683: RTCPeerConnection.getStats(mediaStreamTrack, callback) does not work correct
Chrome

**Description**
I call `RTCPeerConnection.getStats(mediaStreamTrack, callback)` and I expect to get result in the callback. But my callback will be called without arguments.
Because of this code:
```
      // When spec-style getStats is supported, return those when called with
      // either no arguments or the selector argument is null.
      if (origGetStats.length === 0 && (arguments.length === 0 ||
          typeof arguments[0] !== 'function')) {
        return origGetStats.apply(this, []);
      }
```
As we can see the comment contradicts the code.  
`or the selector argument is null.` and `typeof arguments[0] !== 'function'`

p.s. I can do a merge request if you confirm it as a bug.
**Steps to reproduce**
Just call the `RTCPeerConnection.getStats(mediaStreamTrack, callback)` and see the empty arguments inside the callback.

**Expected results**
Stats as an argument inside the callback

**Actual results**
There are no any arguments (because of `return origGetStats.apply(this, []);`)

Keksinautin:
**Description**
The full [desc](https://github.com/webrtc/adapter/issues/683)
```js
      // When spec-style getStats is supported, return those when called with
      // either no arguments or the selector argument is null.
      if (origGetStats.length === 0 && (arguments.length === 0 ||
          typeof arguments[0] !== 'function')) {
        return origGetStats.apply(this, []);
      }
```
As we can see the comment contradicts the code.
or the selector argument is null. and typeof arguments[0] !== 'function'

**Purpose**
When I call `RTCPeerConnection.getStats(mediaStreamTrack, callback)` I would like to be here:  
`return origGetStats.apply(this, [successCallbackWrapper_, arguments[0]]);`

### DIFF
--- a/src/js/chrome/chrome_shim.js
+++ b/src/js/chrome/chrome_shim.js
@@ -538,7 +538,7 @@ var chromeShim = {
       // When spec-style getStats is supported, return those when called with
       // either no arguments or the selector argument is null.
       if (origGetStats.length === 0 && (arguments.length === 0 ||
-          typeof arguments[0] !== 'function')) {
+          selector == null)) {
         return origGetStats.apply(this, []);
       }
 


KaptenJansson:
Thanks for contributing. Could you fix the errors as seen in the travis runs? Thanks!

KaptenJansson:
Sorry did not read the original issue careful enough. Closing this.

PR #688, opened 2017-10-03
safari: small fixes in offerToReceive

jcague:
**Description**
Safari doesn't seem to implement transceiver.getDirection() function and the library throws an error if we try to call it.

Also fix a wrong access to audioTransceiver when we wanted to use videoTransceiver.

### DIFF
--- a/src/js/safari/safari_shim.js
+++ b/src/js/safari/safari_shim.js
@@ -264,9 +264,9 @@ var safariShim = {
               transceiver.sender.track.kind === 'audio';
         });
         if (offerOptions.offerToReceiveAudio === false && audioTransceiver) {
-          if (audioTransceiver.getDirection() === 'sendrecv') {
+          if (audioTransceiver.direction === 'sendrecv') {
             audioTransceiver.setDirection('sendonly');
-          } else if (audioTransceiver.getDirection() === 'recvonly') {
+          } else if (audioTransceiver.direction === 'recvonly') {
             audioTransceiver.setDirection('inactive');
           }
         } else if (offerOptions.offerToReceiveAudio === true &&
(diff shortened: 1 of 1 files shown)


jcague:
I signed it

src/js/safari/safari_shim.js
```diff
@@ -263,7 +263,8 @@ var safariShim = {
           return transceiver.sender.track &&
               transceiver.sender.track.kind === 'audio';
         });
-        if (offerOptions.offerToReceiveAudio === false && audioTransceiver) {
+        if (offerOptions.offerToReceiveAudio === false && audioTransceiver
+            && typeof audioTransceiver.getDirection === 'function') {
```
fippo: hrm... looks like getDirection was removed. I'd prefer this:
```'getDirection' in window.RTCRtpTransceiver.prototype```

Unfortunately the lack of getDirection breaks the functionality of this whole block :-/

File: src/js/safari/safari_shim.js [hunk 2]
```diff
@@ -278,10 +279,11 @@ var safariShim = {
           return transceiver.sender.track &&
               transceiver.sender.track.kind === 'video';
         });
-        if (offerOptions.offerToReceiveVideo === false && videoTransceiver) {
+        if (offerOptions.offerToReceiveVideo === false && videoTransceiver
+            && typeof videoTransceiver.getDirection === 'function') {
           if (videoTransceiver.getDirection() === 'sendrecv') {
             videoTransceiver.setDirection('sendonly');
-          } else if (audioTransceiver.getDirection() === 'recvonly') {
+          } else if (videoTransceiver.getDirection() === 'recvonly') {
```
fippo: good catch!

fippo (approved):
LGTM with style nits

src/js/safari/safari_shim.js
fippo: ah... it is not getDirection. It needs to be `audioTransceiver.direction ===`

fippo:
can you remove `&& 'getDirection' in window.RTCRtpTransceiver.prototype) {` and replace the .getDirection() calls with .direction?

(need to file a spec-bug, having a setXXX and checking with .XXX seems inconsistent)

fippo:
ah... we still need [setDirection()](http://w3c.github.io/webrtc-pc/#dom-rtcrtptransceiver-setdirection) (and I suspect you understand why I am confused now) :-)

jcague:
@fippo yes, and I didn't try the API in the latest change. Now it should work in all cases.

fippo (approved):

fippo:
(this should still go in the next patch before we do a major break)

KaptenJansson (approved):

KaptenJansson:
Merging with test failure as they are unrelated.

PR #689, opened 2017-10-05
chrome: bump native chrome addTrack to 63+

fippo:
lets postpone using the native one a bit further, there still seems to be something broken with DTMF (but behind flag)

@henbos PTAL

### DIFF
--- a/src/js/chrome/chrome_shim.js
+++ b/src/js/chrome/chrome_shim.js
@@ -231,7 +231,7 @@ var chromeShim = {
     var browserDetails = utils.detectBrowser(window);
     // shim addTrack and removeTrack.
     if (window.RTCPeerConnection.prototype.addTrack &&
-        browserDetails.version >= 62) {
+        browserDetails.version >= 63) {
       return;
     }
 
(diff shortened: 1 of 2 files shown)


henbos (approved):
LGTM

fippo:
tests seem flaky, locally it passes

fippo:
@henbos I'm a bit concerned that we still get failures with experimental platform features flag on. I can reproduce with this build locally. Maybe something interferes now?

We might decide to update the expectation, I doubt 62 is fix-able now.

fippo:
the second failure is caused by `'ontrack' in RTCPeerConnection.prototype and us not overriding that. That looks like something we might have to either override or extend the shim to polyfill .transceiver even in native ontrack

henbos:
I believe ["track event / is called by setRemoteDescription during renegotiation"](https://github.com/webrtc/adapter/blob/7160e47653b9cd6394f0835b524ef5c4d779cc00/test/e2e/ontrack.js#L98) fails because the shim relies on streams being added events ([which only fires for new streams](https://crbug.com/768131)), and when track2 is added stream1 already exists. I believe this should always fail and we should update the expected files.

What's the shim status of transceiver (not supported yet)?

What's the flaky bit? From what I can tell we do things based on promises resolving, no race, and they should either succeed or fail every time. If something is flaky maybe we're doing something wrong with shim promises?

fippo:
we do shim .transceiver in crude (but effective) ways: https://github.com/webrtc/adapter/blob/master/src/js/chrome/chrome_shim.js#L54

fippo:
added a commit to update one test expectation. jsfiddle for the renegotiation one:
https://jsfiddle.net/p1kueu86/
Still happens in chrome-unstable for me, you might want to take a look.

PR #692, opened 2017-10-06
chrome: don't shim getUserMedia constraints in chrome 61+

### LINKED ISSUE #553: Chrome 59 supports spec-constraints
Chrome 59 supports spec-constraints for getUserMedia so we should stop translating constraints.

@guidou: the release notes are quite unhelpful about saying where this was actually fixed, can you point to a bug? This might be https://bugs.chromium.org/p/chromium/issues/detail?id=682887 but that has no associated CLs so it is hard to judge what works and what is still left for https://bugs.chromium.org/p/chromium/issues/detail?id=657733

fippo:
fixes #553. Since this is a rather drastic change in behaviour
this should be a major version bump.

based on top of the bandaid one, can rebase after merge

### DIFF
--- a/src/js/chrome/getusermedia.js
+++ b/src/js/chrome/getusermedia.js
@@ -67,6 +67,9 @@ module.exports = function(window) {
   };
 
   var shimConstraints_ = function(constraints, func) {
+    if (browserDetails.version >= 61) {
+      return func(constraints);
+    }
     constraints = JSON.parse(JSON.stringify(constraints));
     if (constraints && typeof constraints.audio === 'object') {
       var remap = function(obj, a, b) {


src/js/chrome/getusermedia.js
```diff
@@ -84,7 +87,7 @@ module.exports = function(window) {
       // Shim facingMode for mobile & surface pro.
       var face = constraints.video.facingMode;
       face = face && ((typeof face === 'object') ? face : {ideal: face});
-      var getSupportedFacingModeLies = browserDetails.version < 61;
+      var getSupportedFacingModeLies = browserDetails.version < 66;
```
tgabi333: i think there is no need for this check since version is always lower than 61 here

fippo: yeah but if I remove it that might lead to more code changes than I want to do.

KaptenJansson (approved):
cc @henbos @guidou

PR #697, opened 2017-10-11
chrome: update chrome beta w/experimental expectation

fippo:
from #689, upstream bug in https://bugs.chromium.org/p/chromium/issues/detail?id=773613


KaptenJansson (approved):

KaptenJansson:
cc @henbos

PR #700, opened 2017-10-16
firefox: shim removeStream

fippo:
shims removeStream as a polyfill using removeTrack in Firefox
while emitting a deprecation warning.

cc @nils-ohlmeier

### DIFF
--- a/src/js/firefox/firefox_shim.js
+++ b/src/js/firefox/firefox_shim.js
@@ -196,6 +196,20 @@ var firefoxShim = {
         })
         .then(onSucc, onErr);
     };
+  },
+
+  shimRemoveStream: function(window) {
+    if ('removeStream' in window.RTCPeerConnection.prototype) {
+      return;
+    }
+    window.RTCPeerConnection.prototype.removeStream = function(stream) {
+      utils.deprecated('removeStream', 'removeTrack');
+      this.getSenders().forEach(function(sender) {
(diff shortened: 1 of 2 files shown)


src/js/firefox/firefox_shim.js
```diff
@@ -196,6 +196,21 @@ var firefoxShim = {
...
+
+  shimRemoveStream: function(window) {
+    if ('removeStream' in window.RTCPeerConnection.prototype) {
+      return;
+    }
+    window.RTCPeerConnection.prototype.removeStream = function(stream) {
+      utils.deprecated('removeStream', 'removeTrack');
+      this.getSenders().forEach(function(sender) {
+        var tracks = stream.getTracks();
+        if (sender.track && tracks.indexOf(sender.track) !== -1) {
```
jan-ivar: I think this should fit:
```js
        if (sender.track && stream.getTracks().indexOf(sender.track) !== -1) {
```

jan-ivar (approved):
😞 though this makes adapter consistent across browsers, and at least we nag, which is great!

src/js/firefox/firefox_shim.js
fippo: done + squashed

KaptenJansson (approved):

fippo:
minor release please since this is a feature. Well, actually its a non-feature but...

PR #701, opened 2017-10-23
test: update chrome stable (w/experimental) expectation

fippo:
same as #697 but now Chrome62 became stable.

cc @henbos

### DIFF
--- a/test/e2e/expectations/chrome-stable
+++ b/test/e2e/expectations/chrome-stable
@@ -46,13 +46,13 @@ PASS navigator.mediaDevices enumerateDevices returns some audiooutput devices
 PASS MediaStream window.MediaStream exists
 PASS MSID signals track and stream ids
 PASS track event RTCPeerConnection.prototype.ontrack exists
-PASS track event is called by setRemoteDescription during renegotiation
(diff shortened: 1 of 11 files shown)


test/e2e/expectations/chrome-stable
```diff
@@ -46,13 +46,13 @@ PASS navigator.mediaDevices enumerateDevices returns some audiooutput devices
 PASS MediaStream window.MediaStream exists
 PASS MSID signals track and stream ids
 PASS track event RTCPeerConnection.prototype.ontrack exists
-PASS track event is called by setRemoteDescription during renegotiation
+ERR track event is called by setRemoteDescription during renegotiation timeout
```
henbos: It would be good to rename the test to indicate that this is adding a track to an existing stream. This is key to why the track event doesn't fire in 62, it *does* fire during renegotiation for any track added to streams that don't already exist.

henbos (approved):
LGTM with or without comment addressed.

test/e2e/expectations/chrome-stable
fippo: will do (in a separate commit but keeping this PR open)

henbos (approved):

PR #702, opened 2017-10-26
Data channel message size limitations

lgrahl:
Recent changes in the W3C spec will ensure that a `TypeError` is thrown in case
a data channel message is being sent that is too large for the other peer to
receive. However, that size limitation needs to be available to the sender or
it's an annoying guessing game.

This PR adds both access to `maxMessageSize` and throws a `TypeError` in case
the message is too large for the other peer to receive.

### DIFF
--- a/src/js/common_shim.js
+++ b/src/js/common_shim.js
@@ -187,17 +187,16 @@ module.exports = {
       // Note: 65535 bytes is the default value from the SDP spec. Also,
       //       every implementation we know supports receiving 65535 bytes.
       var maxMessageSize = 65535;
-      var match = description.sdp.match(/a=max-message-size:\s*(\d+)/);
-      if (match !== null && match.length >= 2) {
-        maxMessageSize = parseInt(match[1]);
+      var match = SDPUtils.matchPrefix(description.sdp, 'a=max-message-size:');
+      if (match.length > 0) {
+        maxMessageSize = parseInt(match[0]);
       }
       return maxMessageSize;
     };
 
     var isFirefox = function(description) {
       // TODO: Is there a better solution for detecting Firefox?
-      var match = description.sdp.match(/mozilla...THIS_IS_SDPARTA/);
-      return (match !== null && match.length >= 1);
+      return description.sdp.indexOf('mozilla...THIS_IS_SDPARTA') !== -1;
     };
 
     var maybeApplyMaxMessageSize = function(pc) {
--- a/src/js/common_shim.js
+++ b/src/js/common_shim.js
@@ -183,7 +183,7 @@ module.exports = {
       return (match !== null && match.length >= 1);
     };
 
-    var getFirefoxVersion = function(description) {
+    var getRemoteFirefoxVersion = function(description) {
       // TODO: Is there a better solution for detecting Firefox?
       var match = description.sdp.match(/mozilla...THIS_IS_SDPARTA-(\d+)/);
       if (match === null || match.length < 2) {
@@ -243,7 +243,7 @@ module.exports = {
 
       if (sctpInDescription(arguments[0])) {
         // Check if the remote is FF.
-        var isFirefox = getFirefoxVersion(arguments[0]);
+        var isFirefox = getRemoteFirefoxVersion(arguments[0]);
 
         // Get the maximum message size the local peer is capable of sending
         var canSendMMS = getCanSendMaxMessageSize(arguments[0], isFirefox);
--- a/src/js/common_shim.js
+++ b/src/js/common_shim.js
@@ -292,12 +292,9 @@ module.exports = {
   },
 
   shimSendThrowTypeError: function(window) {
-    var browserDetails = utils.detectBrowser(window);
-
-    // Only Firefox 57 has support for this atm
-    if (browserDetails.browser === 'firefox' && browserDetails.version >= 57) {
-      return;
-    }
+    // Note: Although Firefox >= 57 has a native implementation, the maximum
+    //       message size is not applied correctly to all data channels.
+    //       See: https://bugzilla.mozilla.org/show_bug.cgi?id=1426831
 
     var origCreateDataChannel =
       window.RTCPeerConnection.prototype.createDataChannel;
--- a/src/js/common_shim.js
+++ b/src/js/common_shim.js
@@ -179,8 +179,13 @@ module.exports = {
     }
 
     var sctpInDescription = function(description) {
-      var match = description.sdp.match(/m=application\s+\d+\s+[\w/]*SCTP/);
-      return (match !== null && match.length >= 1);
+      var sections = SDPUtils.splitSections(description.sdp);
+      sections.shift();
+      return sections.some(function(mediaSection) {
+        var mLine = SDPUtils.parseMLine(mediaSection);
+        return mLine && mLine.kind === 'application'
+            && mLine.protocol.indexOf('SCTP') !== -1;
+      });
     };
 
     var getRemoteFirefoxVersion = function(description) {
--- a/src/js/common_shim.js
+++ b/src/js/common_shim.js
@@ -204,7 +204,7 @@ module.exports = {
       // Note: Although Chrome is technically able to send up to 256 KiB, the
       //       data does not reach the other peer reliably.
       //       See: https://bugs.chromium.org/p/webrtc/issues/detail?id=8419
-      var canSendMaxMessageSize = 65535;
+      var canSendMaxMessageSize = 65536;
       if (browserDetails.browser === 'firefox') {
         if (browserDetails.version < 57) {
           if (remoteIsFirefox === -1) {
@@ -225,9 +225,9 @@ module.exports = {
     };
 
     var getMaxMessageSize = function(description, remoteIsFirefox) {
-      // Note: 65535 bytes is the default value from the SDP spec. Also,
-      //       every implementation we know supports receiving 65535 bytes.
-      var maxMessageSize = 65535;
+      // Note: 65536 bytes is the default value from the SDP spec. Also,
+      //       every implementation we know supports receiving 65536 bytes.
+      var maxMessageSize = 65536;
       var match = SDPUtils.matchPrefix(description.sdp, 'a=max-message-size:');
       if (match.length > 0) {
         maxMessageSize = parseInt(match[0].substr(19), 10);
--- a/src/js/common_shim.js
+++ b/src/js/common_shim.js
@@ -194,9 +194,15 @@ module.exports = {
       return maxMessageSize;
     };
 
-    var isFirefox = function(description) {
+    var getFirefoxVersion = function(description) {
       // TODO: Is there a better solution for detecting Firefox?
-      return description.sdp.indexOf('mozilla...THIS_IS_SDPARTA') !== -1;
+      var match = description.sdp.match(/mozilla...THIS_IS_SDPARTA-(\d+)/);
+      if (match === null || match.length < 2) {
+        return -1;
+      }
+      var version = parseInt(match[1], 10);
+      // Test for NaN (yes, this is ugly)
+      return version !== version ? -1 : version;
     };
 
     var maybeApplyMaxMessageSize = function(pc) {
@@ -214,8 +220,9 @@ module.exports = {
                                     pc._remoteMaxMessageSize);
         }
 
-        // FF can send (at least) 1 GiB towards all other FF.
-        if (browserDetails.browser === 'firefox' && pc._otherPeerIsFirefox) {
+        // Any FF can send (at least) 1 GiB towards all other FF < 57.
+        if (browserDetails.browser === 'firefox' &&
+            pc._otherPeerFirefoxVersion < 57) {
           maxMessageSize = 1073741823;
         }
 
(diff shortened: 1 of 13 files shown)


lgrahl:
I signed it!

src/js/common_shim.js
```diff
@@ -162,5 +162,77 @@ module.exports = {
...
+    var origSetRemoteDescription =
+        window.RTCPeerConnection.prototype.setRemoteDescription;
+    window.RTCPeerConnection.prototype.setRemoteDescription = function() {
+      var pc = this;
+
+      // Determine the maximum message size of the remote peer.
+      // Note: 65535 bytes is the default value from the SDP spec. Also,
+      //       every implementation we know supports receiving 65535 bytes.
+      var maxMessageSize = 65535;
+      var match = arguments[0].sdp.match(/a=max-message-size:\s*(\d+)/);
```
fippo: you might consider requiring sdp and using matchPrefix -- its a sub-dependency of the edge shim currently so wouldn't increase "binary size"

fippo (approved):

lgrahl:
This PR has a few issues I've detected. I'll tick them one by one and then it's ready to be merged (from my side):
- [x] The local maximum message size needs to be determined, too. Chrome for example will fail beyond 256 KiB even if the remote maximum message size is above that.
- [x] Even though Firefox<=56 to Firefox<=56 is capable of sending large messages to one another, this PR will prevent that from being possible. This one is tricky... I resolved this in a rather hacky way by looking for `mozilla...THIS_IS_SDPARTA` in the remote description's SDP.
- [x] `a=max-message-size:0` would fail at the moment.
- [x] Tests needed.
- [x] Use `matchPrefix` instead of regular expressions (where possible).

lgrahl:
The new tests should be fine now but Travis is still failing (in other tests) - ideas?

lgrahl:
FWIW, here's a list of determined `maxMessageSize` values:
* FF<=56 to FF<=56: 1073741823 bytes (slightly incorrect but fine)
* FF<=56 to FF>=57: 1073741823 bytes (slightly incorrect but fine)
* FF<=56 to Chrome: 16384 bytes
* FF<=56 to RAWRTC: 16384 bytes (very incorrect but fine)
* FF>=57 to FF<=56: 1073741823 bytes (slightly incorrect but fine)
* FF>=57 to FF>=57: 1073741823 bytes
* FF>=57 to Chrome: 65535 bytes
* FF>=57 to RAWRTC: 1073741823 bytes
* Chrome to FF<=56: 65535 bytes
* Chrome to FF>=57: 65535 bytes
* Chrome to Chrome: 65535 bytes
* Chrome to RAWRTC: 65535 bytes
* RAWRTC to FF<=56: 65535 bytes
* RAWRTC to FF>=57: 1073741823 bytes
* RAWRTC to Chrome: 65535 bytes
* RAWRTC to RAWRTC: unlimited (0)

Couldn't test Safari as I don't own any Apple products.
If you want to know why the *incorrect*'s are fine, just ping me.

lgrahl:
There's one last quirk remaining: I don't know why FF >= 57 doesn't raise an exception in the tests when sending a message that's too large. It does behave correctly when testing this by hand.

lgrahl:
I'm done. Can be reviewed. Some FF stable test timed out on Travis and needs to be restarted (runs fine locally).

src/js/common_shim.js
```diff
@@ -162,5 +162,143 @@ module.exports = {
...
+      });
+    }
+
+    var sctpInDescription = function(description) {
+      var match = description.sdp.match(/m=application\s+\d+\s+[\w/]*SCTP/);
+      return (match !== null && match.length >= 1);
+    };
+
+    var getFirefoxVersion = function(description) {
+      // TODO: Is there a better solution for detecting Firefox?
```
alvestrand: If this is for detecting the remote browser's Firefox version, please call it detectRemoteFirefoxVersion. For local browser, we have utils.detectBrowser().

src/js/common_shim.js
```diff
@@ -162,5 +162,143 @@ module.exports = {
...
+      }
+      var version = parseInt(match[1], 10);
+      // Test for NaN (yes, this is ugly)
+      return version !== version ? -1 : version;
+    };
+
+    var getCanSendMaxMessageSize = function(description, remoteIsFirefox) {
+      // Every implementation we know can send at least 64 KiB.
+      // Note: Although Chrome is technically able to send up to 256 KiB, the
+      //       data does not reach the other peer reliably.
```
alvestrand: If you have an issue number for Chrome here, please add it. (If not, please please please file one!)

test/e2e/expectations/chrome-beta
```diff
@@ -2,12 +2,12 @@ PASS addIceCandidate after setRemoteDescription resolves when called with null
 PASS addIceCandidate after setRemoteDescription resolves when called with undefined
 PASS addTrack throws an exception if the track has already been added
 PASS addTrack throws an exception if the track has already been added via addStream
-PASS addTrack throws an exception if addStream is called with a stream containing a track already added
+ERR addTrack throws an exception if addStream is called with a stream containing a track already added AssertionError
```
alvestrand: Why have these tests started failling with the patch?

test/e2e/maxMessageSize.js
```diff
@@ -0,0 +1,149 @@
...
+    });
+  }
+
+  function patchMaxMessageSizeFactory(maxMessageSize) {
+    return ((description) => {
+      description.sdp = description.sdp.replace(
+        /^a=max-message-size:\s*(\d+)\s*$/gm, '');
+      description.sdp = description.sdp.replace(
+        /(^m=application\s+\d+\s+[\w\/]*SCTP.*$)/m,
+        '$1\r\na=max-message-size:' + maxMessageSize);
```
alvestrand, line 49: Won't this insert a second max-message-size line into your SDP if it already has one?

test/e2e/maxMessageSize.js
```diff
@@ -0,0 +1,149 @@
...
+        expect(pc1.sctp.maxMessageSize).to.equal(1073741823);
+        expect(pc2.sctp.maxMessageSize).to.equal(1073741823);
+      }
+    });
+  });
+
+  it('send largest possible single message', () => {
+    // Note: Patching to 65535 here as anything beyond that will take too long.
+    const maxMessageSize = 65535;
+    const patchMaxMessageSize = patchMaxMessageSizeFactory(maxMessageSize);
```
alvestrand: Will this test always fail if the negotiated size is < 65535?

alvestrand (commented):
Nice add, and nice tests, but I worry about the tests that have started failing with AssertionError, where you just patched the expected files (test/e2e/expectations/chrome-beta).

src/js/common_shim.js
lgrahl: Ok.

src/js/common_shim.js
lgrahl: Good point. I'll check if there is one already or file it.

test/e2e/expectations/chrome-beta
lgrahl: Ah, it wasn't my intention to commit this. There was an issue with tests but @fippo told me it might have been a new Chrome beta version. I'll rebase against master and we'll see if it works.

test/e2e/maxMessageSize.js
lgrahl: No, the `replace` in line 40 will remove the entire `a=max-message-size` line if there is one.

test/e2e/maxMessageSize.js
lgrahl: The short answer is *yes*, but it's a little bit more complicated than that. Every implementation supports 64 KiB unless one that doesn't handle EOR correctly fiddles with usrsctp's buffer sizes (256 KiB send/receive buffer by default results in the partial delivery size of 64 KiB) which no one has done so far. But there's at least one case where the negotiated size is 16 KiB (Firefox to Chrome). That's not a problem here because these tests will always create a peer connection with the same browser.

src/js/common_shim.js
```diff
@@ -162,5 +162,143 @@ module.exports = {
...
+        }
+      }
+      return canSendMaxMessageSize;
+    };
+
+    var getMaxMessageSize = function(description, remoteIsFirefox) {
+      // Note: 65535 bytes is the default value from the SDP spec. Also,
+      //       every implementation we know supports receiving 65535 bytes.
+      var maxMessageSize = 65535;
+      var match = SDPUtils.matchPrefix(description.sdp, 'a=max-message-size:');
```
fippo: where is the sdp package imported? I don't understand why this works ;-)

lgrahl: It was already imported. :)

https://github.com/lgrahl/adapter/blob/ba8c0fa4ed39cc79f66ff418f0a292a83ae7b6ef/src/js/common_shim.js#L11

fippo: why does github not show it in the diff? Also note that you need to add it to package.json, currently it is a child dependeny

lgrahl:
Rebased against master.

src/js/common_shim.js
lgrahl: Because I didn't add it. It was already there. It looks like it is in `package.json` already? https://github.com/webrtc/adapter/blob/master/package.json#L23

fippo: I need new 👓

lgrahl: No worries. Me too. :grin: And it's better to double-check.

src/js/common_shim.js
```diff
@@ -162,5 +162,143 @@ module.exports = {
...
+    if (!('sctp' in window.RTCPeerConnection.prototype)) {
+      Object.defineProperty(window.RTCPeerConnection.prototype, 'sctp', {
+        get: function() {
+          return typeof this._sctp === 'undefined' ? null : this._sctp;
+        }
+      });
+    }
+
+    var sctpInDescription = function(description) {
+      var match = description.sdp.match(/m=application\s+\d+\s+[\w/]*SCTP/);
```
fippo: how about this:
```
      var sections = SDPUtils.splitSections(description.sdp);
      sections.shift();
      return sections.some(function(mediaSection) {
        var mLine = SDPUtils.parseMLine(mediaSection);
        return mLine && mLine.kind === 'application'
            && mLine.protocol === 'DTLS/SCTP';
      });
```
if we have SDPUtils anyway... needs a min version of 2.4.0 however

lgrahl: Personally, I don't mind either solutions. Be aware that it may not always be `DTLS/SCTP` but also `UDP/DTLS/SCTP` and `TCP/DTLS/SCTP`, see: https://tools.ietf.org/html/draft-ietf-mmusic-sctp-sdp-26#section-4.2

src/js/common_shim.js
lgrahl: Tracked [here](https://bugs.chromium.org/p/webrtc/issues/detail?id=8419)

alvestrand:
Someone modified the DTMF test to use a transceiver. Transceivers aren't in Chrome stable.
I don't understand how this could be caused by this PR, so perhaps mainline is busted?

fippo:
@alvestrand no. The transceiver check fails (https://travis-ci.org/webrtc/adapter/jobs/299697848#L2760) but that is in the expectation file.

The DTMF/addTrack test which fails [against expectation](https://travis-ci.org/webrtc/adapter/jobs/299697848#L2795) has been flaky (on travis) for quite some time. More so with experimental webplatform features if I recall correctly. Hitting rerun (which requires commit bit unfortunately) a couple of times usually solves it :-/

(I would only do so after squashing)

lgrahl:
Anything left preventing this from being merged?

KaptenJansson (approved):

KaptenJansson:
Set for major release since this a new shim?

fippo:
minor version is for new features. Unless 🐔 🐔 🐔

lgrahl:
~~Ping.~~

Edit: I just found a minor bug: The default MMS should be `65536` instead of `65535`... always gets me. :unamused: Will push a fix.

Edit: [This also affects FF >= 57](https://bugzilla.mozilla.org/show_bug.cgi?id=1425697). If we can't get a backport fix, this PR will need updating once more.

lgrahl:
FF 57 will not be fixed, so I'll add a fix to this PR.
I'll also add some further tests to go through a lot of negotiated MMS cases because they're pretty complicated now.

lgrahl:
Some tests disconnected and need to be restarted. (Some still fail in other places but I guess you know that already as it's also happening on master.)

lgrahl:
Rebased against master.
Currently investigating why the test runner doesn't catch the exception raised by the native `send` call of Firefox >= 57.

lgrahl:
Resolved the problem with the test. It turned out to be [a bug in Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1426831) and I had to dramatically decrease the maximum message size.

Please, don't be irritated by me adding more commits. It's just that I didn't want to hold back these changes. It's good to merge now (from my side) and I'll backlog further changes in another branch until this is getting merged.

Edit: URL fixed.

fippo:
finding bugs by writing code... 👍 👍 👍

KaptenJansson:
@alvestrand any thoughts?

alvestrand:
Looks good to me, and the "update branch" button worked (for a wonder). I'll merge it once the checks complete.

PR #708, opened 2017-11-08

henbos:


henbos:
PTAL janson. Hey, how do we make sure I can commit with this account?

KaptenJansson:
Use a google email

henbos:
(Fixed email problem in #710.)

PR #709, opened 2017-11-08

henbos:
Sorry this is a duplicate of #708, trying to reupload with email config updated.


henbos:
(Fixed email problem in #710.)

PR #710, opened 2017-11-08

henbos:

(Sorry for spam, this is a duplicate of #708 and #709, email should be fixed now... fingers crossed.)


henbos:
PTAL jansson, worked this time :)

KaptenJansson (approved):

PR #721, opened 2017-11-23
test: relax tests relying on matching error strings, only use type

fippo:
relaxes the checking of some error strings. Not sure if this is complete but lets see what *else* but unstable breaks.


henbos (approved):
lgtm

PR #722, opened 2017-11-23
utils: change browser detection to use UA sniffing instead of prefixed stuff [MAJOR] [WIP]

### LINKED ISSUE #712: Chrome will be identified as Safari if webkitGetUserMedia and/or webkitRTCPeerConnection are removed
The problematic code is here:
https://github.com/webrtc/adapter/blob/2f9ba8c9125cc6e79eaa416c177a60ce213d0a7c/src/js/utils.js#L99

By two different fallback paths, if either or both are removed, Chrome would be identified as Safari, and that will likely have some other consequences.

These are the Chromium issues for removing these aliases:
https://bugs.chromium.org/p/chromium/issues/detail?id=690582
https://bugs.chromium.org/p/chromium/issues/detail?id=692736

When I looked into this a month or so ago, I gave up as soon as I noticed this pattern in adapter.js.

What would need to happen is that either detectBrowser is removed entirely so that nothing anywhere assumes that the combination of certain APIs implies that it's any particular browser. This is likely difficult, if feature detection worked for everything then presumably this mechanism wouldn't be there.

Once fixed, does anyone have an idea of the half-life of adapter.js in the wild? How many years would one have to wait until usage of current versions has dropped to 1% of the current usage, for example?

fippo:
fixes #712 

Note: major bump required for safety reasons

### DIFF
--- a/src/js/utils.js
+++ b/src/js/utils.js
@@ -92,36 +92,21 @@ module.exports = {
       return result;
     }
 
-    // Firefox.
-    if (navigator.mozGetUserMedia) {
+    if (navigator.userAgent.match(/Firefox\/(\d+)\./)) { // Firefox.
       result.browser = 'firefox';
       result.version = extractVersion(navigator.userAgent,
           /Firefox\/(\d+)\./, 1);
-    } else if (navigator.webkitGetUserMedia) {
-      // Chrome, Chromium, Webview, Opera, all use the chrome shim for now
-      if (window.webkitRTCPeerConnection) {
-        result.browser = 'chrome';
-        result.version = extractVersion(navigator.userAgent,
-          /Chrom(e|ium)\/(\d+)\./, 2);
-      } else { // Safari (in an unpublished version) or unknown webkit-based.
-        if (navigator.userAgent.match(/Version\/(\d+).(\d+)/)) {
-          result.browser = 'safari';
-          result.version = extractVersion(navigator.userAgent,
-            /AppleWebKit\/(\d+)\./, 1);
-        } else { // unknown webkit-based browser.
-          result.browser = 'Unsupported webkit-based browser ' +
-              'with GUM support but no WebRTC support.';
-          return result;
-        }
-      }
-    } else if (navigator.mediaDevices &&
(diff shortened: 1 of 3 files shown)


src/js/utils.js
```diff
@@ -92,27 +92,15 @@ var utils = {
...
-          result.browser = 'safari';
-          result.version = this.extractVersion(navigator.userAgent,
-            /AppleWebKit\/(\d+)\./, 1);
-        } else { // unknown webkit-based browser.
-          result.browser = 'Unsupported webkit-based browser ' +
-              'with GUM support but no WebRTC support.';
-          return result;
-        }
-      }
     } else if (navigator.mediaDevices &&
```
foolip: It's harmless, but does the `navigator.mediaDevices` check need to be here?

fippo: I *think* all public versions of Edge shipped with navigator.mediaDevices.getUserMedia. Will poke folks but i am all for removing it

src/js/utils.js
```diff
@@ -92,27 +92,15 @@ var utils = {
...
-      // Chrome, Chromium, Webview, Opera, all use the chrome shim for now
-      if (window.webkitRTCPeerConnection) {
-        result.browser = 'chrome';
-        result.version = this.extractVersion(navigator.userAgent,
+    } else if (navigator.userAgent.match(/Chrom(e|ium)\/(\d+)\./)) {
+      // Chrome, Chromium, Webview, Opera, all use the chrome shim.
+      result.browser = 'chrome';
+      result.version = this.extractVersion(navigator.userAgent,
           /Chrom(e|ium)\/(\d+)\./, 2);
-      } else { // Safari (in an unpublished version) or unknown webkit-based.
```
fippo: safari didn't ship with webkitRTCPeerConnection. Maybe webkitgtk / epiphany does?

fippo:
surprisingly this worked. I'll run a few forward merges against my (rather complex) codebase to see if there is stuff not covered by the tests.

src/js/utils.js
fippo: shipped with windows 10 TH2 (10586) so we're good.

KaptenJansson:
Closing this in favor of #749

PR #726, opened 2017-11-24
chrome: shim getLocalStreams() with native removeTrack()

fippo:
shims the legacy non-spec together with the native addTrack to
provide a behaviour similar to Firefox.

@henbos not sure how one would do removeTrack. Say we have this:
```
var audioSender = pc.addTrack(audioTrack, stream);
var videoSender = pc.addTrack(videoTrack, stream);
pc.removeTrack(audioSender);
pc.getLocalStreams() // should be [stream]
pc.removeTrack(videoSender);
pc.getLocalStreams() // should be []
```
tests for that fail currently (yay). We should probably also have a test that tests that getLocalStreams() is still [stream] after the first removeTrack

### DIFF
--- a/src/js/chrome/chrome_shim.js
+++ b/src/js/chrome/chrome_shim.js
@@ -228,12 +228,88 @@ module.exports = {
     }
   },
 
+  shimAddTrackRemoveTrackWithNative: function(window) {
+    // shim addTrack/removeTrack with native variants in order to make
+    // the interactions with legacy getLocalStreams behave as in other browsers.
+    // Keeps a mapping stream.id => [stream, rtpsenders...]
+    window.RTCPeerConnection.prototype.getLocalStreams = function() {
+      var pc = this;
+      this._shimmedLocalStreams = this._shimmedLocalStreams || {};
+      return Object.keys(this._shimmedLocalStreams).map(function(streamId) {
+        return pc._shimmedLocalStreams[streamId][0];
+      });
+    };
+
+    var origAddTrack = window.RTCPeerConnection.prototype.addTrack;
+    window.RTCPeerConnection.prototype.addTrack = function(track, stream) {
+      if (!stream) {
+        return origAddTrack.apply(this, arguments);
+      }
+      this._shimmedLocalStreams = this._shimmedLocalStreams || {};
+
+      var sender = origAddTrack.apply(this, arguments);
+      if (!this._shimmedLocalStreams[stream.id]) {
+        this._shimmedLocalStreams[stream.id] = [stream, sender];
+      } else if (this._shimmedLocalStreams[stream.id].indexOf(sender) === -1) {
+        this._shimmedLocalStreams[stream.id].push(sender);
+      }
+      return sender;
+    };
+
+    var origAddStream = window.RTCPeerConnection.prototype.addStream;
+    window.RTCPeerConnection.prototype.addStream = function(stream) {
+      var pc = this;
+      this._shimmedLocalStreams = this._shimmedLocalStreams || {};
+
+      stream.getTracks().forEach(function(track) {
+        var alreadyExists = pc.getSenders().find(function(s) {
+          return s.track === track;
+        });
+        if (alreadyExists) {
+          throw new DOMException('Track already exists.',
+              'InvalidAccessError');
+        }
+      });
+      var existingSenders = pc.getSenders();
+      origAddStream.apply(this, arguments);
+      var newSenders = pc.getSenders().filter(function(newSender) {
+        return existingSenders.indexOf(newSender) === -1;
+      });
+      this._shimmedLocalStreams[stream.id] = [stream].concat(newSenders);
+    };
+
+    var origRemoveStream = window.RTCPeerConnection.prototype.removeStream;
+    window.RTCPeerConnection.prototype.removeStream = function(stream) {
+      this._shimmedLocalStreams = this._shimmedLocalStreams || {};
+      delete this._shimmedLocalStreams[stream.id];
+      return origRemoveStream.apply(this, arguments);
+    };
+
+    var origRemoveTrack = window.RTCPeerConnection.prototype.removeTrack;
+    window.RTCPeerConnection.prototype.removeTrack = function(sender) {
+      var pc = this;
+      this._shimmedLocalStreams = this._shimmedLocalStreams || {};
+      if (sender) {
+        Object.keys(this._shimmedLocalStreams).forEach(function(streamId) {
+          var idx = pc._shimmedLocalStreams[streamId].indexOf(sender);
+          if (idx !== -1) {
+            pc._shimmedLocalStreams[streamId].splice(idx, 1);
+          }
+          if (pc._shimmedLocalStreams[streamId].length === 1) {
+            delete pc._shimmedLocalStreams[streamId];
+          }
+        });
+      }
+      return origRemoveTrack.apply(this, arguments);
+    };
+  },
+
   shimAddTrackRemoveTrack: function(window) {
     var browserDetails = utils.detectBrowser(window);
     // shim addTrack and removeTrack.
     if (window.RTCPeerConnection.prototype.addTrack &&
         browserDetails.version >= 64) {
-      return;
+      return this.shimAddTrackRemoveTrackWithNative(window);
     }
 
     // also shim pc.getLocalStreams when addTrack is shimmed
--- a/test/e2e/expectations/chrome-beta
+++ b/test/e2e/expectations/chrome-beta
@@ -28,7 +28,7 @@ PASS URL.createObjectURL shim works for video using setAttribute
 PASS dtmf RTCRtpSender.dtmf exists on audio senders
 PASS dtmf RTCRtpSender.dtmf does not exist on video senders
 PASS dtmf inserts DTMF when using addStream
-PASS dtmf inserts DTMF when using addTrack
+ERR dtmf inserts DTMF when using addTrack timeout
 PASS getStats returns a Promise
 PASS getStats resolves the Promise with a Map(like)
 PASS getUserMedia navigator.getUserMedia exists
@@ -46,7 +46,7 @@ PASS navigator.mediaDevices enumerateDevices returns some audiooutput devices
 PASS MediaStream window.MediaStream exists
 PASS MSID signals track and stream ids
 PASS track event RTCPeerConnection.prototype.ontrack exists
-ERR track event is called when setRemoteDescription adds a new track to an existing stream timeout
+PASS track event is called when setRemoteDescription adds a new track to an existing stream
 PASS track event is called by setRemoteDescription track event
 PASS track event is called by setRemoteDescription ontrack
 PASS track event the event has a track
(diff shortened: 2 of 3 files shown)


src/js/chrome/chrome_shim.js
```diff
@@ -227,12 +227,65 @@ var chromeShim = {
     }
   },
 
+  shimAddTrackRemoveTrackWithNative: function(window) {
+    // shim addTrack/removeTrack with native variants in order to make
+    // the interactions with legacy getLocalStreams behave as in other browsers.
+    window.RTCPeerConnection.prototype.getLocalStreams = function() {
+      this._shimmedLocalStreams = this._shimmedLocalStreams || [];
```
henbos: Observations:
- Native getLocalStreams() will return the correct set of streams for addStream/removeStream cases, but ignores addTrack/removeTrack cases.
- addTrack/removeTrack cases does not rely on stream.getTracks().

Based on those observations I suggest we keep track of which streams belongs to which tracks according to addTrack()/removeTrack() and define getLocalStreams() as the native's getLocalStreams() plus any additional streams that we keep track of. This can be done in (at least) two ways:

_Alternative 1:_

Shim senders to have a set of streams.

addTrack():
- Shim the resulting sender to have streams = the streams argument.

removeTrack()
- Remove the shimmed sender.

getLocalStreams()
- Get native.getLocalStreams().
- For each sender in shimmed senders, and for each stream in sender's streams, add stream to the result (avoid duplicates).

Assuming this doesn't interfere with other shimming?

_Alternative 2:_

Use maps to keep track of which track belong to which stream, e.g. have maps "stream -> LocalStream", "track -> LocalTracks". The LocalTrack knows of all streams it belongs to. The LocalStream knows of all tracks it belongs to.

addTrack()
- For each stream, get or create its LocalStream, and add track to its tracks list.
- Get or create the track's LocalTrack and for each stream, add it to the LocalTrack's streams.

removeTrack()
- Get the track's LocalTrack, for each stream in its streams, get the stream's LocalStream. For each such LocalStream, remove the track from its streams. If LocalStream.streams is now empty, unmap LocalStream.
- Unmap track's LocalTrack.

getLocalStreams()
- Get native.getLocalStreams().
- For each LocalStream, add to the resulting set of streams (avoid duplicates).

Both alternatives should make addTrack/addStream play together nicely.
What do you think?

henbos (commented):
Suggestions to solve the multiple addTrack() calls with the same stream problem in a lengthy comment below :)

src/js/chrome/chrome_shim.js
fippo: problem is that there are cases (and tests I think) where you add via addStream but remove via removeTrack. The interactions between those APIs are... both complicated and not specified :-/

henbos: If you try to mix them can we simply make them fail? I think we need to support both, but I don't think we need to care about supporting mixed usage of legacy and non-legacy APIs. WDYT?

henbos: Or just let undefined behavior happen :)

src/js/chrome/chrome_shim.js
```diff
@@ -227,12 +227,65 @@ var chromeShim = {
...
+      if (this._shimmedLocalStreams.indexOf(stream) === -1) {
+        this._shimmedLocalStreams.push(stream);
+      }
+      return origAddTrack.apply(this, arguments);
+    };
+
+    var origAddStream = window.RTCPeerConnection.prototype.addStream;
+    window.RTCPeerConnection.prototype.addStream = function(stream) {
+      var pc = this;
+      this._shimmedLocalStreams = this._shimmedLocalStreams || [];
```
fippo: i wonder if we should bother or take the compat hit... its in an error case so not very relevant

src/js/chrome/chrome_shim.js
```diff
@@ -227,12 +227,65 @@ var chromeShim = {
...
+
+      stream.getTracks().forEach(function(track) {
+        var alreadyExists = pc.getSenders().find(function(s) {
```
fippo: i wonder if we should bother or take the compat hit... its in an error case so not very relevant

src/js/chrome/chrome_shim.js
fippo: one workaround here would be to keep a mapping track -> streams that is filled by addTrack/addStream and remove the stream when there is no track associated with it any longer

fippo: done in the new commit, lets see how that goes

fippo:
this looks pretty good actually. The test failures are caused by M64 becoming beta so we need to update those expectations.

src/js/chrome/chrome_shim.js
```diff
@@ -228,12 +228,83 @@ module.exports = {
     }
   },
 
+  shimAddTrackRemoveTrackWithNative: function(window) {
+    // shim addTrack/removeTrack with native variants in order to make
+    // the interactions with legacy getLocalStreams behave as in other browsers.
+    // Keeps a mapping stream.id => [stream, track id...]
```
henbos: Instead of an array, can we use a dictionary { stream, trackIds }? ".stream" is more readable than "[0]"

Or a class if we want helper methods like add/removeTrackIds to guard against duplicates or check if all tracks have been removed.

src/js/chrome/chrome_shim.js
```diff
@@ -228,12 +228,83 @@ module.exports = {
...
+          return s.track === track;
+        });
+        if (alreadyExists) {
+          throw new DOMException('Track already exists.',
+              'InvalidAccessError');
+        }
+      });
+      if (!this._shimmedLocalStreams[stream.id]) {
+        this._shimmedLocalStreams[stream.id] =
+            [stream].concat(stream.getTracks());
```
henbos: Should we watch out for duplicate tracks in case addTrack/addStream mix?

src/js/chrome/chrome_shim.js
```diff
@@ -228,12 +228,83 @@ module.exports = {
...
+    };
+
+    var origAddTrack = window.RTCPeerConnection.prototype.addTrack;
+    window.RTCPeerConnection.prototype.addTrack = function(track, stream) {
+      this._shimmedLocalStreams = this._shimmedLocalStreams || {};
+
+      if (!this._shimmedLocalStreams[stream.id]) {
+        this._shimmedLocalStreams[stream.id] = [stream, track];
+      } else if (this._shimmedLocalStreams[stream.id].indexOf(track) === -1) {
+        this._shimmedLocalStreams[stream.id].push(track);
```
henbos: Should we watch out for duplicate tracks in case addTrack/addStream mix?

src/js/chrome/chrome_shim.js
```diff
@@ -228,12 +228,83 @@ module.exports = {
...
+        this._shimmedLocalStreams[stream.id] =
+            [stream].concat(stream.getTracks());
+      }
+      return origAddStream.apply(this, arguments);
+    };
+
+    var origRemoveStream = window.RTCPeerConnection.prototype.removeStream;
+    window.RTCPeerConnection.prototype.removeStream = function(stream) {
+      this._shimmedLocalStreams = this._shimmedLocalStreams || {};
+      delete this._shimmedLocalStreams[stream.id];
```
henbos, line 284: Should we remove trackIds, and only delete the entry if track IDs are empty? E.g. if you addStream(stream) and addTrack(otherTrack, stream) and then removeStream(stream) you would still have the addTrack-sender with the stream. Or is guarding against addTrack/addStream mixes a lost cause?

henbos (commented):
Looking good, some comments.

src/js/chrome/chrome_shim.js
henbos: How about a test for this?

henbos:
On second though, stream -> { stream, trackIds } is problematic if sender.replaceTrack is supported, because which trackIds are associated with the stream can change. It is also irrelevant to the question "which streams are getLocalStreams()?"

Might I suggest stream -> { stream, isNativeLocalStream, senders }?
addStream-streams get isNativeLocalStream = true. addTrack-streams get an associated sender. Only when isNativeLocalStream = false (removeStream called or addStream was never called for that stream in the first place) and senders.length == 0 do we delete the entry.

This will not break if sender.replaceTrack is called because it doesn't matter which track a stream is associated with.
This should also be safe in combination with stream.addTrack, though I have not thought about that.

The alternative is adding _stream to senders and doing "for each shimmed senders for each stream".

src/js/chrome/chrome_shim.js
fippo: i'm trying to keep this simple. Unfortunately javascript objects only allow string keys, otherwise stream => trackIds would have worked

src/js/chrome/chrome_shim.js
fippo: we do that here. addTrack will throw so we just have to make sure our bookkeeping doesn't take them into account

src/js/chrome/chrome_shim.js
fippo: duplicates should cause a throw above

src/js/chrome/chrome_shim.js
fippo: hm. that is an interesting question. I would expect removeStream(stream) to remove *all* tracks of the stream, regardless how they were added. Firefox doesn't do removeStream natively but we shimmed it in a similar way.

src/js/chrome/chrome_shim.js
```diff
@@ -228,12 +228,83 @@ module.exports = {
...
+      }
+      return origAddTrack.apply(this, arguments);
+    };
+
+    var origAddStream = window.RTCPeerConnection.prototype.addStream;
+    window.RTCPeerConnection.prototype.addStream = function(stream) {
+      var pc = this;
+      this._shimmedLocalStreams = this._shimmedLocalStreams || {};
+
+      stream.getTracks().forEach(function(track) {
```
fippo, line 264: I think this is done because chrome does not behave like that. Do you have a bug tracking that @henbos? If you don't want to bother natively because mixed use is hopeless that is ok too ;-)

henbos: I'm OK with not bothering with mixed use cases. Will you do "streams -> { stream, senders }"?

fippo: yes, currently testing if this works as I want :-)

fippo: updated (and inserted as a separate commit, can squash before merging)

src/js/chrome/chrome_shim.js
```diff
@@ -228,12 +228,85 @@ module.exports = {
...
+        return pc._shimmedLocalStreams[streamId][0];
+      });
+    };
+
+    var origAddTrack = window.RTCPeerConnection.prototype.addTrack;
+    window.RTCPeerConnection.prototype.addTrack = function(track, stream) {
+      this._shimmedLocalStreams = this._shimmedLocalStreams || {};
+
+      var sender = origAddTrack.apply(this, arguments);
+      if (!this._shimmedLocalStreams[stream.id]) {
```
henbos: The stream can be undefined in the pc.addTrack(track); case, you should guard against that.

henbos (approved):
LGTM % comment

src/js/chrome/chrome_shim.js
fippo: fixed (and squashed) -- if there is no stream, this will skip this block

PR #734, opened 2017-12-13
fix style nits

fippo:
see individual commits


KaptenJansson (approved):
Thanks, nice cleanup!

PR #736, opened 2017-12-13
test: check for navigator.getUserMedia existence before wrapping it

fippo:
since it is a legacy API that does not exist in Firefox


fippo:
@jan-ivar PTAL :-)

jan-ivar (approved):

PR #737, opened 2017-12-17
safari: dont assume transceiver.setDirection exists

fippo:
it was replaced wіth a setter in the spec

@youennf can you take a look? I am making assumptions about how you'll make the transition, maybe there is a better way to determine that .direction is set-able

(test failures are due to M64 with native addTrack being stable, unrelated to this change)


fippo:
LGTM from @youennf [here](https://github.com/webrtc/adapter/commit/16d7b6fd0c67aff85a0e9bbb0f7edc05599d45c4#commitcomment-26395487) (now squashed so doesn't show up)

KaptenJansson (approved):

PR #749, opened 2018-01-19
utils: change detectBrowser to not rely on webkitRTCPeerConnection

fippo:
safer change than #722 as it still relies on feature detection via
prefixes but does not take into account unpublished safari versions.


fippo:
big advantage: this can fly as patch version bump ;-)

KaptenJansson (approved):

PR #758, opened 2018-01-30
firefox: dont shim max-message-size when RTCPeerConnection is not defined

### LINKED ISSUE #757: Firefox RTCPeerConnection undefined on shimMaxMessageSize
Firefox 58

**Description**
When you try to use the webrtc-adapter in version 6.1.0 with `media.peerconnection.enabled ==  false` (in the about:config panel), the javascript throws and exception:

```
TypeError: window.RTCPeerConnection is undefined
Stack trace:
shimMaxMessageSize@webpack-internal:///2410:122:9
module.exports@webpack-internal:///2401:99:7
```

**Steps to reproduce**
* Go to 'about:config' url in firefox, and disable the option `media.peerconnection.enabled`.
* Load this html:
```
<html>
    <head>
        <script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/webrtc-adapter/6.1.0/adapter.min.js"></script>
    </head>
    <body>
    </body>
</html>
```
* Open the console too see the exception (if you don't see the exception, reload the page).

**Expected results**
No exception thrown.

**Actual results**
Exception thrown.

fippo:
fixes #757

### DIFF
--- a/src/js/common_shim.js
+++ b/src/js/common_shim.js
@@ -114,7 +114,7 @@ module.exports = {
   },
 
   shimMaxMessageSize: function(window) {
-    if (window.RTCSctpTransport) {
+    if (window.RTCSctpTransport || !window.RTCPeerConnection) {
       return;
     }
     var browserDetails = utils.detectBrowser(window);


fippo:
chrome stable test failure is just version change chore (#759). Firefox beta fails due to transceivers.
Firefox stable breakage is odd but doesn't seem critical... @lgrahl can you take a look?

KaptenJansson (approved):

lgrahl:
Stable "broke" because FF 58 is now stable and it was FF 57 formerly. So, the expectations will have to be updated every once in a while (as long as they all pass). Maybe we want to change this for the future but I don't have a good solution so far.

fippo:
will update, thanks. We have the problem of updating like this in other expectations too so no worries.

fippo:
uh wait. You wrote the version into the expectation? :-/

lgrahl:
Yeah. Problematic?

fippo:
lots of chore. Will take a look and see if there is a better way.

Does the main change (trivial) look good to you btw?

lgrahl:
I added the version number, so people can test it with specific browser versions. But be aware this is a version comparator, so for example FF>=58 will apply to anything above FF 58 as long as the expected results don't change. It only requires a change now because expected results between FF 57 and FF 58 are different. (Sorry if this description is more confusing than helping...)

> Does the main change (trivial) look good to you btw?

Sure.

PR #762, opened 2018-01-31
Added correct direction for new firefox 59 transceiver implementation

kekkokk:
**Description**
The new Firefox 59 (currently beta) implements transceivers.
This means that the legacy offerToReceiveAudio and offerToReceive video no longer works and the result is an always sendrecv peerconnection.

**Purpose**
make legacy parameters works


fippo:
I hope Firefox will fix this before release so i'd rather not merge it. Lets keep this open/on the table until we know more though.

fippo:
https://bugzilla.mozilla.org/show_bug.cgi?id=1425618 -- fixed in FF59 so we don't need this (phew!)

kekkokk:
nice! have a good day!

PR #766, opened 2018-02-01
Put all github markdown in .github folder

KaptenJansson:
**Description**
Found this in https://github.com/blog/2111-issue-and-pull-request-templates
**Purpose**
Cleanup root folder

@fippo


fippo:
use git mv instead? But much in favor of doing this!

KaptenJansson:
git mv was a lot nicer ;) Done.

PR #767, opened 2018-02-01
Remove outdated test instructions

KaptenJansson:
**Description**
@fippo 

**Purpose**


fippo (approved):
missing that during cleanup, thanks for taking the trash out!

PR #777, opened 2018-02-16
Stop shimSendThrowTypeError from throwing error in Firefox with peerConnection disabled

DMarby:
**Description**
Related to https://github.com/webrtc/adapter/issues/757. When using version 6.1.1 of this library in Firefox with `media.peerconnection.enabled.` set to false, it will throw an error at shimSendThrowTypeError.

**Purpose**
This fixes the library in Firefox.

### DIFF
--- a/src/js/common_shim.js
+++ b/src/js/common_shim.js
@@ -245,6 +245,10 @@ module.exports = {
   },
 
   shimSendThrowTypeError: function(window) {
+    if (!window.RTCPeerConnection) {
+      return;
+    }
+
     // Note: Although Firefox >= 57 has a native implementation, the maximum
     //       message size can be reset for all data channels at a later stage.
     //       See: https://bugzilla.mozilla.org/show_bug.cgi?id=1426831


fippo (approved):

fippo:
note: this should also check for RTCPeerConnection.createDataChannel, otherwise it will add RTCPeerConnection.createDataChannel in Edge. Can do after merge.

lgrahl:
It's explicitly not applied to Edge right now. But yeah, it would be cleaner to check for the prototype.

fippo:
@lgrahl it is actually: https://github.com/webrtc/adapter/blob/ded85225d9ea5f571f6031ed100434583e72206f/src/js/adapter_factory.js#L118
(i noticed the munge-sdp sample breaking...)

lgrahl:
Interesting. I remember me explicitly not adding it to Edge, so I'm a bit puzzled. But no worries, let's just fix it then.

KaptenJansson (approved):

PR #779, opened 2018-02-21
chrome: override legacy createDTMFSender

fippo:
@alvestrand @foolip PTAL. Not sure this is a good idea but... ;-)

Still undecided on the error to throw when there is no sender for the track. Chrome seems to throw a SyntaxError or a NotSupportedError (the latter for adding video tracks?!)

### DIFF
--- a/src/js/chrome/chrome_shim.js
+++ b/src/js/chrome/chrome_shim.js
@@ -188,6 +188,25 @@ module.exports = {
           return this._dtmf;
         }
       });
+    } else if (typeof window === 'object' && window.RTCPeerConnection &&
+        ('getSenders' in window.RTCPeerConnection.prototype) &&
+        (window.RTCRtpSender && 'dtmf' in window.RTCRtpSender.prototype)) {
+      // override legacy createDTMFSender.
+      window.RTCPeerConnection.prototype.createDTMFSender = function(track) {
+        if (track.kind !== 'audio') {
+          throw new DOMException('Track is not an audio track.',
+              'TypeError');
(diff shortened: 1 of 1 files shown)


fippo:
ironicanlly this times out with an unrelated dtmf error in Firefox.. :joy:

foolip:
I guess my question is why? Isn't it better if the thing disappears once it disappears in browsers?

fippo:
this drives the usage you/chrome sees down, making it easier to remove it.
Also adapter is a bit more nagging than Chrome :-)

foolip:
The usage we're seeing isn't very problematic:
https://www.chromestatus.com/metrics/feature/timeline/popularity/1642

I'll defer to @alvestrand, but I don't think this seems that useful for the purpose of removing, and when removing we should recommend a standalone snippet, not using all of adapter.js, unless people already are.

fippo:
well, doing it inside adapter in addition to recommending such a snippet has greater reach. Its pretty cheap to do here and removing it just means another major version bump in the future -- but that is something developers can decide for themselves, not chrome forcing them to.

alvestrand:
So this will in practice affect UMA counts, not functionality, right?
The other difference to the shim that is in the chrome code is that dtmfsender.track attribute will not be set if you do it this way - we left that in as a deliberate difference ("track" attribute is nonstandard).

KaptenJansson:
So what should be done here?

fippo:
probably nothing. lets see how those metrics turn out...

PR #783, opened 2018-03-04
Adding adapter support for iOS WKWebView

zaxy78:
**Description**
When using adapter.js inside an iOS In-App Browser (WKWebView) - adpater classifies it as a "unsupported browser" even though most of the WebRTC APIs **are** supported. 
"detect_browser" checks only for `getUserMedia()` which is exactly the API that is not supported.

This fix adds Web Views on iPhone & iPad as a `safari` supported browser

**Purpose**
Despite the fact that getUserMedia() is still not supported on this Web View version of Safari on iOS,
adapter.js can and does help with the other APIs when running on WKWebView and I see no reason supporting it. 
I used adapter.js myself for some RTCPeerConnection shim work inside the Web View, 
but only after I added it as a supported browser.

Offering back this fix to the community. 
Added a Unit Test too.

see issue #784

### DIFF
--- a/src/js/utils.js
+++ b/src/js/utils.js
@@ -159,14 +159,8 @@ module.exports = {
       result.browser = 'edge';
       result.version = extractVersion(navigator.userAgent,
           /Edge\/(\d+).(\d+)$/, 2);
-    } else if (navigator.mediaDevices &&
+    } else if (window.RTCPeerConnection &&
         navigator.userAgent.match(/AppleWebKit\/(\d+)\./)) { // Safari.
-      result.browser = 'safari';
-      result.version = extractVersion(navigator.userAgent,
-          /AppleWebKit\/(\d+)\./, 1);
-    } else if (navigator.userAgent.match(/AppleWebKit\/(\d+)\./) &&
-        (navigator.userAgent.match(/iPhone|iPad/))) {
-      // WKWebView on iOS; Where mediaDevices is still unsupported
       result.browser = 'safari';
       result.version = extractVersion(navigator.userAgent,
           /AppleWebKit\/(\d+)\./, 1);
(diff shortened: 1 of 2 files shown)


zaxy78:
This relates to changes in #749 (and its deserted alternative #722 )

Basing detection on `getUserMedia` _excluded_ the iOS WebView, which does support `RTCPeerConnection` and works just fine for _receiving_ streams.. 

Instead of changing the detection mechanism, I just allowed support `iPhone` and `iPad` browsers even if they don't have `getUserMedia` - WKWebView and SFSafariViewController

If there is any reason not to support it, please mention... otherwise - Adapter.js helped me 😃

src/js/utils.js
```diff
@@ -164,6 +164,12 @@ module.exports = {
       result.browser = 'safari';
       result.version = extractVersion(navigator.userAgent,
           /AppleWebKit\/(\d+)\./, 1);
+    } else if (navigator.userAgent.match(/AppleWebKit\/(\d+)\./) &&
```
fippo: can you merge that with the block above and make that depend on the presence of RTCPeerConnection in window instead of getUserMedia?

test/unit/detectBrowser.js
```diff
@@ -61,4 +61,15 @@ describe('detectBrowser', () => {
     expect(browserDetails.browser).to.equal('safari');
     expect(browserDetails.version).to.equal(604);
   });
+
+  it('detects iOSWKWebView if navigator.mediaDevices doens\'t exist', () => {
```
fippo: nit: detects safari

fippo (approved):
could ship as-is but i'd like to see some small changes. Thanks!

KaptenJansson (approved):

Issue #785, opened 2018-03-06
Safari shim - ignores offer options when sent with bits values


zaxy78:
- [x] I have provided steps to reproduce (e.g. a link to a [jsfiddle](https://jsfiddle.net/))
- [x] I have provided browser name, version and adapter.js version
- [x] This issue only happens when adapter.js is used

## Versions affected
Safari 11 

## Description

while converting `options` object to respective audio/video `addTransceiver` calls, the shim only detects objects with `true` or `false` values, even though legacy api used to support `0` and `1` values as well.

## Steps to reproduce

Send this object as an offer options:
```
var offerOptions = {
    offerToReceiveAudio: 0,
    offerToReceiveVideo: 1
};
```

## Expected results
treat the options the same as if it would be 
```
var offerOptions = {
    offerToReceiveAudio: false,
    offerToReceiveVideo: true
};
```
detect and call `addTransceiver` with 'audio' and/or 'video' respectively. 

## Actual results
`addTransceiver` is not called. `offerToReceive` is being ignored.

zaxy78:
Working on a fix, will upload a PR soon

PR #786, opened 2018-03-07
Safari bits offer options

zaxy78:
Fix for #785 + unit tests (legacy createOffer shim fix)

### DIFF
--- a/src/js/safari/safari_shim.js
+++ b/src/js/safari/safari_shim.js
@@ -255,16 +255,11 @@ module.exports = {
     window.RTCPeerConnection.prototype.createOffer = function(offerOptions) {
       var pc = this;
       if (offerOptions) {
+        pc.offerToReceiveAudio = !!pc.offerToReceiveAudio; // support bit value
         var audioTransceiver = pc.getTransceivers().find(function(transceiver) {
           return transceiver.sender.track &&
               transceiver.sender.track.kind === 'audio';
         });
-        var isBit = offerOptions.offerToReceiveAudio;
-        if (Number.isInteger(isBit)) {
-          offerOptions.offerToReceiveAudio = (isBit === 0) ? false : true;
-          utils.deprecated('Integer values (0,1) to offerToReceiveAudio',
-          'boolean values (true/false)');
-        }
         if (offerOptions.offerToReceiveAudio === false && audioTransceiver) {
           if (audioTransceiver.direction === 'sendrecv') {
             if (audioTransceiver.setDirection) {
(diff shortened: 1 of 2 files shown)


src/js/safari/safari_shim.js
```diff
@@ -259,6 +259,12 @@ module.exports = {
           return transceiver.sender.track &&
               transceiver.sender.track.kind === 'audio';
         });
+        var isBit = offerOptions.offerToReceiveAudio;
```
KaptenJansson: Since you are using it in two places and its quite a few lines, perhaps pull it out to a helper function instead?

KaptenJansson (changes requested):

src/js/safari/safari_shim.js
```diff
@@ -278,6 +279,7 @@ module.exports = {
           pc.addTransceiver('audio');
         }
 
+        pc.offerToReceiveVideo = !!pc.offerToReceiveVideo; // support bit value
```
zaxy78: found a much simpler way to fix this. 
but gave up the 'deprecation' message I've put there.

src/js/safari/safari_shim.js
```diff
@@ -278,6 +280,8 @@ module.exports = {
           pc.addTransceiver('audio');
         }
 
+        // support bit value
+        offerOptions.offerToReceiveVideo = !!offerOptions.offerToReceiveVideo;
```
zaxy78: @KaptenJansson - I found a much simpler way to fix this.
but gave up the 'deprecation' message I've put there.

fippo: if offerToReceiveVideo is unset the !! will make it false which may change the direction. The simplest way might be to check for "this isn't undefined"

zaxy78:
@KaptenJansson looks better?

src/js/safari/safari_shim.js
```diff
@@ -278,6 +282,11 @@ module.exports = {
           pc.addTransceiver('audio');
         }
 
+
+        if (offerOptions.offerToReceiveVideo !== undefined) {
+          // support bit values
+          offerOptions.offerToReceiveVideo = !!offerOptions.offerToReceiveVideo;
+        }
```
zaxy78: @KaptenJansson looks better? I think It's ready for merge

src/js/safari/safari_shim.js
```diff
@@ -255,6 +255,10 @@ module.exports = {
     window.RTCPeerConnection.prototype.createOffer = function(offerOptions) {
       var pc = this;
       if (offerOptions) {
+        if (offerOptions.offerToReceiveAudio !== undefined) {
```
KaptenJansson: I would prefer typeof instead even though its not used everywhere in this repo.

`if (typeof offerOptions.offerToReceiveVideo !== 'undefined') {
.... }`

KaptenJansson (changes requested):

src/js/safari/safari_shim.js
zaxy78: @KaptenJansson  done

zaxy78:
Fixed and Ready to merge.

Also, @KaptenJansson, when do you plan to have the next release?

I think after #783, we can add "Support for iOS WKWebView" and make a new release

zaxy78:
@KaptenJansson, thanks for the release publish. any reason not to merge this?
I fixed your comments..

KaptenJansson (approved):

PR #788, opened 2018-03-08
test: update removetrack-twice expectation for chrome

fippo:
chrome changed behaviour in removeTrack due to a bug @henbos fixed.
The test is somewhat silly and doesn't really work in the "stage 2"
model which cleans up senders. We can change the expectation with
unified plan where this makes more sense.

(needs to be trickled down into beta and stable over time)


henbos (approved):
LGTM, thanks

PR #789, opened 2018-03-08
test: improve DTMF test to wait for canInsertDTMF

fippo:
using the pattern from https://github.com/w3c/web-platform-tests/blob/master/webrtc/RTCDTMFSender-helper.js

### DIFF
--- a/test/e2e/dtmf.js
+++ b/test/e2e/dtmf.js
@@ -104,7 +104,22 @@ describe('dtmf', () => {
               pc2.iceConnectionState === 'completed') && resolve());
       })
       .then(() => {
-        let sender = pc1.getSenders().find(s => s.track.kind === 'audio');
+        if (!(window.RTCDTMFSender &&
+            'canInsertDTMF' in window.RTCDTMFSender.prototype)) {
+          return;
+        }
+        return new Promise((resolve) => {
+          setTimeout(function canInsert() {
+            const sender = pc1.getSenders().find(s => s.track.kind === 'audio');
+            if (sender.dtmf.canInsertDTMF) {
+              return resolve();
+            }
+            setTimeout(canInsert, 10);
+          }, 0);
+        });
+      })
+      .then(() => {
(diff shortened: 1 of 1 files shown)


fippo:
@alvestrand can you take a look? Also happy to contribute the small style improvements back to wpt ;-)

fippo:
and ofc the flakyness in chrome stable triggers :joy:

fippo:
https://github.com/w3c/webrtc-pc/issues/1619#issuecomment-366963505 suggests this should not be necessary but practically speaking it makes sense.

test/e2e/dtmf.js
```diff
@@ -104,7 +104,22 @@ describe('dtmf', () => {
...
+            'canInsertDTMF' in window.RTCDTMFSender.prototype)) {
+          return;
+        }
+        return new Promise((resolve) => {
+          setTimeout(function canInsert() {
+            const sender = pc1.getSenders().find(s => s.track.kind === 'audio');
+            if (sender.dtmf.canInsertDTMF) {
+              return resolve();
+            }
+            setTimeout(canInsert, 10);
```
jan-ivar, line 117: This is a terrible API IMHO. `canInsertDTMF` should be for determining whether DTMF is possible in the current configuration, not for figuring out ***when*** to start to send DTMF.

I think we need to fix the spec to specify when this boolean can be expected to be true, e.g. in the `"connected"` event (from https://github.com/w3c/webrtc-pc/issues/1619#issuecomment-366963505).

fippo: it is but this test has been flaky forever. I still think there is a chrome bug here but we'll see

alvestrand: The code turned out to eat DTMF sent before connection was in ready state, which was an even more terrible API.
The thing that SHOULD work is to not look until connection is established - but even waiting for state changes had issues.

alvestrand (approved):

PR #791, opened 2018-03-09
Update deps

KaptenJansson:
**Description**

**Purpose**


fippo (approved):

fippo:
updating the deps should be safe -- unless you want to try npm update --no-save

PR #793, opened 2018-03-12
edge: add RTCDTMFSender alias

fippo:
**Description**
adds RTCDTMFSender if it does not exist. Its a simple alias to
RTCDtmfSender as described in http://draft.ortc.org/#rtcdtmfsender2*

**Purpose**
Less things to worry about ;-)


KaptenJansson (approved):
The devil is in the details....

PR #794, opened 2018-03-12
common: check RTCIceCandidate exists before trying to shim

fippo:
fixes an exception in firefox with peerconnection disabled

Also adds a very, very basic unit test to detect issues like this in the future.


KaptenJansson (approved):
Nice!

PR #807, opened 2018-03-29
Update max-message-size for FF>=60

lgrahl:
The [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1426831) has been fixed for FF 60 and we can send large messages again.

Also updates the test expectations.

### DIFF
--- a/test/e2e/maxMessageSize.js
+++ b/test/e2e/maxMessageSize.js
@@ -182,6 +182,8 @@ describe('maxMessageSize', () => {
 
     // Map specific browser versions to a test case.
     // You can use the following version comparators: '<=', '>=' and '=='.
+    // Multiple comparators can be chained with ',' which is equivalent to the
+    // logical AND.
     //
     // The innermost array contains the other browser's version, the expected
     // maximum message size and the offer SDP.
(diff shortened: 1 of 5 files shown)


lgrahl:
Failing tests seem unrelated.

fippo:
*sigh* silly googlebot :-/
I signed it (but you probably looked at the wrong email address)

lgrahl:
Rebased but googlebot didn't notice. :wink: Nothing I can do at this point.

lgrahl:
Can we get this merged?

test/e2e/maxMessageSize.js
```diff
@@ -182,6 +182,8 @@ describe('maxMessageSize', () => {
 
     // Map specific browser versions to a test case.
     // You can use the following version comparators: '<=', '>=' and '=='.
+    // Multiple comparators can be chained with ',' which is equivalent to the
+    // logical AND.
```
alvestrand, line 186: Nit: I'd argue that it's a logical OR, because it makes the whole rule match if either the left hand side or the right hand side of the expression match.

lgrahl: Mhh, it was my intention to make it a logical AND and both expressions need to match. Unless I'm mistaken, the `.every` should take care of that?

```js
browserVersions.split(',').every(...)
```

alvestrand: Seems I didn't read the one usage carefully enough, it's      'firefox>=58,firefox<=59': [ which will indeed be only meaningful with "and". but comma is so frequently used for "inclusion" in selection lists - since you're defining your own language here anyway, why not use "&"?

lgrahl: I roughly borrowed that from Python's version specifiers (when defining a dependency such as `aiohttp>=1.2.3,<2`). I don't have a strong opinion on what character to use. :slightly_smiling_face:

alvestrand: wottehell, if Python does it there's no hope for consistency anyway. Approved.

alvestrand (approved):

PR #809, opened 2018-04-09
chrome/firefox: add shim for sender/receiver getStats

fippo:
adds a naive shim for sender/receiver getStats which returns
the full, unfilterered peerconnection stats

### DIFF
--- a/src/js/chrome/chrome_shim.js
+++ b/src/js/chrome/chrome_shim.js
@@ -11,6 +11,47 @@
 var utils = require('../utils.js');
 var logging = utils.log;
 
+/* iterates the stats graph recursively. */
+function walkStats(stats, base, resultSet) {
+  if (!base || resultSet.has(base.id)) {
+    return;
+  }
+  resultSet.set(base.id, base);
+  Object.keys(base).forEach(function(name) {
+    if (name.endsWith('Id')) {
+      walkStats(stats, stats.get(base[name]), resultSet);
+    } else if (name.endsWith('Ids')) {
+      base[name].forEach(function(id) {
+        walkStats(stats, stats.get(id), resultSet);
+      });
+    }
+  });
+}
+
+/* filter getStats for a sender/receiver track. */
+function filterStats(result, track, outbound) {
+  var streamStatsType = outbound ? 'outbound-rtp' : 'inbound-rtp';
+  var filteredResult = new Map();
+  if (track === null) {
+    return filteredResult;
+  }
+  var trackStats = [];
+  result.forEach(function(value) {
+    if (value.type === 'track' &&
+        value.trackIdentifier === track.id) {
+      trackStats.push(value);
+    }
+  });
+  trackStats.forEach(function(trackStat) {
+    result.forEach(function(stats) {
+      if (stats.type === streamStatsType && stats.trackId === trackStat.id) {
+        walkStats(result, stats, filteredResult);
+      }
+    });
+  });
+  return filteredResult;
+}
+
 module.exports = {
   shimGetUserMedia: require('./getusermedia'),
   shimMediaStream: function(window) {
@@ -189,6 +230,77 @@ module.exports = {
         }
       });
     }
+  },
+
+  shimSenderGetStats: function(window) {
+    if (!(typeof window === 'object' && window.RTCPeerConnection &&
+        window.RTCRtpSender)) {
+      return;
+    }
+    if (window.RTCRtpSender && 'getStats' in window.RTCRtpSender.prototype) {
+      return;
+    }
+    var origGetSenders = window.RTCPeerConnection.prototype.getSenders;
+    if (origGetSenders) {
+      window.RTCPeerConnection.prototype.getSenders = function() {
+        var pc = this;
+        var senders = origGetSenders.apply(pc, []);
+        senders.forEach(function(sender) {
+          sender._pc = pc;
+        });
+        return senders;
+      };
+    }
+
+    var origAddTrack = window.RTCPeerConnection.prototype.addTrack;
+    if (origAddTrack) {
+      window.RTCPeerConnection.prototype.addTrack = function() {
+        var sender = origAddTrack.apply(this, arguments);
+        sender._pc = this;
+        return sender;
+      };
+    }
+    window.RTCRtpSender.prototype.getStats = function() {
+      var sender = this;
+      return this._pc.getStats().then(function(result) {
+        /* Note: this will include stats of all senders that
+         *   send a track with the same id as sender.track as
+         *   it is not possible to identify the RTCRtpSender.
+         */
+        return filterStats(result, sender.track, true);
+      });
+    };
+  },
+
+  shimReceiverGetStats: function(window) {
+    if (!(typeof window === 'object' && window.RTCPeerConnection &&
+        window.RTCRtpSender)) {
+      return;
+    }
+    if (window.RTCRtpSender && 'getStats' in window.RTCRtpReceiver.prototype) {
+      return;
+    }
+    var origGetReceivers = window.RTCPeerConnection.prototype.getReceivers;
+    if (origGetReceivers) {
+      window.RTCPeerConnection.prototype.getReceivers = function() {
+        var pc = this;
+        var receivers = origGetReceivers.apply(pc, []);
+        receivers.forEach(function(receiver) {
+          receiver._pc = pc;
+        });
+        return receivers;
+      };
+    }
+    utils.wrapPeerConnectionEvent(window, 'track', function(e) {
+      e.receiver._pc = e.srcElement;
+      return e;
+    });
+    window.RTCRtpReceiver.prototype.getStats = function() {
+      var receiver = this;
+      return this._pc.getStats().then(function(result) {
+        return filterStats(result, receiver.track, false);
+      });
+    };
   },
 
   shimSourceObject: function(window) {
--- a/src/js/adapter_factory.js
+++ b/src/js/adapter_factory.js
@@ -73,6 +73,8 @@ module.exports = function(dependencies, opts) {
       chromeShim.shimOnTrack(window);
       chromeShim.shimAddTrackRemoveTrack(window);
       chromeShim.shimGetSendersWithDtmf(window);
+      chromeShim.shimSenderGetStats(window);
+      chromeShim.shimReceiverGetStats(window);
 
       commonShim.shimRTCIceCandidate(window);
       commonShim.shimMaxMessageSize(window);
--- a/src/js/adapter_factory.js
+++ b/src/js/adapter_factory.js
@@ -96,6 +96,8 @@ module.exports = function(dependencies, opts) {
       firefoxShim.shimPeerConnection(window);
       firefoxShim.shimOnTrack(window);
       firefoxShim.shimRemoveStream(window);
+      firefoxShim.shimSenderGetStats(window);
+      firefoxShim.shimReceiverGetStats(window);
 
       commonShim.shimRTCIceCandidate(window);
       commonShim.shimMaxMessageSize(window);
--- a/src/js/firefox/firefox_shim.js
+++ b/src/js/firefox/firefox_shim.js
@@ -199,6 +199,68 @@ module.exports = {
     };
   },
 
+  shimSenderGetStats: function(window) {
+    if (!(typeof window === 'object' && window.RTCPeerConnection &&
+        window.RTCRtpSender)) {
+      return;
+    }
+    if (window.RTCRtpSender && 'getStats' in window.RTCRtpSender.prototype) {
+      return;
+    }
+    var origGetSenders = window.RTCPeerConnection.prototype.getSenders;
+    if (origGetSenders) {
+      window.RTCPeerConnection.prototype.getSenders = function() {
+        var pc = this;
+        var senders = origGetSenders.apply(pc, []);
+        senders.forEach(function(sender) {
+          sender._pc = pc;
+        });
+        return senders;
+      };
+    }
+
+    var origAddTrack = window.RTCPeerConnection.prototype.addTrack;
+    if (origAddTrack) {
+      window.RTCPeerConnection.prototype.addTrack = function() {
+        var sender = origAddTrack.apply(this, arguments);
+        sender._pc = this;
+        return sender;
+      };
+    }
+    window.RTCRtpSender.prototype.getStats = function() {
+      return this.track ? this._pc.getStats(this.track) :
+          Promise.resolve(new Map());
+    };
+  },
+
+  shimReceiverGetStats: function(window) {
+    if (!(typeof window === 'object' && window.RTCPeerConnection &&
+        window.RTCRtpSender)) {
+      return;
+    }
+    if (window.RTCRtpSender && 'getStats' in window.RTCRtpReceiver.prototype) {
+      return;
+    }
+    var origGetReceivers = window.RTCPeerConnection.prototype.getReceivers;
+    if (origGetReceivers) {
+      window.RTCPeerConnection.prototype.getReceivers = function() {
+        var pc = this;
+        var receivers = origGetReceivers.apply(pc, []);
+        receivers.forEach(function(receiver) {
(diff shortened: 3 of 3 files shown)


src/js/chrome/chrome_shim.js
```diff
@@ -191,6 +191,65 @@ module.exports = {
...
+    var origAddTrack = window.RTCPeerConnection.prototype.addTrack;
+    if (origAddTrack) {
+      window.RTCPeerConnection.prototype.addTrack = function() {
+        var sender = origAddTrack.apply(this, arguments);
+        sender._pc = this;
+        return sender;
+      };
+    }
+    window.RTCRtpSender.prototype.getStats = function() {
+      return this._pc.getStats();
```
henbos: sender/receiver.getStats() is not the same as pc.getStats().

Running this...
https://jsfiddle.net/8aLysna5/19/
https://jsfiddle.net/8aLysna5/20/
It seems that Firefox (52.6.0) does not implement sender/receiver.getStats(), and the selector version that they do implement, pc.getStats(track), DOES filter out unrelated outboundrtp streams.

If Firefox's sender/receiver getStats() was the same as pc.getStats() then this might make sense, but it doesn't seem to be the case and Chrome is shipping proper support in M67.

henbos: If we want to shim this, how about performing filtering in accordance to the [stats selection algorithm](https://w3c.github.io/webrtc-pc/#dfn-stats-selection-algorithm)?
See [validateStatsGraph](https://cs.chromium.org/chromium/src/third_party/WebKit/LayoutTests/external/wpt/webrtc/RTCPeerConnection-track-stats.https.html?l=586) (validates rather than collects, but it's the same logic).

fippo: So we might have to shim for Firefox ESR too. @jan-ivar when was sender.getStats added?

Can we just shim it as pc.getStats(sender.track)?

jan-ivar: `sender.getStats` and `receiver.getStats` were implemented in [Firefox 55](https://bugzilla.mozilla.org/show_bug.cgi?id=1355220), and they're identical to `pc.getStats(transceiver.sender.track)` and `pc.getStats(transceiver.receiver.track)` respectively in our implementation (i.e. there's still work needed to get it to spec).

ESR goes to 60 on [May 9th](https://www.mozilla.org/en-US/firefox/organizations/) FWIW.

fippo:
so... if we want to do filtering, is the following correct (for the sender):
```
search for type=track with
    trackIdentifier === track.id
search for type=outbound-rtp with trackId pointing to type=track pointing to the one above
type=codec from outbound-rtp.codecІd
type=transport from outbound-rtp.transportId
type=certificate from transport.localCertificateId
type=certificate from transport.remoteCertificateId
type=candidate-pair from transport.selectedCandidatePairId
type=candidate from candidate-pair.localCandidateId
type=candidate from candidate-pair.remoteCandidateId
```
I am a bit surprised we have no forward pointer from the type=track to the type=outbound-rtp or did I miss it?

jan-ivar:
The pointer goes the other way, using [trackId](https://w3c.github.io/webrtc-stats/#dom-rtcoutboundrtpstreamstats-trackid). E.g.:
```js
let stats = await sender.getStats();
let rtpStat = [...stats.values()].find(s => s.type == "outbound-rtp");
let trackStat = stats.get(rtpStat.trackId);
```
so you could do:
```js
rtpStat = [...stats.values()].find(s => s.type == "outbound-rtp" &&
                                        s.trackId == trackStat.id);
```

jan-ivar:
Maybe the reason is simulcast?

src/js/chrome/chrome_shim.js
fippo: added a shim for Firefox ;-)

src/js/chrome/chrome_shim.js
```diff
@@ -11,6 +11,69 @@
 var utils = require('../utils.js');
 var logging = utils.log;
 
+/* filter getStats for a sender/receiver track */
+function filterStats(result, track) {
+  var filteredResult = new Map();
+  if (track === null) {
+    return filteredResult;
+  }
+  var senderStats;
```
fippo: this has the wrong name now... trackStats? senderreceiverstats?

src/js/chrome/chrome_shim.js
```diff
@@ -11,6 +11,69 @@
 var utils = require('../utils.js');
 var logging = utils.log;
 
+/* filter getStats for a sender/receiver track */
+function filterStats(result, track) {
```
henbos: By explicitly calling out all the things you interested in you are baking in a lot of assumptions about what stats are or are not there. This would be sensitive to any lower layer changes to what stats are or are not available, and would require you to do a lot more testing if you want to be confident about it. I think the correct approach is to solve this with a function that recursively explores the stats graph, looking at the member names to figure out if something is a reference or not.

If you do this, the code will be shorter and easier to verify it is doing the correct thing since it does not assume what should or shouldn't be there based on spec and implementation status. [validateStatsGraph](https://cs.chromium.org/chromium/src/third_party/WebKit/LayoutTests/external/wpt/webrtc/RTCPeerConnection-track-stats.https.html?l=586) shows how to recursively explore the stats graph based on member names ending with 'Id' (string) or 'Ids' (array of strings) and its only a few lines of code. Instead of a "visited ids" set you'd produce a new stats report.

What you have is more efficient, since you don't end up looking at a bunch of irrelevant members, but on this layer it would be hard to verify it is correct without looking up all the expected stats in a typical call and I don't think the performance difference is significant in this context.

Any reason not to do this as above?

fippo: the only assumption I am making is that if someId is set then the stats contain something for that id. This can become quite relevant, e.g. an unconnected sender will not have a selectedCandidatePairId

 Your approach is more generic and forward-looking but we are dealing with code that has a very explicit expiration date (:soon:)

henbos: I meant assumptions about what members for Ids exist (e.g. if fooId is added, your code won't include the referenced foo or any ones referenced by foo), but considering sender/receiver.getStats() is about to ship this code does not strictly need to be generic and forward-looking.

src/js/chrome/chrome_shim.js
```diff
@@ -11,6 +11,63 @@
...
+      senderStats = value;
+    }
+  });
+  if (!senderStats) {
+    return filteredResult;
+  }
+  filteredResult.set(senderStats.id, senderStats);
+
+  var rtpStream;
+  result.forEach(function(value) {
```
henbos: nit: change "value" to "stat" or "stats"

src/js/chrome/chrome_shim.js
```diff
@@ -11,6 +11,63 @@
...
+function filterStats(result, track) {
+  var filteredResult = new Map();
+  if (track === null) {
+    return filteredResult;
+  }
+  var senderStats;
+  result.forEach(function(value) {
+    if (value.type === 'track' &&
+        value.trackIdentifier === track.id) {
+      senderStats = value;
```
henbos: senderStats should be called trackStat. There are both sender and track dictionaries, only track stats are implemented though, but calling it senderStats is incorrect.

If there are two senders sending the same track this shim will be broken, and take the last one iterated, so sender1.getStats() could return stats for sender2. This is probably not something you can fix without sender stats support but I would want to handle this case nicely, i.e. if you reach this line and trackStat is already set (if there are multiple trackStats for the same track) then either reject the promise or return stats for both trackStats (do trackStats.push, and below do for each trackStat in trackStats...).

src/js/chrome/chrome_shim.js
```diff
@@ -11,6 +11,63 @@
...
+    }
+  });
+  if (!rtpStream) {
+    return filteredResult;
+  }
+
+  filteredResult.set(rtpStream.id, rtpStream);
+  if (rtpStream.codecId) {
+    filteredResult.set(rtpStream.codecId,
+        result.get(rtpStream.codecId));
```
henbos: Sound assumption: The codecId has a refrence to the transport, but that's ok, it should be the same transport as the rtpStream is also referencing.

src/js/chrome/chrome_shim.js
```diff
@@ -11,6 +11,63 @@
...
+  }
+
+  var transport = result.get(rtpStream.transportId);
+  if (transport) {
+    filteredResult.set(transport.id, transport);
+    ['localCertificateId', 'remoteCertificateId'].forEach(function(id) {
```
henbos: Unsound assumption: The certificate is not referencing anything. This is incorrect if there is a certificate chain in which case certificates can reference other certificates. Not sure if or when we generate certificate chains or how often you would want to explore those from a sender.getStats() call so maybe nobody will notice...

src/js/chrome/chrome_shim.js
```diff
@@ -11,6 +11,63 @@
...
+      if (transport[id]) {
+        filteredResult.set(transport[id], result.get(transport[id]));
+      }
+    });
+    if (transport.selectedCandidatePairId) {
+      var candidatePair = result.get(transport.selectedCandidatePairId);
```
henbos: Sound assumption: The candidate pair's transport reference is referencing the same transport that we've already explored, but that's ok, should be true.

src/js/chrome/chrome_shim.js
```diff
@@ -11,6 +11,63 @@
...
+      filteredResult.set(transport.selectedCandidatePairId,
+          candidatePair);
+      ['localCandidateId', 'remoteCandidateId'].forEach(function(id) {
+        if (candidatePair[id]) {
+          filteredResult.set(candidatePair[id], result.get(candidatePair[id]));
+        }
+      });
+    }
+  }
+  return filteredResult;
```
henbos: And ofc, sound assumption: Of all the stats objects we explored, there are no unknown fooId members that we didn't cover. This is not ok unless we know this code will be retired, which it will in M67.

src/js/chrome/chrome_shim.js
```diff
@@ -11,6 +11,63 @@
...
+
+  filteredResult.set(rtpStream.id, rtpStream);
+  if (rtpStream.codecId) {
+    filteredResult.set(rtpStream.codecId,
+        result.get(rtpStream.codecId));
+  }
+
+  var transport = result.get(rtpStream.transportId);
+  if (transport) {
+    filteredResult.set(transport.id, transport);
```
henbos: Unsound assumption: The transport is not referencing any other transports. This assumption does not hold if we have RTCP transport. The shim is broken in this case.

src/js/chrome/chrome_shim.js
```diff
@@ -191,6 +248,73 @@ module.exports = {
...
+      window.RTCPeerConnection.prototype.addTrack = function() {
+        var sender = origAddTrack.apply(this, arguments);
+        sender._pc = this;
+        return sender;
+      };
+    }
+    window.RTCRtpSender.prototype.getStats = function() {
+      var sender = this;
+      return this._pc.getStats().then(function(result) {
+        return filterStats(result, sender.track);
```
henbos: What is sender.track is null? And from previous comment: What if there are multiple senders whose track is the same?

henbos (commented):
Because you're not doing a recursive stats exploration, verifying this code makes sense involved manually looking at the stats graph (https://photos.app.goo.gl/6P0i4AV1E3szctTl1).

I found two assumptions that do not hold. Honestly I still think a recursive call is both easier to review, shorter and less prone to edge case bugs, so I still suggest you do that. Otherwise, fix the "unsound assumptions" below.

Other than that what needs to be fixed: sender.track = null case and sender1.track == sender2.track case. See comments.

src/js/chrome/chrome_shim.js
henbos: Sound assumption: We've not implemented remoteId

src/js/chrome/chrome_shim.js
fippo: if track is null this will return an empty set.

fippo:
you've convinced me that recursive is easier :-)
Tests will fail, needs rebase (and squash before merging)

src/js/chrome/chrome_shim.js
```diff
@@ -11,6 +11,51 @@
 var utils = require('../utils.js');
 var logging = utils.log;
 
+/* iterates the stats graph recursively */
+function walkStats(stats, base, resultSet) {
+  if (!base || resultSet.has(base.id)) {
+    return;
+  }
+  resultSet.set(base.id, base);
+  Object.keys(base).forEach(function(name) {
```
henbos, line 20: nit: arrow function? :)

src/js/chrome/chrome_shim.js
```diff
@@ -11,6 +11,51 @@
 var utils = require('../utils.js');
 var logging = utils.log;
 
+/* iterates the stats graph recursively */
+function walkStats(stats, base, resultSet) {
```
henbos: Yay! :D

src/js/chrome/chrome_shim.js
```diff
@@ -11,6 +11,51 @@
...
+  }
+  var trackStats = [];
+  result.forEach(function(value) {
+    if (value.type === 'track' &&
+        value.trackIdentifier === track.id) {
+      trackStats.push(value);
+    }
+  });
+  trackStats.forEach(function(trackStat) {
+    filteredResult.set(trackStat.id, trackStat);
```
henbos: Hmm. Since you walkStats() with rtpStreamStats, it will reference the track stats and include it anyway, so there shouldn't be a need to do this. The danger with adding it to the result already means it will be excluded from any recursive walkStats(). So if there is something that is only referenced (directly or indirectly) by the track stats that you otherwise can't get to from the rtpStreamStats, those objects will not be included in the result.

Either just remove this line, or if you know of a case where we have a track stats but not a corresponding rtpStreamStats (is that the case before SLD/SRD or did we fix that?), feel free to perform walkStats() on the trackStat.

src/js/chrome/chrome_shim.js
```diff
@@ -11,6 +11,51 @@
...
+
+    var rtpStreamStats;
+    result.forEach(function(stats) {
+      if (stats.type === streamStatsType && stats.trackId === trackStat.id) {
+        rtpStreamStats = stats;
```
henbos: In theory, there could be multiple rtp streams per track (simulcast), so do walkStats() here instead in case there are multiple stats that meet this criterion.

src/js/chrome/chrome_shim.js
```diff
@@ -11,6 +11,51 @@
...
+  if (track === null) {
+    return filteredResult;
+  }
+  var trackStats = [];
+  result.forEach(function(value) {
+    if (value.type === 'track' &&
+        value.trackIdentifier === track.id) {
+      trackStats.push(value);
+    }
+  });
```
henbos, line 44: If there are multiple senders with the same track, this will include results for all senders. In the RTCPeerConnection.getStats(track) case this should per-spec be rejected. But in the RTCRtpSender.getStats() case, since we can't distinguish senders, I think it's better to include all sender stats for the track than to fail. Either failing or leaving as-is is OK by me but can you add a comment about this if you proceed? Something like:

// In the outbound case this will include sender stats for all senders that send
// this track instead of just the sender that getStats() was invoked on. We do
// this because it is not possible to distinguish which sender the
// track/outbound-rtp stats belong to without a native sender.getStats()
// implementation.

henbos (approved):
LGTM with comments addressed.

src/js/chrome/chrome_shim.js
fippo: still on ES5

src/js/chrome/chrome_shim.js
fippo: done, but down in the sender wrapper

src/js/chrome/chrome_shim.js
fippo: also its shorter :-)

fippo:
and squashed

henbos:
Merge ahoy!

PR #811, opened 2018-04-21
update test expectations for Chrome 66

fippo:
which just became stable. See #805 and #788


fippo:
and firefox nightly seems to have fixed stuff too

henbos (approved):

PR #812, opened 2018-04-21
support maxMessageSize in datachannel.send on receive side too

fippo:
wraps dc.send also on the receiver side in the ondatachannel event
in addition to the createDataChannel method.

### DIFF
--- a/src/js/common_shim.js
+++ b/src/js/common_shim.js
@@ -261,16 +261,9 @@ module.exports = {
     //       message size can be reset for all data channels at a later stage.
     //       See: https://bugzilla.mozilla.org/show_bug.cgi?id=1426831
 
-    var origCreateDataChannel =
-      window.RTCPeerConnection.prototype.createDataChannel;
-    window.RTCPeerConnection.prototype.createDataChannel = function() {
-      var pc = this;
-      var dataChannel = origCreateDataChannel.apply(pc, arguments);
-      var origDataChannelSend = dataChannel.send;
-
-      // Patch 'send' method
-      dataChannel.send = function() {
-        var dc = this;
+    function wrapDcSend(dc, pc) {
+      var origDataChannelSend = dc.send;
+      dc.send = function() {
         var data = arguments[0];
         var length = data.length || data.size || data.byteLength;
         if (length > pc.sctp.maxMessageSize) {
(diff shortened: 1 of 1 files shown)


fippo:
needs rebase to make tests pass

src/js/common_shim.js
```diff
@@ -258,16 +258,9 @@ module.exports = {
...
-
-      // Patch 'send' method
-      dataChannel.send = function() {
-        var dc = this;
+    function wrapDcSend(dc, pc) {
+      var origDataChannelSend = dc.send;
+      dc.send = function() {
         var data = arguments[0];
         var length = data.length || data.size || data.byteLength;
         if (length > pc.sctp.maxMessageSize) {
```
lgrahl: We can correct a few minor things here as well:
* `if (pc.sctp !== null && length > pc.sctp.maxMessageSize) ...`
* Throw `TypeError` instead of `DOMException`

lgrahl (commented):
LGTM. Should we add an e2e test to check if the send shim works when creating a channel on the remote side?

src/js/common_shim.js
fippo: done. Also only doing the check if the readyState is open to pass through the "not open yet" error.

fippo:
will add!

src/js/common_shim.js
```diff
@@ -258,26 +258,30 @@ module.exports = {
...
-      var dataChannel = origCreateDataChannel.apply(pc, arguments);
-      var origDataChannelSend = dataChannel.send;
-
-      // Patch 'send' method
-      dataChannel.send = function() {
-        var dc = this;
+    function wrapDcSend(dc, pc) {
+      var origDataChannelSend = dc.send;
+      dc.send = function() {
         var data = arguments[0];
```
alvestrand, line 267: not part of this CL, but can you tell me why function() and arguments[0] is cleaner than function(data_in); var data = data_in)?

alvestrand (approved):

src/js/common_shim.js
fippo: the intention is to convey the semantics a bit better. If I would write
```
dataChannel.send = function(data) {
   ... do something with data
  origDataChannel.apply(dc, arguments)
})
```
that would be less clear. Downside is that adapter changes RTCDataChannel.prototype.send.length (which is the number of arguments) but apart from idl tests nothing would notice.

PR #813, opened 2018-04-22
firefox: shim RTCDataChannel

fippo:
which used to exist as window.DataChannel. See
  https://bugzilla.mozilla.org/show_bug.cgi?id=1173851

@nils-ohlmeier can you please take a look?

### DIFF
--- a/src/js/firefox/firefox_shim.js
+++ b/src/js/firefox/firefox_shim.js
@@ -275,5 +275,13 @@ module.exports = {
         }
       });
     };
-  }
+  },
+
+  shimRTCDataChannel: function(window) {
+    // rename DataChannel to RTCDataChannel (native fix in FF60):
+    // https://bugzilla.mozilla.org/show_bug.cgi?id=1173851
+    if (window.DataChannel && !window.RTCDataChannel) {
+      window.RTCDataChannel = window.DataChannel;
+    }
+  },
 };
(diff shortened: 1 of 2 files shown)


src/js/firefox/firefox_shim.js
```diff
@@ -213,5 +213,12 @@ module.exports = {
         }
       });
     };
-  }
+  },
+
+  shimRTCDataChannel: function(window) {
+    // rename DataChannel to RTCDataChannel: https://bugzilla.mozilla.org/show_bug.cgi?id=1173851
```
nils-ohlmeier: Might help to mention here that this landed in Fx 60.

src/js/firefox/firefox_shim.js
```diff
@@ -213,5 +213,12 @@ module.exports = {
...
+    if (window.DataChannel && !window.RTCDataChannel) {
+      window.RTCDataChannel = window.DataChannel;
+    }
```
nils-ohlmeier: Should adapter also provide for some time a shim in the other direction with a deprecation warning, for apps relying on the DataChannel existing in Firefox?

nils-ohlmeier (approved):
LGTM, just two suggestions. Feel to address or drop them.

src/js/firefox/firefox_shim.js
fippo: hrm... going the warning might be tricky. I think the only way would be to define a getter on window which seems like overkill. I don't its important enough, people should fix it relatively quickly.

alvestrand (approved):

PR #814, opened 2018-04-27
chrome: shim RTCPeerConnection.getStats(track)

fippo:
using sender/receiver stats

after this RTCPeerConnection-track-stats.https.html only fails in
* replaceTrack(): original track attachment stats present after replacing
* RTCPeerConnection.getStats(track) throws InvalidAccessError when there are multiple senders for the track

(at some point i want WPT to run here...)

### DIFF
--- a/src/js/chrome/chrome_shim.js
+++ b/src/js/chrome/chrome_shim.js
@@ -232,74 +232,119 @@ module.exports = {
     }
   },
 
-  shimSenderGetStats: function(window) {
+  shimSenderReceiverGetStats: function(window) {
     if (!(typeof window === 'object' && window.RTCPeerConnection &&
-        window.RTCRtpSender)) {
+        window.RTCRtpSender && window.RTCRtpReceiver)) {
       return;
     }
-    if (window.RTCRtpSender && 'getStats' in window.RTCRtpSender.prototype) {
-      return;
-    }
-    var origGetSenders = window.RTCPeerConnection.prototype.getSenders;
-    if (origGetSenders) {
-      window.RTCPeerConnection.prototype.getSenders = function() {
-        var pc = this;
-        var senders = origGetSenders.apply(pc, []);
-        senders.forEach(function(sender) {
-          sender._pc = pc;
+
+    // shim sender stats.
+    if (!('getStats' in window.RTCRtpSender.prototype)) {
+      var origGetSenders = window.RTCPeerConnection.prototype.getSenders;
+      if (origGetSenders) {
+        window.RTCPeerConnection.prototype.getSenders = function() {
+          var pc = this;
+          var senders = origGetSenders.apply(pc, []);
+          senders.forEach(function(sender) {
+            sender._pc = pc;
+          });
+          return senders;
+        };
+      }
+
+      var origAddTrack = window.RTCPeerConnection.prototype.addTrack;
+      if (origAddTrack) {
+        window.RTCPeerConnection.prototype.addTrack = function() {
+          var sender = origAddTrack.apply(this, arguments);
+          sender._pc = this;
+          return sender;
+        };
+      }
+      window.RTCRtpSender.prototype.getStats = function() {
+        var sender = this;
+        return this._pc.getStats().then(function(result) {
+          /* Note: this will include stats of all senders that
+           *   send a track with the same id as sender.track as
+           *   it is not possible to identify the RTCRtpSender.
+           */
+          return filterStats(result, sender.track, true);
         });
-        return senders;
       };
     }
 
-    var origAddTrack = window.RTCPeerConnection.prototype.addTrack;
-    if (origAddTrack) {
-      window.RTCPeerConnection.prototype.addTrack = function() {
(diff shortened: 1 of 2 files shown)


src/js/chrome/chrome_shim.js
```diff
@@ -301,6 +325,29 @@ module.exports = {
...
+
+    var origGetStats = window.RTCPeerConnection.prototype.getStats;
+    window.RTCPeerConnection.prototype.getStats = function() {
+      var pc = this;
+      if (arguments.length > 0 &&
+          arguments[0] instanceof window.MediaStreamTrack) {
+        var track = arguments[0];
+        var sender = pc.getSenders().find(function(s) {
+          return s.track === track;
+        });
```
henbos: Instead of find, do forEach and...
```
if (s.track === track) {
  if (sender)
    throw InvalidAccessError('There are more than one sender or receiver for the track.');
  sender = s;
}
```

src/js/chrome/chrome_shim.js
```diff
@@ -301,6 +325,29 @@ module.exports = {
...
+        var receiver = pc.getReceivers().find(function(r) {
+          return r.track === track;
+        });
```
henbos: Here to, forEach instead of find and...
```
if (r.track === track) {
  if (sender || receiver)
    throw InvalidAccessError('There are more than one sender or receiver for the track.');
  receiver = r;
}
```

src/js/chrome/chrome_shim.js
```diff
@@ -301,6 +325,29 @@ module.exports = {
...
+        if (receiver) {
+          return receiver.getStats();
+        } else if (!sender) {
```
henbos: You forgot to do sender.getStats().

src/js/chrome/chrome_shim.js
```diff
@@ -301,6 +325,29 @@ module.exports = {
...
+          return Promise.reject(new DOMException(
+            'No sender or receiver found for track',
+            'InvalidAccessError'));
```
henbos: nit: If you want to align with the exact error message I chose to be shipped in M67, change this to: "There is no sender or receiver for the track." The error messages I suggested in the code snippets above also align with Chrome's native's error messages.

src/js/chrome/chrome_shim.js
```diff
@@ -270,6 +270,30 @@ module.exports = {
...
+        if (sender) {
+          return sender.getStats();
+        } else if (!receiver) {
+          return Promise.reject(new DOMException(
+            'No sender or receiver found for track',
+            'InvalidAccessError'));
+        }
+      }
+      return origGetStats.apply(pc, arguments);
+    };
```
henbos: Why do you have to shim twice? If you have to, put it in a doShimPeerConnectionGetStats helper with a comment saying why it was needed twice and call it from both places.

henbos (commented):
The spec is explicit about if multiple senders or receivers are sending the same track then throw InvalidAccessError (https://w3c.github.io/webrtc-pc/#dom-rtcpeerconnection-getstats) and it's easy to check when you're looping senders/receivers, plus it would mean passing another WPT test so I think it's worth it - see comments.

src/js/chrome/chrome_shim.js
henbos: * I guess I mean promise.reject, not throw?

src/js/chrome/chrome_shim.js
fippo: I only want to do it when I shim sender/receiver stats... did sender/receiver stats ship in the same release? then I should just merge the two functions...

src/js/chrome/chrome_shim.js
fippo: no, that is the weird part because sender and receiver stats are currently two different shims this will be done in the other shim

henbos:
Is it ready? Did you want me to take another look? Please say when you do :)

fippo:
Its ready but its friday evening... :-)

src/js/chrome/chrome_shim.js
```diff
@@ -232,74 +232,119 @@ module.exports = {
...
-      return e;
-    });
-    window.RTCRtpReceiver.prototype.getStats = function() {
-      var receiver = this;
-      return this._pc.getStats().then(function(result) {
-        return filterStats(result, receiver.track, false);
-      });
+        pc.getReceivers().forEach(function(r) {
+          if (r.track === track) {
+            if (receiver) {
```
henbos, line 326: "sender || receiver" or below "err || (sender && receiver)" or else this will pass if you have a sender.track == receiver.track but just one of each.

henbos (approved):
LGTM % comment

src/js/chrome/chrome_shim.js
fippo: how could the same track object (not just the same id) be on both sender and receiver?
anyway, prevented below and squashed

henbos: Sending a receiving track? That is allowed, right?
pc.addTrack(pc.getReceivers()[0].track);

PR #817, opened 2018-05-09
util: make wrapPeerConnectionEvent configurable

### LINKED ISSUE #751: TypeError: Cannot redefine property: ontrack
Google Chrome
Version 63.0.3239.132

webrtc-adapter@6.1.0

**Description**
  57 |     unwrappedCb]);
  58 | };
  59 | 
  60 | Object.defineProperty(proto, 'on' + eventNameToWrap, {
  61 |   get: function() {
  62 |     return this['_on' + eventNameToWrap];
  63 |   },

I think requried check proto here: https://github.com/webrtc/adapter/blob/master/src/js/utils.js#L60

**Steps to reproduce**

**Expected results**

**Actual results**

fippo:
fixes #751

### DIFF
--- a/src/js/utils.js
+++ b/src/js/utils.js
@@ -71,7 +71,9 @@ function wrapPeerConnectionEvent(window, eventNameToWrap, wrapper) {
         this.addEventListener(eventNameToWrap,
             this['_on' + eventNameToWrap] = cb);
       }
-    }
+    },
+    enumerable: true,
+    configurable: true
   });
 }
 


src/js/utils.js
```diff
@@ -71,7 +71,9 @@ function wrapPeerConnectionEvent(window, eventNameToWrap, wrapper) {
         this.addEventListener(eventNameToWrap,
             this['_on' + eventNameToWrap] = cb);
       }
-    }
+    },
+    enumerable: true,
+    configurable: true
```
alvestrand, line 76: Is it even theoretically possible to write a test for this?
Would it be more logical to make it "enumerable: proto.enumerable"?
I have no idea what it will affect, but just have to assume that you know what you're doing.

alvestrand (approved):
Wish I understood....

src/js/utils.js
fippo: Yes, we could do this as an e2e test:
```
describe('wrapPeerconnectionEvent', () => {
  it('does not throw when executed twice for the same event', () => {
    adapter.utils.wrapPeerConnectionEvent(window, 'something', e => e);
    const wrapAgain = () => {
      adapter.utils.wrapPeerConnectionEvent(window, 'something', e => e);
    };
    expect(wrapAgain).not.to.throw();
  });
});
```
requires exposing utils as adapter.utils though.

Throws on master, passes on this branch

fippo: fails-works
i'd still rather avoid adding this test because its more maintenance than useful

alvestrand: so the case where it makes a difference is if someone else is wrapping the event again?
OK, I'll trust you on this one. Someone can always come back and read the history if they really want to know why the line is there and shouldn't be deleted.

fippo: yeah. in the bug reported adapter was included twice from different sources which attempted to wrap ontrack twice. Silly but... ;-)

PR #798, opened 2018-03-18
edge: move filterIceServers back to adapter

fippo:
moves filterIceServers back to adapter.js from
  https://github.com/otalk/rtcpeerconnection-shim

This is an edge-specific workaround and I'd prefer it in adapter following the "all the ugly workarounds belong here" rationale

PR #823, opened 2018-05-24
fix edge regression

fippo:
when called without any arguments this was throwing an error after #798

### DIFF
--- a/src/js/edge/edge_shim.js
+++ b/src/js/edge/edge_shim.js
@@ -70,7 +70,7 @@ module.exports = {
     var RTCPeerConnectionShim = shimRTCPeerConnection(window,
         browserDetails.version);
     window.RTCPeerConnection = function(config) {
-      if (config.iceServers) {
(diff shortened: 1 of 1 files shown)


src/js/edge/edge_shim.js
```diff
@@ -70,7 +70,7 @@ module.exports = {
     var RTCPeerConnectionShim = shimRTCPeerConnection(window,
         browserDetails.version);
     window.RTCPeerConnection = function(config) {
-      if (config.iceServers) {
+      if (config && config.iceServers) {
```
alvestrand, line 73: Should there be a test for this - that new RTCPeerConnection() works in all cases?

fippo: https://github.com/webrtc/adapter/blob/master/test/e2e/rtcpeerconnection.js#L16-L21

alvestrand (approved):

fippo:
can you make a patch release please?

PR #825, opened 2018-05-29
introduce babel for transpiling

fippo:
uses babel for transpiling to es5. Since the module is es5 this is affecting just the build system

WIP

PR #829, opened 2018-06-12
The shim for webOS 3.0.0+ fixing data-channel

sergeyshalyapin:
Hi,

I want to propose shim for webOS 3.x, which fixes problems with data-channels.

**Description**
Fixes (polyfills) for ES, few object and array methods, that are not supported by webOS 3.0.0. And the data-channel creation and workflow no more fails.

**Purpose**
SmartTV uses Chromium v38 based on blink. So I decided to make updates in the shim of Chrome browser.

Please, have a look and contact me if there any questions on styling or placing.
Thanks, Sergey.

### DIFF
--- a/src/js/chrome/chrome_shim.js
+++ b/src/js/chrome/chrome_shim.js
@@ -873,8 +873,8 @@ module.exports = {
         };
       });
     }
-   
-   // webOS 3.x support. Web engine of SmartTV based on Chromium v38. 
+
+   // webOS 3.x support. Web engine of SmartTV based on Chromium v38.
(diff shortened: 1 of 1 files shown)


sergeyshalyapin:
@googlebot I signed it!

tgabi333:
Personally i think these polyfills should not to be included in this repository. This repo only shims webrtc and not any other missing feature from old browsers. There are a lot of good libs providing such funcionality. If you like to support old chrome, you should include general ES6 polyfills and after webrtc-adapter.

fippo:
maybe #825 solves the  issue by using babel?

PR #837, opened 2018-06-29
chrome: avoid issues with adblock plus

### LINKED ISSUE #831: Error in browser with adblock extension + webrtc-adapter

## Versions affected
**Browser - Chrome 67.0.3396.87**
**adapter.js - v6.1.5**
**adblock extension - now..**

## Description
it appears only if the adblock extension is installed.
attached the screenshots with the exact places where there is a bug
----

---

---

## Steps to reproduce
1) Install adblock extension
2) Use application with `webrtc-adapter`
3) Try to connect to other user (must execute `new RTCPeerConnection(...)`)

## Expected results
works fine, no error

## Actual results
```
Uncaught (in promise) TypeError: Cannot assign to read only property 'iceServers' of object '#<Object>'
window.RTCPeerConnection | @ | chrome_shim.js?fa98:597
WrappedRTCPeerConnection | @ | blob:http://localhos…6d-31ed32b197c2:300
...
```

fippo:
fixes #831

### DIFF
--- a/src/js/chrome/chrome_shim.js
+++ b/src/js/chrome/chrome_shim.js
@@ -738,20 +738,15 @@ module.exports = {
       var OrigPeerConnection = window.RTCPeerConnection;
       window.RTCPeerConnection = function(pcConfig, pcConstraints) {
         if (pcConfig && pcConfig.iceServers) {
-          var newIceServers = [];
           for (var i = 0; i < pcConfig.iceServers.length; i++) {
             var server = pcConfig.iceServers[i];
             if (!server.hasOwnProperty('urls') &&
                 server.hasOwnProperty('url')) {
               utils.deprecated('RTCIceServer.url', 'RTCIceServer.urls');
-              server = JSON.parse(JSON.stringify(server));
               server.urls = server.url;
-              newIceServers.push(server);
-            } else {
-              newIceServers.push(pcConfig.iceServers[i]);
+              pcConfig.iceServers[i] = server;
             }
           }
-          pcConfig.iceServers = newIceServers;
         }
         return new OrigPeerConnection(pcConfig, pcConstraints);
       };


fippo:
a potential problem here is that this modifies the input argument... i.e.
```
var config = {iceServers: [{url: 'stun:stun.example'}]}
var pc = new RTCPeerConnection(config);
```
will modify config. Which seems quite aggressive

fippo:
@dagingaa wdyt?

src/js/chrome/chrome_shim.js
```diff
@@ -738,20 +738,15 @@ module.exports = {
...
             var server = pcConfig.iceServers[i];
             if (!server.hasOwnProperty('urls') &&
                 server.hasOwnProperty('url')) {
               utils.deprecated('RTCIceServer.url', 'RTCIceServer.urls');
-              server = JSON.parse(JSON.stringify(server));
               server.urls = server.url;
-              newIceServers.push(server);
-            } else {
-              newIceServers.push(pcConfig.iceServers[i]);
+              pcConfig.iceServers[i] = server;
```
dagingaa, line 747: This step is unnecessary though (but provides better readability). Seeing as server is an object and the object itself is being modified, saving it back here doesn't do anything new, as the object in the array still references the same object.

I'm wondering if we should do `.map` here instead...

```
pcConfig.iceServers = pcConfig.iceServers.map(server => {
  if (!server.hasOwnProperty('urls') && server.hasOwnProperty('url')) {
    utils.deprecated...
    server.urls = server.url;
  }
  return server;
});
```

but this would reassign pcConfig.iceServers to a new array, which this PR removes, so I might not be understanding the problem correctly wrt ABP.

src/js/chrome/chrome_shim.js
```diff
@@ -738,20 +738,15 @@ module.exports = {
...
       window.RTCPeerConnection = function(pcConfig, pcConstraints) {
         if (pcConfig && pcConfig.iceServers) {
-          var newIceServers = [];
           for (var i = 0; i < pcConfig.iceServers.length; i++) {
             var server = pcConfig.iceServers[i];
             if (!server.hasOwnProperty('urls') &&
                 server.hasOwnProperty('url')) {
               utils.deprecated('RTCIceServer.url', 'RTCIceServer.urls');
-              server = JSON.parse(JSON.stringify(server));
               server.urls = server.url;
```
dagingaa, line 746: Random question, this will make the object contain both `url` and `urls`. Is that desirable? Should we just return `{ urls }` or is there some backward compatible reason for keeping `url`?

fippo: the main reason for doing this these days is to warn about url because Chrome itself doesn't want to. So maybe we should just avoid this problem at all by removing the code.

urls is a clear winner:
https://www.chromestatus.com/metrics/feature/timeline/popularity/1657 -- 0.9%
https://www.chromestatus.com/metrics/feature/timeline/popularity/1656 -- 0.15%

src/js/chrome/chrome_shim.js
dagingaa: Read myself up on the issue now, the map wouldn't work, so disregard that.

Sigh Adblock...

fippo:
closing in favor of #868

Issue #835, opened 2018-06-23
Undefined `constraint` field into `OverconstrainedError`


yak80:
# Please read first!
Please use [discuss-webrtc](https://groups.google.com/forum/#!forum/discuss-webrtc) for general technical discussions and questions.

- [x] I have provided steps to reproduce (e.g. a link to a [jsfiddle](https://jsfiddle.net/))
- [x] I have provided browser name, version and adapter.js version
- [x] This issue only happens when adapter.js is used

**Note: If the checkboxes above are not checked (which you do after the issue is posted), the issue will be closed.**

## Versions affected

**Browser name including version (e.g. Chrome 64.0.3282.119)**
 Chrome 67.0.3396.87 MacOs

**adapter.js (e.g. 6.1.0)**
webrtc-adapter: 6.2.1

## Description
When I connect to not correct media device it return wrong error.

## Steps to reproduce
Exec `navigator.getUserMedia` with wrong device constraints(device  '4803cc..' not connected to PC):
```
constraints= {"audio":{"deviceId":{"exact":"default"}},"video":{"height":{"ideal":480},"width":{"ideal":640},"deviceId":{"exact":"4803cce5e1f6b324950c15cd9f55ae0ba87e08df68b2c8c746fdd210c994f14f"}}}
```
`origGetUserMedia` throw correct error: 
```
{name: "OverconstrainedError", message: null, constraint: "deviceId"}
```

but adapter throw error with undefined `constraint`:
```
{name: "OverconstrainedError", message: null, constraint: undefined}
```

Looks like prob is this [line](https://github.com/webrtc/adapter/blob/04385e855315077cf1d61febc26f5400632f6e3a/src/js/chrome/getusermedia.js#L151), it wait `constraintName` field, but receive `constraint`. Locally hot-fixed with next code:
```
      constraint: (typeof e.constraintName !==  'undefined')
          ? e.constraintName
          : e.constraint,
```

## Expected results
Must thrown error with correct `constraint` field:
```
{name: "OverconstrainedError", message: null, constraint: "deviceId"}
```

## Actual results
```
{name: "OverconstrainedError", message: null, constraint: undefined}
```

fippo:
@jan-ivar can you take a look? You know mediacapture-streams better than I do

jan-ivar:
> `origGetUserMedia` throw correct error:

[...]

PR #838, opened 2018-06-29
chrome: only shim error codes in Chrome < 64

fippo:
(see discussion in #835)
Error names were fixed in M64, see https://groups.google.com/forum/#!msg/discuss-webrtc/fIWg5n67xHo/QIhRnv6vBgAJ
Not using the code in new versions means we have less code in the path and a smaller chance of introducing issues like #835


guidou:
lgtm

henbos (approved):
stampy stampy!

PR #839, opened 2018-07-05
chrome: fix ONN in SRD

fippo:
fixes https://bugs.chromium.org/p/chromium/issues/detail?id=740501#c22
to eat ONN when fired while the signalingState is not stable

### DIFF
--- a/src/js/chrome/chrome_shim.js
+++ b/src/js/chrome/chrome_shim.js
@@ -898,5 +898,15 @@ module.exports = {
       }
       return nativeAddIceCandidate.apply(this, arguments);
     };
+  },
+
+  fixNegotiationNeeded: function(window) {
+    utils.wrapPeerConnectionEvent(window, 'negotiationneeded', function(e) {
+      var pc = e.target;
+      if (pc.signalingState !== 'stable') {
+        return;
+      }
+      return e;
+    });
   }
 };
--- a/src/js/utils.js
+++ b/src/js/utils.js
@@ -25,7 +25,8 @@ function extractVersion(uastring, expr, pos) {
 }
 
 // Wraps the peerconnection event eventNameToWrap in a function
-// which returns the modified event object.
+// which returns the modified event object (or false to prevent
+// the event).
 function wrapPeerConnectionEvent(window, eventNameToWrap, wrapper) {
   if (!window.RTCPeerConnection) {
     return;
@@ -37,7 +38,10 @@ function wrapPeerConnectionEvent(window, eventNameToWrap, wrapper) {
       return nativeAddEventListener.apply(this, arguments);
     }
     var wrappedCallback = function(e) {
-      cb(wrapper(e));
+      var modifiedEvent = wrapper(e);
+      if (modifiedEvent) {
+        cb(modifiedEvent);
+      }
     };
     this._eventMap = this._eventMap || {};
     this._eventMap[cb] = wrappedCallback;
--- a/test/e2e/negotiationneeded.js
+++ b/test/e2e/negotiationneeded.js
@@ -1,0 +1,55 @@
+/*
+ *  Copyright (c) 2018 The WebRTC project authors. All Rights Reserved.
+ *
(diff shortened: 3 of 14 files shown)


src/js/chrome/chrome_shim.js
```diff
@@ -898,5 +898,16 @@ module.exports = {
       }
       return nativeAddIceCandidate.apply(this, arguments);
     };
+  },
+
+  fixNegotiationNeeded: function(window) {
+    utils.wrapPeerConnectionEvent(window, 'negotiationneeded', function(e) {
+      var pc = e.target;
+      console.log(pc.signalingState);
```
jan-ivar: remove debug cruft

test/e2e/negotiationneeded.js
```diff
@@ -0,0 +1,55 @@
...
+      onnfired = true;
+    };
+    pc.setRemoteDescription({type: 'offer', sdp})
+    .then(() => navigator.mediaDevices.getUserMedia({audio: true}))
+    .then((stream) => pc.addTrack(stream.getTracks()[0], stream))
+    .then(() => {
+      setTimeout(() => {
+        expect(onnfired).to.equal(false);
+        done();
+      }, 0);
```
jan-ivar, line 52: Missing `catch`. Why not write this with `async`/`await` instead?

jan-ivar (approved):
Looks like an improvement, but I fear insufficient to solve the problem. When I test https://jsfiddle.net/jib1/7otejw5x/ —which uses a similar workaround—in Chrome, it looks like Chrome fires ONN *again* from stable state! This leads to a number of back-and-forth negotiations, as you can tell from the console, until it it looks like it triggers some race and errors out with
```
Uncaught (in promise) DOMException: Failed to set remote answer sdp: Called in wrong state: kStable
```

src/js/utils.js
```diff
@@ -37,7 +38,10 @@ function wrapPeerConnectionEvent(window, eventNameToWrap, wrapper) {
       return nativeAddEventListener.apply(this, arguments);
     }
     var wrappedCallback = function(e) {
-      cb(wrapper(e));
+      var modifiedEvent = wrapper(e);
+      if (modifiedEvent !== false) {
```
jan-ivar: Nit: I prefer
```js
if (modifiedEvent) {
```
or even better
```js
if (!modifiedEvent) {
  return;
}
```
Duck-typing FTW! (the tripple-equals crowd is wrong).

test/e2e/negotiationneeded.js
fippo: if this throws, it will fail the test so we don't need a catch

src/js/utils.js
fippo: I wanted to avoid "eating" the event if nothing/undefined is returned

fippo:
yeah, this does not fix the
pc.addTrack(...)
pc.addTrack(...)
fires ONN twice issue. Still pondering if setting a _negotiationneeded flag similar to the slot but unsure how much effort that is.

src/js/utils.js
jan-ivar: Last I checked, `undefined` is not an event.

jan-ivar:
> fires ONN twice issue

That one can be fixed with a re-entrancy guard:
```js
pc.onnegotiationneeded = async e => {
  if (pc._negotiating == true) return;
  pc._negotiating = true;
  try {
    return cb(e);
  } finally {
    pc._negotiating = false;
  }
}
```
but I believe that's separate from the never-ending renegotiation problem, which AFAICT seems to come from ONN re-firing after SLD(answer) (or maybe SRD(answer)?) brings us back to stable state.

src/js/utils.js
fippo: oh well, if the function that is modified returns something else than an event and its not trying to prevent the event things are going to go wrong anyway. Updated

fippo:
merging this and trying to tackle re-entrancy protection in another PR!

PR #840, opened 2018-07-06
test: bump karma-stability-reporter

fippo:
which now does not need passing tests anymore. This removes
a lot of chore when adding a new test


fippo:
@henbos wdyt? IIRC chrome uses the same strategy for the expectation files, only listing failures

package.json
```diff
@@ -47,7 +47,7 @@
     "karma-mocha": "^1.3.0",
     "karma-mocha-reporter": "^2.2.3",
     "karma-safari-launcher": "^1.0.0",
-    "karma-stability-reporter": "git+https://github.com/fippo/karma-stability-reporter.git#v2.1.0",
+    "karma-stability-reporter": "^3.0.1",
```
henbos, line 50: Does this remove the dependency on a forked karma-stability-reporter in favor of the official karma-stability-reporter?

In any case, bumping the version sounds like a good idea to do every now and then.

test/e2e/expectations/Electron
```diff
@@ -1,100 +1 @@
...
-PASS removeTrack after addStream for an audio/video track after removing all tracks no senders with tracks remain
-PASS removeTrack after addStream for an audio/video track after removing all tracks no local streams remain
-PASS removeTrack after addTrack for an audio/video track after removing a single track only a single sender with a track remains
-PASS removeTrack after addTrack for an audio/video track after removing a single track the local stream remains untouched
-PASS removeTrack after addTrack for an audio/video track after removing all tracks no senders with tracks remain
-PASS removeTrack after addTrack for an audio/video track after removing all tracks no local streams remain
-PASS RTCIceCandidate window.RTCIceCandidate exists
-PASS RTCIceCandidate is augmented in the onicecandidate callback
-PASS RTCIceCandidate is augmented in the icecandidate event
 ERR RTCIceCandidate with empty candidate.candidate does not throw AssertionError
```
henbos, line 1: Are the PASS now optional or are they no longer generated when you reset the expectations?

I don't have a strong opinion on this. On one hand it's spammy, on the other hand if you do inspect an expecations file to find a failing test, it's kind of nice to see which surrounding tests are passing without having to hunt down that test file, since surrounding tests usually test similar behaviors.

henbos (commented):
I'm happy to LGTM this but I have some questions about the new behavior.

package.json
fippo: no, i accidentially published to npm while previous versions were not

test/e2e/expectations/Electron
fippo: PASS is not only optional but undeseriable. Basically to add a test I had to add a "PASS" in most browsers except the browser where the test failed. This created pretty spammy PRs and made adding tests a pain

henbos: Good point

henbos (approved):
Let's do it!

PR #841, opened 2018-07-09
shim getDisplayMedia

fippo:
adds opt-in shims for getDisplayMedia for Firefox and Chrome.

In Firefox the adapter-user has to choose whether to select screen or window like this:
```
if (adapter.browserDetails.browser === 'firefox') {
  adapter.browserShim.shimGetDisplayMedia(window, 'screen');
}
```
whereas in Chrome more integration glue between Chrome and the extension has to provided as a Promise resolving with a sourceId:
```
adapter.browserShim.shimGetDisplayMedia(window, function() {
  })
})
```


src/js/chrome/chrome_shim.js
```diff
@@ -908,5 +908,29 @@ module.exports = {
...
+  },
+
+  shimGetDisplayMedia: function(window, getSourceId) {
+    if ('getDisplayMedia' in window.navigator) {
+      return;
+    }
+    // getSourceId is a function that returns a promise resolving with
+    // the sourceId of the screen to be shared.
+    if (!getSourceId) {
+      return;
```
jan-ivar: Maybe throw an error instead of failing silently here? Also, maybe
```js
if (typeof getSourceId != "function") {
  return;
```
to proof it a bit against pilot error, since this arg is sometimes a function and sometimes a string?

src/js/chrome/chrome_shim.js
```diff
@@ -908,5 +908,29 @@ module.exports = {
...
+    }
+    navigator.getDisplayMedia = function(constraints) {
+      return getSourceId(constraints)
+        .then(function(sourceId) {
+          constraints.video = {mandatory: {
+            chromeMediaSource: 'desktop',
```
jan-ivar: Nit: prefer discrete indents
```js
constraints.video = {
  mandatory: {
```

src/js/chrome/chrome_shim.js
```diff
@@ -908,5 +908,29 @@ module.exports = {
...
+            chromeMediaSourceId: sourceId,
+            maxWidth: window.screen.width,
+            maxHeight: window.screen.height,
```
jan-ivar: In April [we decided to bring back constraints](https://docs.google.com/presentation/d/1sboQqMc5O1-oZ7w0ZY1VUyhY8JG0VxOzU8kAQitoHJU/edit#slide=id.g2c521efd37_0_38) on `getDisplayMedia` modulo `min`, `max` and `exact`, a PR should be forthcoming.

Does Chrome support downscaling? If so, this would override any constraints inputs. Do we want:
```js
maxWidth: constraints.video.width || window.screen.width,
maxHeight: constraints.video.height || window.screen.height,
```
?

src/js/firefox/firefox_shim.js
```diff
@@ -284,4 +284,26 @@ module.exports = {
...
+
+  shimGetDisplayMedia: function(window, preferredMediaSource) {
+    if ('getDisplayMedia' in window.navigator) {
+      return;
+    }
+    navigator.getDisplayMedia = function(constraints) {
+      if (!(constraints && constraints.video)) {
+        var err = new Error('getDisplayMedia without video constraints ' +
+            'is undefined');
+        err.name = 'NotSupportedError';
```
jan-ivar: `NotFoundError` I think, from [getUserMedia](https://w3c.github.io/mediacapture-main/getusermedia.html#dom-mediadevices-getusermedia).

src/js/firefox/firefox_shim.js
```diff
@@ -284,4 +284,26 @@ module.exports = {
       window.RTCDataChannel = window.DataChannel;
     }
   },
+
+  shimGetDisplayMedia: function(window, preferredMediaSource) {
```
jan-ivar, line 288: Do we want to default to "window" maybe?

The reason I mention it is that fullscreen is more "scary" from a privacy perspective.

jan-ivar (approved):
🎉 Thanks for adding this! I wonder if we could enable more functionality by default, but this is a fantastic start!

src/js/chrome/chrome_shim.js
fippo: good question. @guidou: how much of this "mandatory" stuff do we still need? I copied it from ancient calls which still work but maybe there is a better way now? Does one still need width/height scraped from window in order to prevent the returned stream from being 640x480?

src/js/chrome/chrome_shim.js
fippo: Went for console.error

src/js/firefox/firefox_shim.js
```diff
@@ -284,4 +284,26 @@ module.exports = {
...
+    if ('getDisplayMedia' in window.navigator) {
+      return;
+    }
+    navigator.getDisplayMedia = function(constraints) {
+      if (!(constraints && constraints.video)) {
+        var err = new Error('getDisplayMedia without video constraints ' +
```
jan-ivar: We should throw a `DOMException` here. E.g.
```js
var err = new DOMException(
  'getDisplayMedia without video constraints is undefined',
  'NotFoundError');

File: src/js/chrome/chrome_shim.js (hunk 3 above)
Comment (fippo) [reply src/js/chrome/chrome_shim.js]: looks like we can omit that. I just did and got a stream with the right dimensions!

Comment (fippo) [reply src/js/chrome/chrome_shim.js]: which reduces the question to @guidou to "do we need 'mandatory' or is there a closer-to-spec way to pass these non-spec constraints' :-)

File: src/js/firefox/firefox_shim.js (hunk 5 above)
Comment (fippo) [reply src/js/firefox/firefox_shim.js:288]: is choosing a default and calling shimGetDisplayMedia(window) instead of shimGetDisplayMedia(window, preferredMediaSource) really worth it?

**Comment by fippo** [comment]
so there is an additional problem in Chrome here. If it requires an extension and does not use native getScreenMedia, what do we do in the case that there is no extension?

I am currently leaning towards "dear user, do your own error handling in the function you supply" along these lines:
```
adapter.browserShim.shimGetDisplayMedia(window, function() {
     if (!sessionStorage.getScreenMediaJSExtensionId) {
       var err = new Error('you need an extension');
       err.code = 'needextension';
       return Promise.reject(err);
    }
    chrome.runtime.sendMessage(sessionStorage.getScreenMediaJSExtensionId, {type:'getScreen', id: 1}, null,
     function (data) {
       if(!data || data.sourceId === '') { // user canceled
         var error = new Error('NavigatorUserMediaError');
         error.name = 'NotAllowedError';
         reject(error);
       } else {
         resolve(data.sourceId);
       }
     })
  })
})
```
which puts the burden of checking when to install on the developer.

Subsequent code could then just use getDisplayMedia but needs to handle potential errors:
```
navigator.getDisplayMedia({video: true})
.catch(e => {
   if (e.code === 'needextension') {
     // handle installation
  }
})
```
which is slightly less confusing than having to store whether getDisplayMedia was shimmed and then (and only then) doing an upfront check whether getDisplayMedia is going to succeed

fippo:
@alvestrand if you're ok with this too I'd like to do a new release and then a blog post :-)

alvestrand (approved):

PR #842, opened 2018-07-09
chrome: fix ONN (take 2)

fippo:
fixes another instance of ONN not working as it should. See
  https://bugs.chromium.org/p/chromium/issues/detail?id=740501
and bugs linked from there

(not adding test expectations but the test fails without shim and passes with shim, will rebase ontop of #840)

### DIFF
--- a/test/e2e/negotiationneeded.js
+++ b/test/e2e/negotiationneeded.js
@@ -38,16 +38,31 @@ describe('negotiationneeded event', () => {
         'a=rtpmap:111 opus/48000/2\r\n' +
         'a=msid:stream1 track1\r\n' +
         'a=ssrc:1001 cname:some\r\n';
-    var onnfired = false;
+    let onnfired = false;
     pc.onnegotiationneeded = () => {
       onnfired = true;
     };
     pc.setRemoteDescription({type: 'offer', sdp})
     .then(() => navigator.mediaDevices.getUserMedia({audio: true}))
-    .then((stream) => pc.addTrack(stream.getTracks()[0], stream))
-    .then(() => {
+    .then((stream) => {
+      pc.addTrack(stream.getTracks()[0], stream);
       setTimeout(() => {
         expect(onnfired).to.equal(false);
+        done();
+      }, 0);
+    });
+  });
+
+  it('fires only once after adding two tracks', (done) => {
+    let onnfired = 0;
+    pc.onnegotiationneeded = () => {
+      onnfired++;
+    };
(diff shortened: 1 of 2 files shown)


fippo:
the big question here is "does it break existing use-cases"... I don't have an answer to that.

fippo:
@henbos PTAL -- alternatively you can fix the actual bug :-)

test/e2e/negotiationneeded.js
```diff
@@ -52,4 +52,21 @@ describe('negotiationneeded event', () => {
       }, 0);
     });
   });
+
+  it('fires only once after adding two tracks', (done) => {
```
henbos, line 56: nit: "done =>", also I keep forgetting, can we "async done =>" and await instead of then()?

test/e2e/negotiationneeded.js
```diff
@@ -52,4 +52,21 @@ describe('negotiationneeded event', () => {
...
+    pc.onnegotiationneeded = () => {
+      onnfired++;
+    };
+    navigator.mediaDevices.getUserMedia({audio: true, video: true})
+    .then(stream => {
+      stream.getTracks().forEach(t => pc.addTrack(t, stream));
+    })
+    .then(() => {
+      setTimeout(() => {
+        expect(onnfired).to.equal(1);
```
henbos: onnegotiationneeded is fired by "queue a task that fires". Is this why you had to introduce an additional delay with ".then()"? Mixing promises, events and setTimeout() always has my head scratching about which type of "tasks" are executed in what order.

Can you do 1) or 2)?
1. It is not obvious why you don't just do the setTimeout() above, can you add a comment about the additional ".then()"?
2. Alternatively, to minimize assumptions about exactly when events fire and just make it about the number of events firing (there could be an unrelated bug about it being fired later ;) ), can you add the setTimeout() with expectations inside of the first onnegotiationneeded iff onnfired == 0?

test/e2e/negotiationneeded.js
fippo: i'd rather not mix styles. Besides, the better style for what you have in mind is to return a promise in which case one doesn't need the done cb

test/e2e/negotiationneeded.js
fippo: this is indeed more complicated than necessary. Updated, still passes.

dagingaa (approved):

fippo:
i'll punt on this, too much risk and @guidou is fixing it in Chrome.

PR #843, opened 2018-07-09
safari: make addstream polyfill work with addEventListener

### LINKED ISSUE #834: safari shim addstream event doesn't work with addEventListener
https://github.com/webrtc/adapter/blob/master/src/js/safari/safari_shim.js#L99
does not work when using `pc.addEventListener('addstream', ...)`, only with `pc.onaddstream =`

I assume this needs something similar to Chrome's 'track' shim in SRD.

fippo:
fixes #834

### DIFF
--- a/src/js/safari/safari_shim.js
+++ b/src/js/safari/safari_shim.js
@@ -102,12 +102,17 @@ module.exports = {
           return this._onaddstream;
         },
         set: function(f) {
-          var pc = this;
           if (this._onaddstream) {
             this.removeEventListener('addstream', this._onaddstream);
-            this.removeEventListener('track', this._onaddstreampoly);
           }
           this.addEventListener('addstream', this._onaddstream = f);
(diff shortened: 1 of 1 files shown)


fippo:
this is based on this piece of code which works for chrome in a similar situation: https://github.com/webrtc/adapter/blob/0141d08ecaf87bd1daf834849a7fee92420880f8/src/js/chrome/chrome_shim.js#L61-L80

alvestrand:
This means that _onaddstreampoly is set the first time you call setRemoteDescription, rather than before you call anything. Why is that a Good Thing?

fippo:
the current way only works if you set pc.onaddstream = and does not work with addEventListener('addstream').
This does not make a difference in behaviour as SRD is the only thing that can trigger the addstream event

alvestrand (approved):

Issue #832, opened 2018-06-19
Uncaught TypeError: Cannot redefine property: ontrack if adapter.js included twice on the page


lvass74:
## Versions affected

Chrome 63.0.,239.111 on Android 6.0.1
Chrome 63.0.,239.111 on Android 8.0.0

latest adapter.js (https://webrtc.github.io/adapter/adapter-latest.js) 

## Description
If adapter.js is included twice (more than once) in the page and the browser does not provide window.RTCPeerConnection.prototype.ontrack by default then the following error is thrown: "Uncaught TypeError: Cannot redefine property: ontrack"

Before first insertion window.RTCPeerConnection.prototype.ontrack is undefined, after first insertion it is defined but with configurable: false thus redefinition at second insertion fails.

N.B.:
Double inclusion is a real use case when once uses the adapter.js plus some other package that has webrtc-adapter npm package as dependency.

## Steps to reproduce
Reproduction case can be seen on any of the following URLs:
https://s3-eu-west-1.amazonaws.com/tempfilesbucket/COL-1281/ontrack.html
http://jsfiddle.net/az5npxt0/5/

## Expected results
No exception should be thrown

## Actual results
Exception thrown: "Uncaught TypeError: Cannot redefine property: ontrack"

## Suggested solution
"enumerable: true, configurable: true" should be added to https://github.com/webrtc/adapter/blob/master/src/js/chrome/chrome_shim.js#L64 as it was added to https://github.com/webrtc/adapter/blob/master/src/js/utils.js#L75 in https://github.com/fippo/adapter/commit/4f89e1eef675c9391b5d80b953811200726adf70

fippo:
*sigh*
the code around line 64 is guarded by a !('ontrack' in RTCPeerConnection.prototype) but the second include attempts to modify via the other path

When using npm and require() double inclusion is not a problem so I don't really see the use-case. But its easy to fix so #833 should take care of that

PR #854, opened 2018-07-30
firefox_shim: make ontrack shim configurable

kondurake:
If you include adapter-latest.js (v6.3.0) in a web page, old versions of Firefox (in my case 43.0.1) will throw the following:
> TypeError: can't redefine non-configurable property "ontrack"

This issue is similar to #832 (fixed by #833). Except that this one occurs even if you include the adapter only once.

### DIFF
--- a/src/js/firefox/firefox_shim.js
+++ b/src/js/firefox/firefox_shim.js
@@ -35,7 +35,9 @@ module.exports = {
               this.dispatchEvent(event);
             }.bind(this));
           }.bind(this));
-        }
+        },
+        enumerable: true,
+        configurable: true
       });
     }
     if (typeof window === 'object' && window.RTCTrackEvent &&


kondurake:
I signed it!

fippo:
that is a terribly old version which is far beyond anything we support (FF ESR which is 52) but this seems consistent with more recent changes in the chrome shim.

Will merge on CI passes, thank you!

fippo (approved):

kondurake:
Unfortunately, I did not have the opportunity to test it using newer versions. However, my goal was to tell you about the issue and I thought that it did not matter which version I use for testing.

Thank you for your time! I'm very glad that you approved my patch.

fippo:
This will go into version 6.3.1 -- as soon as I've merged some other pending PRs.

Issue #857, opened 2018-07-31
ontrack transceiver shim broken in M69+


fippo:
- [x] I have provided steps to reproduce (e.g. a link to a [jsfiddle](https://jsfiddle.net/))
- [x] I have provided browser name, version and adapter.js version
- [x] This issue only happens when adapter.js is used

## Versions affected
**Browser name including version (e.g. Chrome 69.0.3497.12)**
**adapter.js (e.g. 6.3.0)**

## Description

tests fail currently with this error:
```
  track event
    the event has
      ✖ a transceiver that has a receiver
        HeadlessChrome 69.0.3497 (Linux 0.0.0)
      Error: Uncaught TypeError: Cannot read property 'receiver' of null 
    (node_modules/chai/chai.js:1179)
```
This is because we do not attempt to shim trackEvent.transceiver if `RTCRtpTransceiver in window`.
But while that is true, the transceiver is only set with sdpSemantics set to unified-plan
I'd prefer to avoid an unconditional shim or one that has a specific version...

alvestrand:
You can detect this case by checking if the transceiver has a receiver, I think.
In Unified Plan it will always have one; in plan B .... it seems not.

fippo:
yes, the shim doesn't attempt to overwrite for that reason. But it won't get activated when transceivers are implemented. #861 fixes that but it means there is no way for the shim to automatically kick in when it is no longer required which is very sad.

PR #860, opened 2018-08-02
safari: fix createOffer callbacks API

### LINKED ISSUE #859: CreateOffer only works with promises in Safari 11.1.2
- [ X ] I have provided steps to reproduce
[Jsfiddle](https://jsfiddle.net/cg1b93fm/18/)
- [ X ] I have provided browser name, version and adapter.js version
Safari 11.1.2 on MacOs Sierra 10.12.6 / adapterjs 6.1.3
- [ X ] This issue only happens when adapter.js is used
Not really, with adapter, promise based createOffer works, without both of them fails.

## Description
When creating an offer with constraints { offerToReceiveAudio: true, offerToReceiveVideo: true } with a callback style createoffer, the sdps generated are only:

```
{sdp: "v=0
↵o=- 6889517193703606396 2 IN IP4 127.0.0.1
↵s=-
↵t=0 0
↵a=msid-semantic: WMS
↵", type: "offer"}
```

when using promise based createOffer, everything works fine.
Shouldn't adapter shim the legacy callback style in Safari and produce the same result?

dagingaa:
This change fixes a bug where the legacy callbacks API for Safari
createOffer didn't return as expected when compared to the currently
working promises API.

This was due to the order of shims in adapter_factory, where the
shimCreateOfferLegacy which does trasceiver stuff and overwrites the
createOffer prototype was called after shimCallbacksAPI already fixed
the createOffer function to support callbacks.

Moreover, this happened only because the order of arguments to
createOffer int he first place, where the callbacks come first (wtf)
before the options, meaning that the options argument changes order
between the callback and promise API. If only we could go back in
time...

Fixes #859

**Description**

**Purpose**

### DIFF
--- a/src/js/adapter_factory.js
+++ b/src/js/adapter_factory.js
@@ -134,12 +134,12 @@ module.exports = function(dependencies, opts) {
       commonShim.shimCreateObjectURL(window);
 
       safariShim.shimRTCIceServerUrls(window);
+      safariShim.shimCreateOfferLegacy(window);
       safariShim.shimCallbacksAPI(window);
       safariShim.shimLocalStreamsAPI(window);
       safariShim.shimRemoteStreamsAPI(window);
       safariShim.shimTrackEventTransceiver(window);
       safariShim.shimGetUserMedia(window);
-      safariShim.shimCreateOfferLegacy(window);
 
       commonShim.shimRTCIceCandidate(window);
       commonShim.shimMaxMessageSize(window);


fippo:
test failure is due to #857, i'll fix that tomorrow

fippo:
the rationale for the old order was that you always want a callback (because you need the sdp) but only sometimes need to specify options.

fippo (approved):

fippo:
will do a new release tomorrow after fixing #857.

Also for tradition sake: :cake:

PR #868, opened 2018-08-29
chrome: remove iceServers warning

### LINKED ISSUE #831: Error in browser with adblock extension + webrtc-adapter

fippo:
this battle seems mostly over:
https://www.chromestatus.com/metrics/feature/timeline/popularity/1657 -- urls 0.9%
https://www.chromestatus.com/metrics/feature/timeline/popularity/1656 -- url  0.15%

Fixes #831 in a better way than #837

### DIFF
--- a/src/js/chrome/chrome_shim.js
+++ b/src/js/chrome/chrome_shim.js
@@ -739,35 +739,6 @@ module.exports = {
           }
         });
       }
-    } else {
-      // migrate from non-spec RTCIceServer.url to RTCIceServer.urls
-      var OrigPeerConnection = window.RTCPeerConnection;
-      window.RTCPeerConnection = function(pcConfig, pcConstraints) {
-        if (pcConfig && pcConfig.iceServers) {
-          var newIceServers = [];
-          for (var i = 0; i < pcConfig.iceServers.length; i++) {
-            var server = pcConfig.iceServers[i];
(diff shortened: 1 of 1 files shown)


fippo:
I *think* we don't need a major version break as the only change in behaviour is not logging this anymore.
Third alternative would be not to modify but just warn but that seems quite pointless.

dagingaa (approved):
This seems like the better move here for sure compared to #837

dagingaa:
Agree with minor version bump here. Unless there are some browsers relying on the urls property, and do not support url?

fippo:
this is only for chrome and it still supports both.

PR #885, opened 2018-09-25
travis: try npm test twice

fippo:
band-aid for flaky tests. I am looking at you DTMF...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,7 +40,7 @@ before_script:
 
 script:
   - node_modules/.bin/grunt
-  - npm test
+  - npm test || npm test
 
 after_failure:
   - for file in *.log; do echo $file; echo "======================"; cat $file; done || true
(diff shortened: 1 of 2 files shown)


dagingaa:
Ironic that the test fails... :P If I were Brad now I would tell you to not put a bandaid on it and see what the root cause is and if it can be fixed. If the DTMF tests are flaky should we disable them instead for the time being?

fippo:
the dtmf test needs an additional check. Pondering whether to actually remove it... it has been flaky for ages.

fippo:
Looking at the actual test showed that the anti-flake method was only used in one of the two. Updated... also stop sub-commenting about @bradleythughes :-p

dagingaa:
I couldn’t autocomplete his name for some reason :p

.travis.yml
```diff
@@ -40,7 +40,7 @@ before_script:
 
 script:
   - node_modules/.bin/grunt
-  - npm test
+  - npm test || npm test
```
dagingaa, line 43: Do we still need this then?

dagingaa (approved):

.travis.yml
fippo: i can make sure all the emails go to you... lets keep it :-p

fippo: (there are also some firefox timeout crash things....)

PR #890, opened 2018-10-13
chrome: support width/height in getDisplayMedia

fippo:
using a very simple shim

needs to be backported to 6.1.4 branch


dagingaa (approved):

fippo:
this didn't really work. Passing a maxWidth of undefined throws an error.

fippo:
published 6.4.2 but that was wrong. Now publishing 6.4.3...

PR #891, opened 2018-10-13
package: make adapter an ES6 module

fippo:
which can be imported as
```
import adapter from 'webrtc-adapter';
```

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -18,7 +18,7 @@ module.exports = function(grunt) {
     },
     browserify: {
       adapterGlobalObject: {
-        src: ['./dist/adapter_core.js'],
+        src: ['./dist/adapter_core5.js'],
         dest: './out/adapter.js',
         options: {
           browserifyOptions: {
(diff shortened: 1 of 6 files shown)


fippo:
@dagingaa can you give this a try? Works for me that way. I also wonder if we should get rid of the dist directory, I am not very fond of committing artifacts -- which are on the gh-pages branch anyway.

fippo:
joy... so the ES6 import works but now browserify doesn't work anymore... :-)

Gruntfile.js
```diff
@@ -31,12 +31,12 @@ module.exports = function(grunt) {
       // Use this if you do not want adapter to expose anything to the global
       // scope.
       adapterAndNoGlobalObject: {
```
fippo, line 33: pondering dropping this...

fippo:
@legastero said it works for him a week back or so

dagingaa:
Works for me, I tried it and found out I had to change my webpack config to resolve from "module".

Old config:
`mainFields: ["browser", "jsnext:main", "main"],`

New config:
`mainFields: ["browser", "module", "main"],`

fippo:
fab, will release soon.

fippo:
note: publish like this: https://docs.npmjs.com/adding-dist-tags-to-packages

PR #897, opened 2018-11-01
shim navigator.mediaDevices.getDisplayMedia

### LINKED ISSUE #896: Move getDisplayMedia to navigator.mediaDevices
A heads-up that `getDisplayMedia` just moved to `navigator.mediaDevices.getDisplayMedia` in https://github.com/w3c/mediacapture-screen-share/pull/86.

We should update adapter.js to match ASAP.

fippo:
fixes #896 -- quite notably this is not backward compatible but this is on the ES6 master branch.

The backport to 6.x will need to consider navigator.getDisplayMedia.

### DIFF
--- a/src/js/chrome/chrome_shim.js
+++ b/src/js/chrome/chrome_shim.js
@@ -52,6 +52,7 @@ function filterStats(result, track, outbound) {
 }
 
 export {shimGetUserMedia} from './getusermedia';
+export {shimGetDisplayMedia} from './getdisplaymedia';
 
 export function shimMediaStream(window) {
   window.MediaStream = window.MediaStream || window.webkitMediaStream;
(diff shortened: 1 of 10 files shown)


src/js/chrome/chrome_shim.js
```diff
@@ -734,7 +735,7 @@ export function shimGetDisplayMedia(window, getSourceId) {
         'a function');
     return;
   }
-  navigator.getDisplayMedia = function(constraints) {
+  window.navigator.mediaDevices.getDisplayMedia = function(constraints) {
```
jan-ivar: This will throw if `navigator.mediaDevices` is `undefined`.

We'd probably want to check we're not in HTTP before adding that object, since we just made it [SecureContext]. Something like:
```
if (location.protocol != 'https:') {
  return;
}
if (!("mediaDevices" in window.navigator)) {
  window.navigator.mediaDevices = {};
}
```

src/js/edge/edge_shim.js
```diff
@@ -86,3 +86,15 @@ export function shimReplaceTrack(window) {
...
+
+export function shimGetDisplayMedia(window) {
+  if (!('getDisplayMedia' in window.navigator)) {
+    return;
+  }
+  if (window.navigator.mediaDevices &&
+    'getDisplayMedia' in window.navigator.mediaDevices) {
+    return;
+  }
+  window.navigator.mediaDevices.getDisplayMedia =
```
jan-ivar: Same here.

src/js/firefox/firefox_shim.js
```diff
@@ -158,10 +158,11 @@ export function shimReceiverGetStats(window) {
 }
 
 export function shimGetDisplayMedia(window, preferredMediaSource) {
-  if ('getDisplayMedia' in window.navigator) {
+  if (window.navigator.mediaDevices &&
+    'getDisplayMedia' in window.navigator.mediaDevices) {
     return;
   }
-  navigator.getDisplayMedia = function(constraints) {
+  window.navigator.mediaDevices.getDisplayMedia = function(constraints) {
```
jan-ivar: and here.

jan-ivar (approved):
🎉

src/js/chrome/chrome_shim.js
fippo: I think we have broken compat with versions that do not have navigator.mediaDevices on this branch already.

Does [SecureContext] mean this only exists when using https?

jan-ivar: Yes.

fippo: that probably has implications in other places too :-|

fippo:
updated -- also moved the implementation to separate files just because :-)

jan-ivar (approved):
Yeah that works. Shimming `getDisplayMedia` on pre-`navigator.mediaDevices` browser is probably best left as an exercise for the reader.

PR #899, opened 2018-11-13
Fail silently if unable to shim GetUserMedia & AddTrackRemoveTrack & PeerConnection on chrome

### LINKED ISSUE #895: Script should silently fail if `RTCPeerConnection` is undefined

## Versions affected

Browser: **Google Crawler**
**webrtc-adapter 6.4.3**

## Description

Google crawler uses some special version of Chrome to crawl pages and it has no `RTCPeerConnection` support. That leads to google crawler seeing blank pages for SPA because of a JS error.

## Steps to reproduce

1. Import `webrtc-adapter` somewhere in any SPA (e.g create-react-app) so it's all bundled in a single file
2. Use https://search.google.com/test/mobile-friendly to see that it fails with `Uncaught TypeError: Cannot read property 'prototype' of undefined` that happens on `RTCPeerConnection.prototype`

## Expected results

Google crawler can render the page

## Actual results

Google crawler can not render the page

mariuszbeltowski:
**Description**
Fail silently if unable to shim GetUserMedia & AddTrackRemoveTrack & PeerConnection on chrome

**Purpose**
Fixed #895

### DIFF
--- a/src/js/chrome/chrome_shim.js
+++ b/src/js/chrome/chrome_shim.js
@@ -403,6 +403,9 @@ export function shimAddTrackRemoveTrackWithNative(window) {
 }
 
 export function shimAddTrackRemoveTrack(window) {
+  if (!window.RTCPeerConnection) {
+    return;
+  }
   const browserDetails = utils.detectBrowser(window);
   // shim addTrack and removeTrack.
   if (window.RTCPeerConnection.prototype.addTrack &&
@@ -624,6 +627,9 @@ export function shimPeerConnection(window) {
   if (!window.RTCPeerConnection && window.webkitRTCPeerConnection) {
     // very basic support for old versions.
     window.RTCPeerConnection = window.webkitRTCPeerConnection;
+  }
+  if (!window.RTCPeerConnection) {
+    return;
   }
 
   const origGetStats = window.RTCPeerConnection.prototype.getStats;
--- a/src/js/chrome/getusermedia.js
+++ b/src/js/chrome/getusermedia.js
@@ -11,8 +11,13 @@ import * as utils from '../utils.js';
 const logging = utils.log;
 
 export function shimGetUserMedia(window) {
-  const browserDetails = utils.detectBrowser(window);
   const navigator = window && window.navigator;
+
+  if (!navigator.mediaDevices) {
+    return;
+  }
+
+  const browserDetails = utils.detectBrowser(window);
 
   const constraintsToChrome_ = function(c) {
     if (typeof c !== 'object' || c.mandatory || c.optional) {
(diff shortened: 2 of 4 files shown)


src/js/chrome/getusermedia.js
```diff
@@ -11,9 +11,14 @@ import * as utils from '../utils.js';
 const logging = utils.log;
 
 export function shimGetUserMedia(window) {
-  const browserDetails = utils.detectBrowser(window);
   const navigator = window && window.navigator;
 
+  if (!navigator.mediaDevices) {
```
fippo, line 16: I think this needs to be if (!(navigator && navigator.mediaDevices)) to be perfectly safe.

src/js/chrome/chrome_shim.js
```diff
@@ -403,6 +403,9 @@ export function shimAddTrackRemoveTrackWithNative(window) {
 }
 
 export function shimAddTrackRemoveTrack(window) {
+  if (!window.RTCPeerConnection) {
```
fippo, line 406: I wondered if this is needed in shimAddTrackRemoveTrackWithNative too but it isn't because that path is going through this function.

fippo:
LGTM overall, thank you!
I considered doing this in adapter_factory but this is better. Small suggestion, I can merge without.
Will take care of backporting it to the 6.x tree.

src/js/chrome/getusermedia.js
mariuszbeltowski: To be perfectly safe, we should also protect the other browser shims, but I didn't see the point. This was tested on the Google Crawler as the issue described, where the `navigator` is present. I can add the additional check suggested by you if you prefer :)

fippo: nah, without navigator this will probably blow up in many other places

fippo:
(will merge in the evening)

src/js/chrome/getusermedia.js
fippo: to answer my own question: as long as this is used from adapter_core, navigator will always be defined.

mariuszbeltowski: I will better add some unit tests, to prevent the issue appear again

mariuszbeltowski:
In the meantime, I've added unit tests to cover this issue

fippo (approved):
I very much approve more tests!

fippo:
can you take a look at https://github.com/webrtcHacks/adapter/tree/backport-6.4.7 please? I just took the RTCPeerConnection things because the getUserMedia stuff there because it only assume webkitGetUserMedia exists which is covered by the outside code in adapter_factory

PR #901, opened 2018-11-14
Add eslint rule to enforce ASCII-only identifiers

legastero:
**Description**

Adds eslint rule to enforce that all JS identifiers are ASCII-only.

**Purpose**

Prevent the introduction of confusable Unicode characters, as in https://github.com/webrtcHacks/adapter/issues/900


fippo (approved):

fippo:
many thanks!

PR #902, opened 2018-11-15
Fix broken Edge getDisplayMedia Shim

kekkokk:
broken from version 6.4.5, throws "Invalid calling object"


fippo:
meh... those non-master versions... I should make a full release soon!

Thank you!

PR #907, opened 2018-11-24
bump version to 7.0.0

fippo:


fippo:
actually need to write a changelog for this one...
* adapter is now using ES6. This is a major change and probably broke stuff
* navigator.getDisplayMedia shim is gone. Only navigator.mediaDevices.getDisplayMedia survives.
* getStreamById is gone from Safari shim. It is gone from Chrome too.
* removed lots of cruft from Chrome and Firefox shims. If you need support for browsers which are older than two years stick with the old version

fippo:
this has been working on appearin for a couple of days without much issues. Build system needed to be tweaked a bit but this is to be expected with a major bump.

PR #912, opened 2018-12-02
edge: fix STUN/TURN filtering

### LINKED ISSUE #911: Silent turn servers filtering without "transport=udp" causing harm on Edge 42.17134.1.0
- [x] I have provided steps to reproduce

## Versions affected

**Browser name including version**
Microsoft Edge 42.17134.1.0 (menu -> Settings -> About this app)

**adapter.js**
git 55eb749

## Description

The `filterIceServers` function is currently filtering all turn server without the `?transport=udp` suffix. On the stated Edge version it seems the filter does no longer apply. Even if the filter still applies, filtering without any provided feedback (in the form of console logging) is harmful and should be changed to be more informative.

## Steps to reproduce
- Configure any example using adapter.js needing turn servers;
- Configure a RTCPeerConnection with a RTCConfiguration with a turn server without `?transport=udp`;
- The turn servers are filtered without any feedback and the connection doesn't establish.

In my code, I then apply the following patch to adapter.js:
```
diff --git a/release/adapter.js b/release/adapter.js
index be8efeaa..a1c66e7e 100644
--- a/release/adapter.js
+++ b/release/adapter.js
@@ -1617,7 +1617,7 @@ function filterIceServers(iceServers, edgeVersion) {
         urls = [urls];
       }
       urls = urls.filter(function (url) {
-        var validTurn = url.startsWith('turn') && !url.startsWith('turn:[') && url.includes('transport=udp') && !hasTurn;
+        var validTurn = url.startsWith('turn') && !url.startsWith('turn:[') && !hasTurn;
 
         if (validTurn) {
           hasTurn = true;
@@ -2634,7 +2634,6 @@ function filterIceServers(iceServers, edgeVersion) {
       }
       urls = urls.filter(function(url) {
         var validTurn = url.indexOf('turn:') === 0 &&
-            url.indexOf('transport=udp') !== -1 &&
             url.indexOf('turn:[') === -1 &&
             !hasTurn;
 
```
And the connection, at least in my case, establish correctly. By the way: why the function `filterIceServers` is duplicated?

## Expected results
The `filterIceServers` function should:
- (if the filter doesn't apply anymore on the stated Edge version) selectively allow turn servers without `?transport=udp` suffix  to work;
- warn when any ice server (turn or stun, doesn't matter) is removed, to not cause harm and unexpected behavior.

## Actual results

The filter works silently without any user provided feedback, letting correctly defined connections to fail silently with Edge browser.

fippo:
by filtering STUN ourselves again. Also adds logging.
Fixes #911

### DIFF
--- a/src/js/edge/filtericeservers.js
+++ b/src/js/edge/filtericeservers.js
@@ -28,17 +28,19 @@ export function filterIceServers(iceServers, edgeVersion) {
         urls = [urls];
       }
       urls = urls.filter(url => {
+        // filter STUN unconditionally.
+        if (url.indexOf('stun:') === 0) {
+          return false;
+        }
+
         const validTurn = url.startsWith('turn') &&
             !url.startsWith('turn:[') &&
-            url.includes('transport=udp') &&
-            !hasTurn;
-
-        if (validTurn) {
+            url.includes('transport=udp');
+        if (validTurn && !hasTurn) {
           hasTurn = true;
           return true;
         }
-        return url.indexOf('stun:') === 0 && edgeVersion >= 14393 &&
-            url.indexOf('?transport=udp') === -1;
+        return validTurn && !hasTurn;
       });
 
       delete server.url;
--- a/src/js/edge/edge_shim.js
+++ b/src/js/edge/edge_shim.js
@@ -72,9 +72,9 @@ export function shimPeerConnection(window) {
       browserDetails.version);
   window.RTCPeerConnection = function(config) {
     if (config && config.iceServers) {
-      // TODO: pass browserDetails.version again. See #903 for
-      //       why this is not done.
-      config.iceServers = filterIceServers(config.iceServers);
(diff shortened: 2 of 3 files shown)


src/js/edge/filtericeservers.js
```diff
@@ -37,8 +38,14 @@ export function filterIceServers(iceServers, edgeVersion) {
           hasTurn = true;
           return true;
         }
+        if (url.startsWith('turn:')) {
+          return false;
```
dagingaa: Should this be true?

fippo: oddly locally I get test failures... and the logic seems off...

fippo:
please take another look. I made the logic a bit more obvious.

dagingaa (approved):

fippo:
published as 7.0.1

PR #917, opened 2018-12-10
Chrome: fix addTrack, removeTrack

Krivega:
**Description**
chrome_shim.js:490 Uncaught TypeError: Cannot read property 'shimAddTrackRemoveTrackWithNative' of undefined

**Purpose**
Fix it

### DIFF
--- a/src/js/chrome/chrome_shim.js
+++ b/src/js/chrome/chrome_shim.js
@@ -410,7 +410,7 @@ export function shimAddTrackRemoveTrack(window) {
   // shim addTrack and removeTrack.
   if (window.RTCPeerConnection.prototype.addTrack &&
       browserDetails.version >= 65) {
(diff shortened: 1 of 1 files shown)


dagingaa:
Pushed the wrong button :(

dagingaa (approved):
Thank you for this! `this` here is confusing, but wasn't caught by us since webpack does some magic object-tricks with export currently.

fippo:
out of curiosity, where/how did this happen? Surprised this wasn't caught earlier.

regression in 7.0.0, I don't think its important enough to backport from 7.1.0 but we could for the sake of Dag-Inge training :-)

Krivega:
@fippo just rebuilt my project (CRA) and checked on the latest version of chrome (71.0.3578.80). And indeed, there was no such problem before; apparently, the condition `if (window.RTCPeerConnection.prototype.addTrack` had not passed in chrome before.

PR #923, opened 2019-01-15
Fix overconstraint error due to undefineds on safari

Mazuh:
**Description**
On the latest build, without this PR, this following code breaks on Safari but works fine on other browsers:

```html
<video id="video" width="250px" autoplay autoplay></video>
```

```js
        var constraints = {
            audio: false,
            video: {
                width: {
                    min: undefined,
                    max: undefined,
                },
                height: {
                    min: undefined,
                    max: undefined,
                },
            },
        };

        navigator.mediaDevices.getUserMedia(constraints).then(function(stream) {
            document.getElementById('video').srcObject = stream;
        }).catch(console.error);
```

It throws errors about overconstrained `undefined` dimensions only for Safari, so I'm cleaning the video constraints from any `undefined` value in its object.

**Purpose**
Fix exactly the scenario I described without breaking any other use case for `getUserMedia` or trying to search/cover other constraint types. It is a very different behavior between browsers, and I'm trying to see what happens in this first contribution mine.

### DIFF
--- a/src/js/safari/safari_shim.js
+++ b/src/js/safari/safari_shim.js
@@ -189,6 +189,15 @@ export function shimCallbacksAPI(window) {
 export function shimGetUserMedia(window) {
   const navigator = window && window.navigator;
 
+  if (navigator.mediaDevices && navigator.mediaDevices.getUserMedia) {
+    // shim not needed in Safari 12.1
+    const mediaDevices = navigator.mediaDevices;
+    const _getUserMedia = mediaDevices.getUserMedia.bind(mediaDevices);
+    navigator.mediaDevices.getUserMedia = (constraints) => {
+      return _getUserMedia(shimConstraints(constraints));
+    };
+  }
+
   if (!navigator.getUserMedia && navigator.mediaDevices &&
     navigator.mediaDevices.getUserMedia) {
     navigator.getUserMedia = function(constraints, cb, errcb) {
@@ -196,6 +205,17 @@ export function shimGetUserMedia(window) {
       .then(cb, errcb);
     }.bind(navigator);
   }
+}
+
+export function shimConstraints(constraints) {
+  if (constraints && constraints.video !== undefined) {
+    return Object.assign({},
+      constraints,
+      {video: utils.compactObject(constraints.video)}
+    );
+  }
+
+  return constraints;
 }
 
 export function shimRTCIceServerUrls(window) {
--- a/src/js/utils.js
+++ b/src/js/utils.js
@@ -171,3 +171,25 @@ export function detectBrowser(window) {
 
   return result;
 }
+
+/**
+ * Remove all empty objects and undefined values
+ * from a nested object -- an enhanced and vanilla version
+ * of Lodash's `compact`.
+ */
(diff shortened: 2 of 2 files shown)


jan-ivar (approved):
Lgtm. The only issue is this seems necessary in 12.0.1 and earlier but appears fixed in 12.1, so maybe add a comment to that effect?

We could also do a version check, but the safari version numbers seem too coarse. `if (browserDetails.version < 13)` maybe?

fippo:
cc @youennf

Mazuh:
I'm really glad it helps, folks.

@jan-ivar, want me to code this you described on this PR? Or just I let it as it is (since it's already approved) and open another issue/PR?

jan-ivar:
If you want to update this PR that wfm.

Mazuh:
I updated the code to make it cleaner (no removed lines, just new ones, and separated the shim layer for constraints in a different function).

I tried to do what you wasked, @jan-ivar, but my `browserDetails.version` shows `605` (should it be `12.0.2`?)... So I don't know what to do about this version issue. Should I patch only for `<= 605` instead?

jan-ivar:
That's probably the [webkit version](https://stackoverflow.com/questions/7132845/which-safari-version-corresponds-to-which-webkit-version#7132890).

jan-ivar:
@youennf probably knows the right number here.

youennf:
That is indeed the webkit version.
Note though that recents STP might not show any update in the WebKit version from navigator.userAgent.
Moving to OS version and browser version might be better.

jan-ivar:
Ok thanks. In the interest of getting this landed, maybe a comment will do. The shim should be harmles after all.

Mazuh:
Yeah, it's pretty reasonable to think that a key with undefined value is not something to consider at all. So, what should I do here?

jan-ivar:
@Mazuh Please add a code comment that "this is fixed in 12.1" or similar.

Mazuh:
did it.

I could do a jsdoc-like if needed.

jan-ivar:
This looks sufficient, thanks! One tip: In the future please add updates in new commits so I don't have to review everything again.

Mazuh:
sure, thanks, @jan-ivar, just did it because people often ask me to squash commits on other free repos
(and if you click on that 'force-pushed' word on githubs interface you can still see each change I made), but i'll keep that in mind next time I contribute here!

glad to help!

jan-ivar:
@Mazuh thanks, I didn't know that! And thanks for this contribution!

fippo:
Part of the 7.2.0  release (once we figure out why npm is not letting me publish), thank you!

Mazuh:
Great, I'm waiting for this release to fix another issue on FreeSWITCH project.

kekkokk:
It happens to me with Safari 12.1.1 that if min and max are specified in the video constraints of width or height, the gum fail with an overconstraint error:

[CodePen](https://codepen.io/kekkokk/pen/YzKQzqL)

this seems to be fixed in 12.1.2

The only way to make it works reliably is to only set ideal constraints.
It is a known bug?

Issue #863, opened 2018-08-03
ignore addIceCandidate with empty candidate string


fippo:
we should ignore addIceCandidate where candidate is the empty string which is the new way to signal end-of-candidates.

No browsers emit this yet. Main problem is determining when to stop ignoring this...

jan-ivar:
The best we can do here I think is to make guesses as to when browsers might fix this, and ignore for versions older than that guess, and then keep updating adapter until browsers fix this. :-(

fippo:
yeah. we might consider doing this only when we know a fix version... that way we at least remove the backward compat worry

fippo:
Chrome 74 (?) just started not throwing on candidate: "" so we might ignore it in Chrome < 74 for backward compat.

fippo:
Hrm... so Firefox Nightly just started generating those candidates (h/t @docfaraday).
Afaics Chrome still/again (?!) throws on
```
pc.addIceCandidate({sdpMLineIndex: 0, candidate: ""})
```

fippo:
#983 fixed this for Chrome and Firefox. Interestingly, Chrome M77 still throws natively.
Remains an issue for Safari where @dagingaa filed [an issue](https://bugs.webkit.org/show_bug.cgi?id=199316)

fippo:
@youennf can you take a look at the webkit issue filed please and let me know what the plans for this are? I can't polyfill it without an "this is no longer required" indication

mpizenberg:
Just got bitten by this for Safari. The symptom was that the non-safari peer would see the safari one but not vice-versa.

hthetiot:
> Chrome 74 (?) just started not throwing on candidate: "" so we might ignore it in Chrome < 74 for backward compat.

I respectfully disagree with that statement, we logs all Javascript Errors and I saw following errors on Chrome 81+ when using webrtc-adapter, we ended up filtering empty candidate string, we also suspect that the empty candidate is due webrtc-adapter itself but we have not proven that.

> bad ice candidate (OperationError): {"type":"candidate","label":2,"id":"2","candidate":"candidate:3641010140 1 udp 2122260223 169.254.60.68 53319 typ host generation 0 ufrag u0l3 network-id 3"}

snettah:
if you get this error on Safari (14 for me)

`Unhandled Rejection (OperationError): Expect line: candidate:<candidate-str>`

check your candidate.candidate if the string il empty, I've got this error only with Safari.

```
[...]

PR #934, opened 2019-03-23
test: fix hasOwnProperty test

fippo:
chrome beta/unstable fail the ice candidate augmentation test since hasOwnProperty('address') fails even though candidate.address is set. Changes the test to avoid this.

@jan-ivar while its Chrome, do you understand this? Was the test semantics (it continues to work in Firefox) wrong?

### DIFF
--- a/test/e2e/expectations/chrome-beta
+++ b/test/e2e/expectations/chrome-beta
@@ -1,1 +1,0 @@
-ERR RTCIceCandidate with empty candidate.candidate does not throw AssertionError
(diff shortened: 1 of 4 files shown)


fippo:
the other test failure is odd. It looks like the native chrome beta/unstable ignores candidate: "" now.
```
new RTCIceCandidate({sdpMid: "foo", candidate: ""})
```
(see #863)

Lets see if that shows up in the next release notes...

fippo:
aha: https://groups.google.com/a/chromium.org/forum/?utm_medium=email&utm_source=footer#!msg/blink-dev/T6zlVFZhRVY/-bOrMI_rDQAJ
@guidou: not sure whether to clap or shake fist so... i'll do both :-)

Posting this to discuss-webrtc would have been great, I found out the not-so-nice way.

dagingaa (approved):

Issue #764, opened 2018-01-31
Chrome will be identified as Safari if webkitGetUserMedia is removed


foolip:
A continuation of https://github.com/webrtc/adapter/issues/712, or a subset thereof. After some changes this is the remaining problematic code:
https://github.com/webrtc/adapter/blob/0a293a9803dcd612cb73c147218a81d56405d6bd/src/js/utils.js#L151

This is probably blocking https://bugs.chromium.org/p/chromium/issues/detail?id=692736, as being identified as Safari might have other consequences.

KaptenJansson:
What about using something unique instead of a related WebRTC Api like getUserMedia?

Like window.chrome ?

KaptenJansson:
@fippo @alvestrand

rtzm:
I've run into this issue as well when trying to use this library for an app that I want to run on mobile. The browser when running on Chrome for iOS is being read as Safari version 604. Can /Chrome/ be read in from the User Agent with something like `navigator.userAgent.match(/Chrom(e|ium)\/(\d+)\./))`?

Otherwise, this is a really helpful library! thanks a bunch, all!

fippo:
@rtzm but in that case you want the shim code to be the safari shim even though it is chrome?

@KaptenJansson hrm... how does that behave in chrome on ios?

rtzm:
I just manually tested window.chrome and window.chrome.webstore on chrome for iOS, and both are undefined, although both should work for Chrome for desktop. Made a PR to add this feature detection here.

It also looks like detecting the correct method for getUserMedia on Chrome for iOS is moot, because iOS 11 doesn't allow Chrome to implement WebRTC: http://www.openradar.me/33571214.

zaxy78:
@rtzm all browser on iOS are built on top of (WKWebVIew with) Apple's WebKIt engine. 
Sadly, all Chrome has is a nice wrapped package for us. that's it. (see [Google's past announcement](https://blog.chromium.org/2017/01/open-sourcing-chrome-on-ios.html))

So where getUserMedia() works on Safari, it's still blocked on WKWebView..

fippo:
window.chrome.webstore does not work in Brave unfortunately which supports RTCPeerConnection (and is, no surprise, using the chrome version.
And just checking for window.chrome yields true in Edge :-(

rtzm:
haha, ok, closing my PR, since it looks like this won't be moving forward this way ¯\_(ツ)_/¯

[...]

PR #936, opened 2019-03-28
drop support for insecure contexts

### LINKED ISSUE #935: chrome 74 is identified as safari on http
a variant of #764:
Chrome 74 makes getUserMedia securecontext, i.e. not available when using http and not https. This leads to chrome being detected as Safari (due to UA sniffing, bad!)
I expect Mozilla to follow here which means Firefox detection would be broken too

One potential solution might be to not run any shims on http and detect as unknown browser there. This isn't going to work great but there are just two legitimate use-cases on http:
1) recveive-only video
2) datachannel-only
so the impact might be limited.

fippo:
getUserMedia was recently made [SecureContext] and does not exist when running on http. This lead, similar to the scenario in #764 to Chrome being detected as Safari, activating the wrong shims.

RTCPeerConnection will still be available but can only be used for datachannels and receive-only contexts where adapter doesn't offer much benefit.

Fixes #935, requires a major version bump.

### DIFF
--- a/src/js/utils.js
+++ b/src/js/utils.js
@@ -144,6 +144,11 @@ export function detectBrowser(window) {
     return result;
   }
 
+  if (window.isSecureContext === false) {
+    result.browser = 'Insecure Contexts are no longer supported.';
+    return result;
+  }
+
   if (navigator.mozGetUserMedia) { // Firefox.
     result.browser = 'firefox';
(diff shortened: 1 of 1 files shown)


dagingaa (approved):

jan-ivar (approved):

fippo:
see #942 for an alternative approach

PR #938, opened 2019-03-29
Fix sctpInDescription shim to not TypeError on rollback.

jan-ivar:
Fixes https://github.com/webrtcHacks/adapter/issues/937

### DIFF
--- a/src/js/common_shim.js
+++ b/src/js/common_shim.js
@@ -78,6 +78,9 @@ export function shimMaxMessageSize(window) {
   }
 
   const sctpInDescription = function(description) {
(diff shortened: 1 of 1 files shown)


fippo (approved):

lgrahl (commented):
I'm not really familiar with rollbacks but, other than that, fine by me.

jan-ivar:
Here's an [example](https://stackoverflow.com/a/55455764/918910). All that matters here is not tripping over the absence of SDP:
```js
await pc.setLocalDescription({type: "offer"});
```

fippo:
will ride along with next patch version

PR #948, opened 2019-04-11
Safari: Only compact objects in the constraints.

eireksten:
**Description**
Modified the isObject check in compactObject to properly identify objects.

**Purpose**
Adding the advanced property to the constraints object caused an error in Safari, since compactObject converted it to an object with keys 0, 1 and so on. This update fixes the isObject check so that only proper objects are compacted.

Also, passing in null would cause an error to be thrown previously (since isObject was true)

### DIFF
--- a/src/js/utils.js
+++ b/src/js/utils.js
@@ -177,23 +177,32 @@ export function detectBrowser(window) {
 }
 
 /**
+ * Checks if something is an object.
+ *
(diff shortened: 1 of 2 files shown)


fippo:
@dagingaa can you take a look?

dagingaa (approved):

PR #954, opened 2019-04-27
detect safari/unified-plan

fippo:
Adds a helpful property for detection of unified plan as described in
  https://trac.webkit.org/changeset/236144/webkit/


dagingaa (approved):

fippo:
I assume you'll do the next patch/minor version ;-)

PR #960, opened 2019-05-07
test: change msid test

### LINKED ISSUE #958: msid test failure in FF nightly
https://travis-ci.org/webrtcHacks/adapter/jobs/528222710#L1532
shows that [this test](https://github.com/webrtcHacks/adapter/blob/master/test/e2e/msid.js#L56) fails in Firefox Nightly (68.0a1)

@jan-ivar why did that change? I know we can't and should not rely on track ids anymore but so far this worked locally so I'd like to understand the rationale before "fixing" that test.

fippo:
changes the msid test since Firefox no longer puts the track id into the local sdp
Fixes #958

(FF nightly test is still going to fail due to #959)

### DIFF
--- a/test/e2e/msid.js
+++ b/test/e2e/msid.js
@@ -53,7 +53,7 @@ describe('MSID', () => {
     });
   });
 
-  it('puts the track msid attribute into the localDescription', () => {
+  it('puts the stream msid attribute into the localDescription', () => {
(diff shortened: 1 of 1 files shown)


fippo:
FF nightly test is green again.

dagingaa (approved):

Issue #956, opened 2019-05-03
iceConnectionState is busted in Chrome 75(beta), leading to test failures


fippo:
see https://bugs.chromium.org/p/chromium/issues/detail?id=959128
-- tests are failing

fippo:
also: lets make tests mandatory in canary (and ff nightly) again

fippo:
this seems fixed but now there are new failures both in unstable and beta, largely due to SCTP

PR #961, opened 2019-05-14
test: fail if FF nightly fails

fippo:
nightly seems to be pretty stable these days. Failing in nightly will point out changes like #960 faster

@jan-ivar I assume that I can always nag you when things start failing :-)


dagingaa (approved):

fippo:
Nightly passed, test failure is due to #956

Issue #964, opened 2019-05-17
chrome beta/unstable test failures


fippo:
we seem to have new failures in unstable in beta now that the ICE issues are gone:
```
+ERR addIceCandidate "before each" hook for "resolves when called with null" timeout
+ERR maxMessageSize send largest possible single message AssertionError
+ERR maxMessageSize throws an exception if the message is too large AssertionError
+ERR maxMessageSize throws an exception if the message is too large (using a secondary data channel) AssertionError
```
@lgrahl can you take a look at the datachannel ones please? I'll take a look at why addIceCandidate(null) fails while addIceCandidate(undefined) still passes

fippo:
wrt addIceCandidate I've pulled the test into a fiddle:
https://jsfiddle.net/ujoLxd56/
-- this fails with both null and undefined while the karma test only fails with null.
With a local adapter build it looks like the main addIceCandidate shim is not called.
Bad revision: 650671
Good revision:  638880
Bisect:
```
python bisect-builds.py --use-local-cache -a linux64 -b 650671 -g 638880 --user-data-dir=/tmp https://jsfiddle.net/ujoLxd56/
```
(test: does console log an error?) points to
https://chromium.googlesource.com/chromium/src/+log/59490dd45840550cc25232167d13d30795fd9f85..ee949cd4fef997e3878cfba822535c2614174951

https://github.com/webrtcHacks/adapter/blob/master/test/e2e/addIceCandidate.js#L37-L38
adds a candidate without sdpMLineIndex. 

@guidou: I don't understand why this would throw in one case but not the other?!
Either way, I think addIceCandidate with just sdpMid is ok. cc-ing @alvestrand for a second opinion.

nils-ohlmeier:
AFAIK the spec says you have to have either sdpMLineIndex or sdpMid set. So adding an ICE candidate without MLineIndex should work. Isn't there a WPT test which should cover this?

fippo:
there is a test named `Add candidate with only valid sdpMid should succeed` which passes.
From webrtc-internals the difference is that in the fiddle sdpMLineIndex is null while in the wpt test it is 0.
The difference here is adapter explicitly calling new RTCIceCandidate. Changing the test to
```
    return pc.setRemoteDescription(sessionDesc)
      .then(() => pc.addIceCandidate(new RTCIceCandidate({
        candidate: candidateStr1,
        sdpMid: sdpMid1 })));
[...]

PR #968, opened 2019-05-23
test: update chrome beta (with experimental platform features flag) tests

fippo:
to allow some failures. See #964

(this should make things green again)


fippo:
all green!

Issue #970, opened 2019-05-31
Please point to a common-js module in package.json's "main" field


cam-narzt:
- [x] I have provided steps to reproduce (e.g. a link to a [jsfiddle](https://jsfiddle.net/))
- [x] I have provided browser name, version and adapter.js version
- [x] This issue only happens when adapter.js is used

## Versions affected

**Browser name including version (e.g. Chrome 64.0.3282.119)**
N/A (testing environment).

**adapter.js (e.g. 6.1.0)**
`webrtc-adapter` node module version 7.2.3

## Description
The `main` field in `package.json` does not point to a common-js module, rather an es6 module. This means that components that import it cannot be tested in a `create-react-app` app without "ejecting" the app, which is a lot of work.

More discussion here: https://github.com/facebook/create-react-app/issues/7100

## Steps to reproduce
run create-react-app to make a boiletplate project: `npx create-react-app test && cd test`
add webrtc-adapter npm module to project: `npm install webrtc-adapter`
add adapter to a component: `echo "import adapter from 'webrtc-adapter';" | cat - src/App.js | tee src/App.js`
write a basic test which tests the component that imported 'webrtc-adapter' (done already for you in boilerplate)
run tests `npm test`

## Expected results
tests run

## Actual results
tests blow up because the environment doesn't compile the es6 module to a common-js module.

```
  ● Test suite failed to run

    Jest encountered an unexpected token

    This usually means that you are trying to import a file which Jest cannot parse, e.g. it's not plain JavaScript.

    By default, if Jest sees a Babel config, it will use that to transform your files, ignoring "node_modules".

    Here's what you can do:
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.

    You'll find more details and examples of these config options in the docs:
    https://jestjs.io/docs/en/configuration.html

    Details:

    /Users/me/Developer/Javascript/test/node_modules/webrtc-adapter/src/js/adapter_core5.js:12
    import {adapterFactory} from './adapter_factory.js';
           ^

    SyntaxError: Unexpected token {

    > 1 | import adapter from 'webrtc-adapter';
        | ^
      2 | import React from 'react';
      3 | import logo from './logo.svg';
      4 | import './App.css';

[...]

PR #975, opened 2019-06-15
point to dist/adapter_core.js in package.json main

fippo:
points main in package.json to the es5-transpiled version of adapter.
Partial fix for #970


fippo:
@legastero says this works for him. It does for me using the steps in #970

PR #978, opened 2019-06-17
ensure shim’s name property matches original function

cam-narzt:
**Description**
Adds function names to shim functions at point of definition.

**Purpose**
This is just a small change to make the shims act a little more like the original functions. As is if you call `RTCIceCandidate.name` on an unshimmed `RTCIceCandidate`, you get the string `"RTCIceCandidate"` but on a shimmed one you get `""`.

### DIFF
--- a/src/js/chrome/chrome_shim.js
+++ b/src/js/chrome/chrome_shim.js
@@ -560,7 +560,7 @@ export function shimAddTrackRemoveTrack(window) {
   }
   ['createOffer', 'createAnswer'].forEach(function(method) {
     const nativeMethod = window.RTCPeerConnection.prototype[method];
-    window.RTCPeerConnection.prototype[method] = function prototype[method]() {
+    const methodObj = {[method]() {
       const args = arguments;
       const isLegacyCall = arguments.length &&
           typeof arguments[0] === 'function';
@@ -579,7 +579,8 @@ export function shimAddTrackRemoveTrack(window) {
       }
       return nativeMethod.apply(this, arguments)
       .then(description => replaceInternalStreamId(this, description));
-    };
+    }};
+    window.RTCPeerConnection.prototype[method] = methodObj[method];
   });
 
   const origSetLocalDescription =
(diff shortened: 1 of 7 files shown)


cam-narzt:
The patch was generated with this command: 
`rg 'window(\.|\[).*=.*function *\(' -l src/ | xargs -n1 -I{} sed -i '' -Ee 's/(.*\.([^\.[:space:]]+)[[:space:]]=)(.*)(function) *\(/\1\3\4 \2\(/g' {}`

Documentation on the name property: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/name

I admit to not having tested this locally, as I wanted to get feedback before spending too much time on this. I ran into the difference in behaviour while writing a tracing proxy of the WebRTC API functions to make sure my tests accurately reflect real browser behaviour.

fippo:
There are some ```function prototype[method]()``` which the linter doesn't like, otherwise this looks good to me. It increases readability of stack traces, right?

CamJN:
Yeah, it helps with stack traces, as well as any debugging/tracing libraries you may be using.

cam-narzt:
Yeah, it helps with stack traces, as well as any debugging/tracing libraries you may be using. 

(Sorry for the double post, used the wrong account).

src/js/chrome/chrome_shim.js
```diff
@@ -666,7 +666,7 @@ export function shimPeerConnection(window) {
     ['setLocalDescription', 'setRemoteDescription', 'addIceCandidate']
         .forEach(function(method) {
           const nativeMethod = window.RTCPeerConnection.prototype[method];
-          window.RTCPeerConnection.prototype[method] = function() {
+          window.RTCPeerConnection.prototype[method] = function prototype[method]() {
```
fippo: this breaks (a couple of times). Not sure if it makes sense to "unroll" this (and other) loop in order to ease debugging

fippo (changes requested):

src/js/chrome/chrome_shim.js
cam-narzt: I pushed a change to fix this to real syntax, but I'm not sure if it will actually set the function name property, so it's a bit of an experiment.

fippo:
now the linter complaints about line length and a couple of variable name conflicts in the safari shim. Its probably best to rename the variables to origSomething as its done elsewhere.

fippo:
git diff -w to the rescue. Nice way of dealing with the iterated methods!
Will go into next release (hopefully this week)

Thank you!

fippo (approved):

PR #979, opened 2019-06-24
fix chrome pc.sctp in plan-b mode

fippo:
tentative fix

### DIFF
--- a/src/js/common_shim.js
+++ b/src/js/common_shim.js
@@ -64,7 +64,7 @@ export function shimRTCIceCandidate(window) {
 }
 
 export function shimMaxMessageSize(window) {
-  if (window.RTCSctpTransport || !window.RTCPeerConnection) {
+  if (!window.RTCPeerConnection) {
     return;
   }
   const browserDetails = utils.detectBrowser(window);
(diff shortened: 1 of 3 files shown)


src/js/common_shim.js
```diff
@@ -164,6 +164,20 @@ export function shimMaxMessageSize(window) {
       window.RTCPeerConnection.prototype.setRemoteDescription;
   window.RTCPeerConnection.prototype.setRemoteDescription = function() {
     this._sctp = null;
+    // Chrome unilaterally decided to not expose .scpt in plan-b mode.
+    // As usual, adapter.js has to do the 'ugly hack' to cover up the mess.
+    if (browserDetails === 'chrome' && ('sctp' in this)) {
```
dagingaa: Should this be negated for sctp?

```suggestion
    if (browserDetails === 'chrome' && !('sctp' in this)) {
```

fippo: the point is to do this once, even in plan-b mode. But it doesn't work as intended...

src/js/common_shim.js
```diff
@@ -164,6 +164,21 @@ export function shimMaxMessageSize(window) {
       window.RTCPeerConnection.prototype.setRemoteDescription;
   window.RTCPeerConnection.prototype.setRemoteDescription = function() {
     this._sctp = null;
+    // Chrome unilaterally decided to not expose .scpt in plan-b mode.
```
lgrahl:
```suggestion
    // Chrome unilaterally decided to not expose .sctp in plan-b mode.
```

lgrahl (dismissed):
IIRC this won't work since https://github.com/webrtcHacks/adapter/pull/979/files#diff-26e41bf001b9f7daa490e681199d92edR67 returns early.

The code should be the same for plan B. Only the above mentioned check needs to be changed to look for plan B... since you only know that once local/remote description is being applied, ugh... it will be ugly.

fippo:
Now there is an odd failure in chrome beta without experimental web platform features when sending a too-large message. This is a Chrome bug and seems to be fixed in Canary.

src/js/common_shim.js
```diff
@@ -164,6 +164,21 @@ export function shimMaxMessageSize(window) {
       window.RTCPeerConnection.prototype.setRemoteDescription;
   window.RTCPeerConnection.prototype.setRemoteDescription = function() {
     this._sctp = null;
+    // Chrome unilaterally decided to not expose .sctp in plan-b mode.
+    // As usual, adapter.js has to do an 'ugly worakaround '
```
fippo: nit: extra space

lgrahl (approved):
Since Plan-B can only be enabled locally, should be fine now.

PR #980, opened 2019-06-27
dont change getStats signature

fippo:
Do not change the getStats signature (number of arguments) since some
libraries copied the adapter.js check which results in chaos.


fippo:
(conflicts with #978)

jan-ivar (approved):
Lgtm.

PR #982, opened 2019-06-27
getStats argument length should be reported as zero

D1plo1d:
**Description**

```js
window.RTCPeerConnection.prototype.getStats.length === 0
```

**Purpose**

This fixes compatibility with simple-peer. Simple Peer only uses the newer promise-based API if getStats.length === 0.

Resolves this issue: https://github.com/feross/simple-peer/issues/470
See Also: https://github.com/webrtcHacks/adapter/issues/981


fippo:
already in #980 which is included in 7.2.5 - let me know if that doesn't fix the issue

PR #983, opened 2019-06-30
ignore addIceCandidate({candidate: ''}) in FF<68 and Chrome <78

fippo:
Ignores addIceCandidate({candidate: ''}) in Firefox up to version 68 which
emits this.

Also done in Chrome < 78, see https://bugs.chromium.org/p/chromium/issues/detail?id=978582

### DIFF
--- a/src/js/chrome/chrome_shim.js
+++ b/src/js/chrome/chrome_shim.js
@@ -702,6 +702,12 @@ export function shimPeerConnection(window) {
         }
         return Promise.resolve();
       }
+      // Firefox 68+ emits and processes {candidate: "", ...}, ignore
+      // in older versions. Native support planned for Chrome M77.
+      if (browserDetails.version < 78 &&
+        arguments[0] && arguments[0].candidate === '') {
+        return Promise.resolve();
+      }
       return nativeAddIceCandidate.apply(this, arguments);
     };
 }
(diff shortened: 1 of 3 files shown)


src/js/chrome/chrome_shim.js
```diff
@@ -702,6 +702,12 @@ export function shimPeerConnection(window) {
         }
         return Promise.resolve();
       }
+      // Firefox 68+ emits and processes {candidate: "", ...}, ignore
+      // in older versions. Native support planned for Chrome M77.
+      if (browserDetails.version < 78 &&
```
jan-ivar, line 707: `M77` != `78`. Are you adding a bit of room, just in case? That might be wise.

But, if they do end up adding native support in M77 then adapter will prevent the connection from going to `"complete"`/`"completed"` state until M78, so there's a trade-off.

jan-ivar (approved):
Thanks!

src/js/chrome/chrome_shim.js
fippo: Yes, a bit of room since its summer in Sweden and @guidou might enjoy the summer rather than fix it :-)

I assume that in the first step they will just silently ignore this at the Chrome layer. It never went to completed on the receiving end anyway

PR #987, opened 2019-07-07
safari: fix trackevent transceiver property

### LINKED ISSUE #984: Safari/iOS overwriting RTCRtpTransceiver in ontrack, missing all properties

## Versions affected
**Browser name including version**
Safari 12.1.1
Safari Technology Preview 13.0 Release 86
iOS 12.3.1

**adapter.js**
7.2.4
6.4.8

## Description
The shim function `shimTrackEventTransceiver` is completely overwriting the already present `transceiver` property on RTCTrackEvent with a brand new object consisting of only the `receiver`.

I am not sure on what version Safari finally implemented full RTCRtpTransceiver support, but it is here now. It seems `shimTrackEventTransceiver` is mostly referencing RTCTransceiver, which I guess is maybe an even older legacy class? However I'm referring to [RTCRtpTransceiver](https://developer.mozilla.org/en-US/docs/Web/API/RTCRtpTransceiver).

## Steps to reproduce
Couldn't use jsfiddle for some reason, the getUserMedia call was breaking a security rule in the iframe (despite the fact that jsfiddle has allow camera attribute iframe). So I just made these quick hosted gists. They are very simple 1 to 1 webrtc demo call you have likely seen before. Our only purpose is to display the `transceiver` object's props in the ontrack event. Examples are using adapter.js 7.2.4, but the `shimTrackEventTransceiver` function seems to have been around forever.

First the working example **without** adapter.js. You should see these keys being printed: **mid, sender, receiver, stopped, direction, currentDirection, stop**. These are the expected props you will also see in Chrome/Firefox/etc.
https://gistcdn.githack.com/Nicholi/0a85acf5cc4edc2d988fdbc94791dc18/raw/e09b542c375da399fe96432194760649ab81a819/normal.html

Now **with** adapter.js, you should find the only property found is the **receiver**.
https://gistcdn.githack.com/Nicholi/29c01bfe39b6e050ddef5f8fae2ca276/raw/673c5b7edaba9985d7b52f637ea36c9224c0ae29/bad_adapter.html

## Expected results
Expected to find all the documented properties on [RTCRtpTransceiver](https://developer.mozilla.org/en-US/docs/Web/API/RTCRtpTransceiver). My personal need is the `mid` property.

## Actual results
The `transceiver` has only 1 property, the `receiver`.

fippo:
Fixes #984.

Test Plan:
- verify that 'transceiver' in RTCTrackEvent.prototype works in current safari
- call SRD with a SDP that will trigger ontrack
- inspect track event, look at transceiver property

### DIFF
--- a/src/js/safari/safari_shim.js
+++ b/src/js/safari/safari_shim.js
@@ -267,11 +267,9 @@ export function shimRTCIceServerUrls(window) {
 
 export function shimTrackEventTransceiver(window) {
   // Add event.transceiver member over deprecated event.receiver
-  if (typeof window === 'object' && window.RTCPeerConnection &&
-      ('receiver' in window.RTCTrackEvent.prototype) &&
-      // can't check 'transceiver' in window.RTCTrackEvent.prototype, as it is
-      // defined for some reason even when window.RTCTransceiver is not.
-      !window.RTCTransceiver) {
+  if (typeof window === 'object' && window.RTCTrackEvent &&
+      'receiver' in window.RTCTrackEvent.prototype &&
+      !('transceiver' in window.RTCTrackEvent.prototype)) {
     Object.defineProperty(window.RTCTrackEvent.prototype, 'transceiver', {
       get() {
         return {receiver: this.receiver};


jan-ivar (approved):
Lgtm. Thanks!

PR #996, opened 2019-07-14
Utils can throw an exception trying to redfined a property wrapper

msarchet:
**Description**

When utils tries to define a property it can try to define an already existing property, thus throwing an exception.

I ran into this issue using the atom-hydra package

[atom-hyra related issue](https://github.com/ojack/atom-hydra/issues/13)

I can't find anything that would be lost by not redefining the event in looking at this code. Happy to address any issues that might need to be resolved.

**Purpose**

Stop utils from redefining a property that already exists.

### DIFF
--- a/src/js/utils.js
+++ b/src/js/utils.js
@@ -61,6 +61,10 @@ export function wrapPeerConnectionEvent(window, eventNameToWrap, wrapper) {
       unwrappedCb]);
   };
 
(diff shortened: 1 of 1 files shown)


fippo:
that utility function typically overrides an existing property so this won't work.
We've had those issues before and the current code uses defineProperty with configurable: true.
Looks like atom-hydra is using webrtc-adapter version two which is five major versions behind the current major and doesn't have that.

PR #997, opened 2019-07-25
Check if e.candidate is empty string in icecandidate listener

molnartimi:
We were facing this issue, if icecandidate events created with an empty string, errors occurred in other functions called by this method. This solution solved it.


fippo:
This can't be right... e.candidate is an object which has no length so you're simply skipping the augmentation.

If you have application code which breaks in Firefox 68 which generates onicecandidate with candidate.candidate equal to '' you'll need to update that code.

fippo:
I still don't think this is required so closing. Please reopen if this is still an issue!

PR #1002, opened 2019-07-31
Bump version to 7.2.9

fippo:
to no longer ignore addIceCandidate(null) in Firefox


fippo:
hold on, lockfile accident...

PR #1004, opened 2019-08-01
Firefox ignores addTransceiver init setEncodings opts

### LINKED ISSUE #998: addTransceiver Firefox ignores init options
##  Description
It seems to me that the addTransceiver function in Firefox is ignoring the sendEncodings parameter during initialization.

As per doc: [MDN](https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/addTransceiver) the constructor takes in input an init object which can contains the sendEncodings options.

When passing this object to Chrome, everything's fine
When passing this object to Firefox, it gets ignored. The only way to apply these parameters, is getting the sender configuration, re-add the sendEncodings parameters and set again the options into the sender.

This last step, however, is not very appreciated by Chrome which complaints about the fact that we are re-setting the rids which are now read only.

This was stated clearly  in this PSA [Spec-compliant Simulcast using addTransceiver API](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/discuss-webrtc/32Y1AUzQ9XU/rfzss0SQAgAJ)

> The simulcast envelope (i.e. the number of layers and their RIDs) cannot be changed and is determined in the call to addTransceiver().
> After the number of layers is determined, the RtpSenderInterface can be queried for the parameters using getParameters()
> and the configuration for layers (not their number or RIDs) can change using setParameters().

I suspect this is a Firefox bug, and if so, adapter can shim the addTransceiver function (only in Firefox) and re-apply the init parameters so we don't have to manage the different cases in our client logic.

Also, if you think this is a bug and not mine misunderstanding, I'll proceed to file a bug issue.

## Steps to reproduce
discalimer: Not an html/css developer here so please be gentle with my example page. Also note that I reused a "playground" page with lots of other things but you can find the culprit on JS line 179.

Chrome 75.0.3770.100, FF 68.0.1

[LINK TO EXAMPLE CODEPEN](https://codepen.io/anon/pen/jgBdzx)

Click get media
Click Create PeerConnection
Click Create Offer
Observe Generated offer

If you change the line 179 in if (true), then Firefox works but chrome complaints

What do you think @fippo? Am I missing something?
thanks

kekkokk:
should fix https://github.com/webrtcHacks/adapter/issues/998


src/js/firefox/firefox_shim.js
```diff
@@ -190,3 +190,79 @@ export function shimRTCDataChannel(window) {
...
+  const origAddTransceiver = window.RTCPeerConnection.prototype.addTransceiver;
+  if (origAddTransceiver) {
+    window.RTCPeerConnection.prototype.addTransceiver =
+      function addTransceiver() {
+        const initParameters = arguments[1];
+        const shouldPerformCheck = initParameters &&
+                                  'sendEncodings' in initParameters;
+        if (shouldPerformCheck) {
+          // If sendEncodings params are provided, validate grammar
+          initParameters.sendEncodings.forEach((encodgingParam) => {
```
jan-ivar: Typo. `s/encodgingParam/encodingParam/`

src/js/firefox/firefox_shim.js
```diff
@@ -190,3 +190,79 @@ export function shimRTCDataChannel(window) {
...
+          // Also note that after the createoffer we are not 100% sure that
+          // the params were asynchronously applied so we might miss the
+          // opportunity to recreate offer.
+          const sender = transceiver.sender;
+          if (sender) {
+            const params = sender.getParameters();
+            if (!('encodings' in params)) {
+              setParametersPromises.push(
+                sender.setParameters({encodings: initParameters.sendEncodings})
+                .catch(() => Promise.resolve())
```
jan-ivar: Nit: `.catch(() => {})` implicitly converts to a promise, so create one is redundant.

src/js/firefox/firefox_shim.js
```diff
@@ -190,3 +190,79 @@ export function shimRTCDataChannel(window) {
...
+        if (shouldPerformCheck) {
+          // Check if the init options were applied. If not we do this in an
+          // asynchronous way and save the promise reference in a global object.
+          // This is an ugly hack, but at the same time is way more robust than
+          // checking the sender parameters before and after the createOffer
+          // Also note that after the createoffer we are not 100% sure that
+          // the params were asynchronously applied so we might miss the
+          // opportunity to recreate offer.
+          const sender = transceiver.sender;
+          if (sender) {
```
jan-ivar: Nit: redundant `if`. All transceivers have a sender.

src/js/firefox/firefox_shim.js
```diff
@@ -190,3 +190,79 @@ export function shimRTCDataChannel(window) {
...
+}
+
+export function shimCreateOffer(window) {
+  // https://github.com/webrtcHacks/adapter/issues/998#issuecomment-516921647
+  // Firefox ignores the init sendEncodings options passed to addTransceiver
+  // https://bugzilla.mozilla.org/show_bug.cgi?id=1396918
+  const origCreateOffer = window.RTCPeerConnection.prototype.createOffer;
+  if (origCreateOffer) {
+    window.RTCPeerConnection.prototype.createOffer = function createOffer() {
+      return Promise.all(setParametersPromises)
```
jan-ivar: Looks good, except I think I'd prefer to add the following lines directly ahead of the above line:
```js
if (!setParametersPromises.length) {
  return origCreateOffer.apply(this, arguments);
}
```
That way we only disturb the timing of *createOffer*'s synchronous validation steps when necessary (mostly an [[IsClosed]] test). It's really only observable in the JS debugger (Promise appears \<rejected> in console immediately), but still.

It'd make me feel better if we didn't disturb the timing of *every* createOffer out there.

jan-ivar (approved):
Awesome, thanks!

jan-ivar:
I guess *technically* we'd want to do this in *createAnswer* too?

src/js/firefox/firefox_shim.js
```diff
@@ -190,3 +190,79 @@ export function shimRTCDataChannel(window) {
     window.RTCDataChannel = window.DataChannel;
   }
 }
+
+let setParametersPromises = [];
+export function shimAddTransceiver(window) {
+  // https://github.com/webrtcHacks/adapter/issues/998#issuecomment-516921647
```
fippo, line 195: what the test caught here is a missing
```
  if (!(typeof window === 'object' && window.RTCPeerConnection)) {
    return;
  }
```

src/js/firefox/firefox_shim.js
fippo: also: const {sender} = transceiver;

src/js/firefox/firefox_shim.js
```diff
@@ -190,3 +190,79 @@ export function shimRTCDataChannel(window) {
...
+          // checking the sender parameters before and after the createOffer
+          // Also note that after the createoffer we are not 100% sure that
+          // the params were asynchronously applied so we might miss the
+          // opportunity to recreate offer.
+          const sender = transceiver.sender;
+          if (sender) {
+            const params = sender.getParameters();
+            if (!('encodings' in params)) {
+              setParametersPromises.push(
+                sender.setParameters({encodings: initParameters.sendEncodings})
```
fippo: you need to set params.encodings and then call setParameters. setParameters should reject but doesn't in Firefox.
Do we need more here like maxBitrate?

src/js/firefox/firefox_shim.js
```diff
@@ -190,3 +190,79 @@ export function shimRTCDataChannel(window) {
...
+      };
+  }
+}
+
+export function shimCreateOffer(window) {
+  // https://github.com/webrtcHacks/adapter/issues/998#issuecomment-516921647
+  // Firefox ignores the init sendEncodings options passed to addTransceiver
+  // https://bugzilla.mozilla.org/show_bug.cgi?id=1396918
+  const origCreateOffer = window.RTCPeerConnection.prototype.createOffer;
+  if (origCreateOffer) {
```
fippo: if you add the same "does RTCPeerConnection exist" check  as above  you can skip this check

src/js/firefox/firefox_shim.js
```diff
@@ -190,3 +190,79 @@ export function shimRTCDataChannel(window) {
     window.RTCDataChannel = window.DataChannel;
   }
 }
+
+let setParametersPromises = [];
```
fippo: doesn't this need to be per peerconnection, i.e. pc._setParametersPromises? Otherwise we create dependencies between peerconnections

kekkokk:
Thanks guys for the suggestions. I reviewed all your points with my colleague @simonemazzoni
I also implemented it for the createanswer but I wasn't really able to test so any help would be appreciated.

src/js/firefox/firefox_shim.js
kekkokk: need some info from @jan-ivar for this topic

src/js/firefox/firefox_shim.js
kekkokk: Totally right, I did not think about it and I thought that was not a nice thing to "inject" parameters in the PeerConnection if possible. But in this case I think is the only solution

fippo:
I wonder if we should have a e2e test along the lines of:
1) call pc.addTransceiver with a video track and sendEncodings
2) call pc.createOffer()
3) check that the offer contains an a=simulcast line.

We would need to make testable in Chrome too since that is how the e2e tests work... but Chrome generates different SDP (slightly)

kekkokk:
Tried to implement a basic e2e test, never did this so be patient. Feel free to rewrite from scratch
thanks @fippo 

ps. I created a new simulcast.js in which we can also add the e2e test for enabling simulcast via sdp munging

test/e2e/simulcast.js
```diff
@@ -0,0 +1,51 @@
...
+    pc1 = new RTCPeerConnection(null);
+  });
+  afterEach(() => {
+    pc1.close();
+  });
+
+  it('using transceivers APIs', (done) => {
+    if (window.adapter.browserDetails.browser === 'edge' ||
+      window.adapter.browserDetails.browser === 'safari') {
+      this.skip();
```
fippo, line 24: this isn't defined in arrow functions (the general recommendation is not to use arrow functions but...) so you'll need to make the it a regular function.

test/e2e/simulcast.js
```diff
@@ -0,0 +1,51 @@
...
+    }
+    var constraints = {video: true};
```
fippo: const

test/e2e/simulcast.js
```diff
@@ -0,0 +1,51 @@
...
+    navigator.mediaDevices.getUserMedia(constraints)
+      .then(function(stream) {
+        const initOpts = {
+          sendEncodings: [
+            {'rid': 'high'},
```
fippo: {rid: 'high'} et al

test/e2e/simulcast.js
```diff
@@ -0,0 +1,51 @@
...
+  });
+
+  it('using transceivers APIs', (done) => {
+    if (window.adapter.browserDetails.browser === 'edge' ||
+      window.adapter.browserDetails.browser === 'safari') {
+      this.skip();
+    }
+    var constraints = {video: true};
+    navigator.mediaDevices.getUserMedia(constraints)
+      .then(function(stream) {
```
fippo: arrow function please

test/e2e/simulcast.js
```diff
@@ -0,0 +1,51 @@
...
+        const initOpts = {
+          sendEncodings: [
+            {'rid': 'high'},
+            {'rid': 'medium', 'scaleResolutionDownBy': 2},
+            {'rid': 'low', 'scaleResolutionDownBy': 3}
```
fippo: i think you need 4 here?

test/e2e/simulcast.js
```diff
@@ -0,0 +1,51 @@
...
+        pc1.addTransceiver(stream.getVideoTracks()[0], initOpts);
+
+        pc1.createOffer().then((offer) => {
+          const simulcastRegex = /a=simulcast:.*send .*high;medium;low/g;
+          expect(simulcastRegex.test(offer.sdp)).to.equal(true);
+          done();
+        })
+        .catch(throwError);
+      })
+      .catch(throwError);
```
fippo: I think instead of the done-throwError you can return a promise chain from GUM to createOffer and the result of this last check

src/js/firefox/firefox_shim.js
jan-ivar: Yeah that line deviated from what I suggested in https://github.com/webrtcHacks/adapter/issues/998#issuecomment-516921647 but I didn't flag it, because this code would only ever run in a Firefox without a proper *setParameters* implementation, so it seemed fine. 

That said, @fippo is right the following would be more correct (avoids suppressing other init options outside of encodings for no reason—Firefox just doesn't support any of them atm):
```js
setParameters(Object.assign({}, init, {encodings: initParameters.sendEncodings}))
```

`maxBitrate` is already part of [RTCRtpEncodingParameters](https://w3c.github.io/webrtc-pc/#dom-rtcrtpencodingparameters) so we should be good there.

src/js/firefox/firefox_shim.js
jan-ivar: Good catch! I missed that.

jan-ivar: > I thought that was not a nice thing to "inject" parameters in the PeerConnection if possible

Could we maybe modify the patch to create it on-demand, i.e. only if `sendEncodings` is actually used?

test/e2e/simulcast.js
kekkokk: Seems that setting to 4 breaks chrome 77 dev.
Did we found a bug? XD

I tried to also make a GUM in 1080p but it gives me only the first 3 layers in 77 dev.

kekkokk: seems that Webrtc is switching from a max of 4 to a max of 3 in m77

[m75](https://chromium.googlesource.com/external/webrtc/+/branch-heads/m75/api/video/video_codec_constants.h)

[m77](https://chromium.googlesource.com/external/webrtc/+/branch-heads/m77/api/video/video_codec_constants.h)

so... idk. I revert the tests from 4 to 3?

test/e2e/simulcast.js
```diff
@@ -0,0 +1,46 @@
...
+        pc1.addTransceiver(stream.getVideoTracks()[0], initOpts);
+
+        return pc1.createOffer().then((offer) => {
+          const simulcastRegex =
+          /a=simulcast:[\s]?send (?:rid=)?original;high;medium;low/g;
+          return expect(simulcastRegex.test(offer.sdp)).to.equal(true);
+        });
+      });
+  });
+});
```
jan-ivar: Add newline.

jan-ivar (approved):
This looks great! Did it stall?

fippo:
no, now its blocked by chrome canary failing for some reason. Arguably that is unrelated to Firefox so...

kekkokk:
Yeah seems that new webrtc codebase does not accept more than 3 layers, so it fails on canary. 
https://github.com/webrtcHacks/adapter/pull/1004#discussion_r310560149

How do we proceed?

fippo:
yeah, lets go for 3 layers and 2/4 scaledown

@orphis: you might want to check the above is intended if anyone uses this.

test/e2e/simulcast.js
Orphis: We're not really switching, the 4th layer would never be used because of some internal limiting of simulcast layer count. Trying to remove that limitation showed that the 4th one was unusable for other reasons, so we properly limit it to 3 now (until we fix the other issue).

test/e2e/simulcast.js
```diff
@@ -0,0 +1,46 @@
...
+    }
+    const constraints = {video: true};
+    return navigator.mediaDevices.getUserMedia(constraints)
+      .then((stream) => {
+        const initOpts = {
+          sendEncodings: [
+            {rid: 'original'},
+            {rid: 'high', scaleResolutionDownBy: 2},
+            {rid: 'medium', scaleResolutionDownBy: 3},
+            {rid: 'low', scaleResolutionDownBy: 4}
```
Orphis, line 33: libwebrtc doesn't support more than 3 simulcast layers, and so does Chrome. You should probably reduce the layer count here.

Orphis (changes requested):
You should only use 3 simulcast layers in your test at most.

fippo:
>           /a=simulcast:[\s]?send (?:rid=)?original;high;medium;low/g;
One last thing to push it over the finish line :-)

kekkokk:
> > ```
> >       /a=simulcast:[\s]?send (?:rid=)?original;high;medium;low/g;
> > ```
> 
> One last thing to push it over the finish line :-)

ooops, I'm on vacation and I'm totally not focused XD

fippo:
will publish 7.3.0 soonish. Enjoy your vacation!

PR #1009, opened 2019-09-01
test: update travis base image

fippo:
uses bionic (ubuntu 18.04) instead of trusty


fippo:
of course, once failure is allowed it no longer happens...

PR #1015, opened 2020-01-08
Bump version to 7.4.0

fippo:
which adds the typescript definition


fippo:
FF nightly failure only happens on travis :man_shrugging:

PR #1021, opened 2020-02-11
Fixing 'Cannot read property 'length' of undefined'

llKoull:
**Description**
PR to fix the issue "Cannot read property 'length' of undefined" https://github.com/webrtcHacks/adapter/issues/1020

### DIFF
--- a/src/js/chrome/chrome_shim.js
+++ b/src/js/chrome/chrome_shim.js
@@ -676,7 +676,8 @@ export function shimPeerConnection(window) {
   }
 
   const addIceCandidateNullSupported =
-    window.RTCPeerConnection.prototype.addIceCandidate.length === 0;
+      window.RTCPeerConnection.prototype.addIceCandidate != null &&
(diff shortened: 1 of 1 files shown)


OlivierJM (changes requested):
You need to change the `!=` to `!==` otherwise eslint won't let that pass and also since the value to check is `null` it should still work fine.

OlivierJM:
@fippo bringing this to your attention, we are also having a similar issue with the adapter.

llKoull:
Ok, i did the change for eslint verification ;)

OlivierJM (approved):
Thanks for making the changes, I hope this gets merged soon.

fippo:
I would still like to understand *why* the errors happens -- in particular when checking for null helps when it should at most be undefined.

ksocha:
@fippo After changing `!=` to `!==` as @OlivierJM requested this solution would not work.

fippo:
overtaken by #1066

Issue #1023, opened 2020-02-26
 TypeError: 'arguments', 'callee', and 'caller' cannot be accessed in this context, when using addTrack() method


dimitrisniras:
# Please read first!
Please use [discuss-webrtc](https://groups.google.com/forum/#!forum/discuss-webrtc) for general technical discussions and questions.

- [x] I have provided steps to reproduce (e.g. a link to a [jsfiddle](https://jsfiddle.net/))
- [x] I have provided browser name, version and adapter.js version
- [x] This issue only happens when adapter.js is used

**Note: If the checkboxes above are not checked (which you do after the issue is posted), the issue will be closed.**

## Versions affected

**Browser name including version (e.g. Chrome 64.0.3282.119)**

Safari Version: 13.0.5 and all 13.x.x versions

**adapter.js (e.g. 6.1.0)**

Version 7.5.0

## Description

When trying to use the `RTCPeerConnection.addTrack(track)` method, I am receiving this error

```
 TypeError: 'arguments', 'callee', and 'caller' cannot be accessed in this context.
```

![image](https://user-images.githubusercontent.com/22957874/75337623-42a18e00-5885-11ea-98ed-c413ec00f3e5.png)

## Steps to reproduce

1) Create a new RTCPeerConnection
2) Get the localStream from getUserMedia()
3) Execute RTCPeerConnection.addTrack(track) for every tack in the localStream

## Expected results

No errors

## Actual results

 TypeError: 'arguments', 'callee', and 'caller' cannot be accessed in this context.

fippo:
This one is going to get ugly. It looks like strict mode is forbidding that: 
https://whereswalden.com/2010/09/08/new-es5-strict-mode-support-now-with-poison-pills/
We *want* to not specify stream here since this would modify the function signature.

From what I can tell the right approach would be something like this:
```
const f = function(one, ...arguments) {}
```
which retains the mandatory argument but gives us access to the optional ones, without modifying the signature. Can you give that a try in the PR?
This probably happens in a number of places...

fippo:
I wonder if there is more here... the "legacy" createOffer stuff might be affected as well, did you test that?

dimitrisniras:
> I wonder if there is more here... the "legacy" createOffer stuff might be affected as well, did you test that?

I didn't have the chance to test this, but probably will have the same problem for every function in Safari is using arguments, or caller/callee, while using strict. So, I could open a new issue and a PR fixing the others as well.

[...]

PR #1024, opened 2020-02-26
Fix. "TypeError: 'arguments', 'callee', and 'caller' cannot be accessed in this context, when using addTrack() method"

dimitrisniras:
**Description**
PR to fix the issue "TypeError: 'arguments', 'callee', and 'caller' cannot be accessed in this context, when using addTrack() method" #1023

### DIFF
--- a/src/js/safari/safari_shim.js
+++ b/src/js/safari/safari_shim.js
@@ -39,7 +39,7 @@ export function shimLocalStreamsAPI(window) {
     };
 
     window.RTCPeerConnection.prototype.addTrack =
-      function addTrack(track, stream) {
+      function addTrack(track, ...stream) {
         if (stream) {
           if (!this._localStreams) {
             this._localStreams = [stream];
--- a/src/js/safari/safari_shim.js
+++ b/src/js/safari/safari_shim.js
@@ -21,20 +21,6 @@ export function shimLocalStreamsAPI(window) {
         return this._localStreams;
       };
   }
-  if (!('addTrack' in window.RTCPeerConnection.prototype)) {
-    const _addTrack = window.RTCPeerConnection.prototype.addTrack;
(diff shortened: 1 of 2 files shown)


test/unit/safari.js
```diff
@@ -31,7 +31,7 @@ describe('Safari shim', () => {
 
     it('shimStreamsAPI existence', () => {
       const prototype = window.RTCPeerConnection.prototype;
-      expect(prototype.addTrack.length).to.equal(1);
```
fippo: the point of not adding stream here is that its an optional argument in the IDL, check `RTCPeerConnection.prototype.addTrack.length` without adapter.
That should not be changed.

src/js/safari/safari_shim.js
```diff
@@ -39,14 +39,15 @@ export function shimLocalStreamsAPI(window) {
     };
 
     window.RTCPeerConnection.prototype.addTrack =
-      function addTrack(track) {
-        const stream = arguments[1];
+      function addTrack(track, ...stream) {
```
fippo: nit: `... streams` please. See [the spec](https://w3c.github.io/webrtc-pc/#rtcpeerconnection-interface-extensions)
  RTCRtpSender addTrack(MediaStreamTrack track, MediaStream... streams);

fippo:
Thank you! LGTM with a small nit. Can you squash or shall I The electron failure is unrelated, i'll take a look and/or merge regardless.

fippo (approved):

src/js/safari/safari_shim.js
dimitrisniras: Fixed that!

fippo:
published as 7.5.1 - sorry for being slow, had planned to fix more thing but... life happened.

PR #1036, opened 2020-05-20
util: fix wrapPeerConnection

### LINKED ISSUE #1034: Incorrect implementation of EventTarget

## Versions affected

- Chrome Version 81.0.4044.129 (Official Build) (64-bit) on Ubuntu 20.04.
- webrtc-adapter 7.6.0

## Description

According to [the spec](https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection), `RTCPeerConnection` inherits from `EventTarget`. `EventTarget`s implement the `addEventListener` method, which takes either a function [or an `EventListener`](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener) instance as the 2nd parameter.

Therefore we should be able to pass any object which implements a `handleEvent` method, and the RTCPeerConnection will invoke the method on any events.

## Steps to reproduce

```javascript
import 'webrtc-adapter'
class Test {
    constructor() {
        this.pc = new RTCPeerConnection({ iceServers: { urls: 'stun:stun.l.google.com:19302' } })
        this.pc.addEventListener('signalingstatechange', this)
        this.pc.createDataChannel('data')
    }
    close() {
        this.pc.removeEventListener('signalingstatechange', this)
        this.pc.close()
    }
    handleEvent(evt) {
        console.log('Received event', evt)
    }
}

new Test()
```

## Expected results

See some logs from the `handleEvent()` method.

## Actual results

With webrtc-adapter, I get this error: `utils.js:43 Uncaught TypeError: cb is not a function`. The stack trace only contains one line - `wrappedCallback @ utils.js:43` - pointing at https://github.com/webrtcHacks/adapter/blob/v7.6.0/src/js/utils.js#L43

Without webrtc-adapter, it works as expected.

fippo:
so it doesn't blindly assume that the handler is a function
Fixes #1034

Test Plan:
```
class Test {
    constructor() {
        this.pc = new RTCPeerConnection({ iceServers: [{ urls: 'stun:stun.l.google.com:19302' }] })
        this.pc.addEventListener('negotiationneeded', this)
        this.pc.createDataChannel('data')
    }
    close() {
        this.pc.removeEventListener('signalingstatechange', this)
        this.pc.close()
    }
    handleEvent(evt) {
        console.log('Received event', evt)
    }
}
const t = new Test(); t.pc.createOffer().then(offer => t.pc.setLocalDescription(offer))
```
shows 'Received event negotiationneeded'

### DIFF
--- a/src/js/utils.js
+++ b/src/js/utils.js
@@ -48,7 +48,10 @@ export function wrapPeerConnectionEvent(window, eventNameToWrap, wrapper) {
       }
     };
     this._eventMap = this._eventMap || {};
-    this._eventMap[cb] = wrappedCallback;
+    if (!this._eventMap[eventNameToWrap]) {
+      this._eventMap[eventNameToWrap] = [];
+    }
+    this._eventMap[eventNameToWrap].push([cb, wrappedCallback]);
     return nativeAddEventListener.apply(this, [nativeEventName,
       wrappedCallback]);
   };
@@ -56,11 +59,23 @@ export function wrapPeerConnectionEvent(window, eventNameToWrap, wrapper) {
   const nativeRemoveEventListener = proto.removeEventListener;
   proto.removeEventListener = function(nativeEventName, cb) {
     if (nativeEventName !== eventNameToWrap || !this._eventMap
-        || !this._eventMap[cb]) {
+        || !this._eventMap[eventNameToWrap]) {
       return nativeRemoveEventListener.apply(this, arguments);
     }
-    const unwrappedCb = this._eventMap[cb];
-    delete this._eventMap[cb];
+    const index = this._eventMap[eventNameToWrap].findIndex(([cb_, _]) => {
+      return cb_ === cb;
+    });
+    if (index === -1) {
+      return nativeRemoveEventListener.apply(this, arguments);
+    }
+    const unwrappedCb = this._eventMap[eventNameToWrap][index];
+    this._eventMap[eventNameToWrap].splice(index, 1);
+    if (this._eventMap[eventNameToWrap].length === 0) {
+      delete this._eventMap[eventNameToWrap];
+    }
+    if (Object.keys(this._eventMap).length === 0) {
+      delete this._eventMap;
+    }
     return nativeRemoveEventListener.apply(this, [nativeEventName,
       unwrappedCb]);
   };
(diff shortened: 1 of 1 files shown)


fippo:
I think this isn't going to work fully since _eventMap indexes by cb which will be the same object. Probably needs to index by [cb, eventname] but that was broken before.

fippo:
@jan-ivar can you review?

src/js/utils.js
```diff
@@ -40,23 +40,42 @@ export function wrapPeerConnectionEvent(window, eventNameToWrap, wrapper) {
...
+          cb(modifiedEvent);
+        }
       }
     };
     this._eventMap = this._eventMap || {};
-    this._eventMap[cb] = wrappedCallback;
+    if (!this._eventMap[eventNameToWrap]) {
+      this._eventMap[eventNameToWrap] = [];
+    }
+    this._eventMap[eventNameToWrap].push([cb, wrappedCallback]);
```
jan-ivar: Good riddance of the old code here, since objects used as keys in regular objects get [turned into strings](https://jsfiddle.net/jib1/ra4nuL32/), which means that the old code would not have distinguished between identical-looking functions that were nonetheless different objects. In contrast, a `Map` would let us use objects as indexes directly (with proper object comparison).

Good call to map things _per name_, to catch reusing the same handler on different topics. The array stuff looks accurate enough, but feels like a poor-man's re-implementation of `Map`, so I'd recommend just using `Map` instead—it's implemented in [all browsers](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map#Browser_compatibility) by now. E.g.
```js
    if (!this._eventMap[eventNameToWrap]) {
      this._eventMap[eventNameToWrap] = new Map();
    }
    this._eventMap[eventNameToWrap].set(cb, wrappedCallback);
```
In fact, I worry that the array version has subtly different semantics if add/remove is [called several times with the same listener](https://jsfiddle.net/jib1/nu1abc5e/).

jan-ivar (changes requested):

fippo:
updated!

jan-ivar (approved):
⭐⭐⭐⭐⭐

PR #1037, opened 2020-05-20
chrome: dont fudge with ONN in unified-plan mode

fippo:
where ONN works mostly as it should

Thanks @bradleythughes


bradleythughes (approved):
This fixes the problem I was having with the ONN event being swallowed. Thanks :)

PR #1042, opened 2020-05-25
safari: dont remove a=extmap-allow-mixed in current safari

fippo:
which does support parsing the extmap-allow-mixed attribute. See
  https://bugs.chromium.org/p/webrtc/issues/detail?id=9985

@youennf i picked a version I could test. If you know what webkit version actually includes [this commit](https://webrtc.googlesource.com/src.git/+/746dd0dbe60526ec7ce2d3dc8a43d6a66a0187d5) that would be great

### DIFF
--- a/src/js/common_shim.js
+++ b/src/js/common_shim.js
@@ -318,12 +318,15 @@ export function shimConnectionState(window) {
 }
 
 export function removeAllowExtmapMixed(window) {
-  /* remove a=extmap-allow-mixed for Chrome < M71 */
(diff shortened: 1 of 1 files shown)


youennf:
Safari 13.1 is probably the first version to get the fix.
If you are using user agent parsing "Version/13.XX" is probably the place to look for.

fippo:
if I recall correctly the UA sniffing was done on the webkit version in order to allow webkitgtk to work.

PR #1044, opened 2020-06-09
Shim sender.getParameters() in Firefox to match up w/addTransceiver shim

jan-ivar:
**Description**
Shim sender.getParameters() to include encodings added by addTransceiver(, {sendEncodings}) 

**Purpose**
Fix cases where JS uses setParameters before createOffer/Answer or uses new SLD/SRD wo/arguments.

PR #1046, opened 2020-06-14
bump version to 7.6.2

fippo:
safari: dont remove a=extmap-allow-mixed in current safari #1042
util: fix wrapPeerConnection #1036
Shim sender.getParameters() in Firefox to match up w/addTransceiver shim #1044

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "webrtc-adapter",
-  "version": "7.6.1",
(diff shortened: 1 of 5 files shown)


fippo:
The build failures are pretty odd, claiming install-firefox.sh isn't available and then only for some Firefox versions. There have been no changes to travis-multirunner recently so this looks bogus.

Issue #662, opened 2017-08-31
RTCPeerConnection.generateCertificate == undefined on Safari


benbro:
**Browsers and versions affected**
Safari 11.1 release 38 (technology preview)

**Description**
I'm using the following code to prefer ECDSA certificate on Chrome:
```javascript
if('generateCertificate' in RTCPeerConnection) {
  RTCPeerConnection.generateCertificate({
            name: 'ECDSA',
            namedCurve: 'P-256'
          }).then((cert) => {
            var pc = new RTCPeerConnection({certificates: [cert]});
          });
}
```
The above code gives in error RTCPeerConnection.generateCertificate is not a function in Safar because:
```javascript
'generateCertificate' in RTCPeerConnection == true
RTCPeerConnection.generateCertificate == undefined
```
I can test it with the following code but it doesn't make sense.
```javascript
if('generateCertificate' in RTCPeerConnection && RTCPeerConnection.generateCertificate) {
  //
}
```
generateCertificate shim on Safari:
https://github.com/webrtc/adapter/blob/master/src/js/safari/safari_shim.js#L234

**Steps to reproduce**

**Expected results**
'generateCertificate' in RTCPeerConnection == false

**Actual results**
'generateCertificate' in RTCPeerConnection == true

PR #1047, opened 2020-06-22
safari: unbreak generateCertificate

fippo:
the fix from #662 is not correct. Also add extra guards for consistency
with other parts of the code.

Test Plan: check that !!RTCPeerConnection.generateCertificate is true when including adapter

### DIFF
--- a/src/js/safari/safari_shim.js
+++ b/src/js/safari/safari_shim.js
@@ -233,6 +233,9 @@ export function shimConstraints(constraints) {
 }
 
 export function shimRTCIceServerUrls(window) {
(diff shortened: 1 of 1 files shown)


fippo:
see also #662

dagingaa (approved):

fippo:
can't reproduce Firefox build failures, something seems wrong with travis-multirunner.

I think that code is obsolete... it was removed for Chrome in 96c0f4f8de6e7f367f2fdc0f09c058bed1066f3d

fippo:
Note: apparently safari is not only not supporting `url:` so this can *not* be removed.

PR #1050, opened 2020-07-24
Fixes for running as an isomorphic web app

mck1117:
When running as an isomorphic web app, a window might not exist.  In fact, the window object might not even exist - so the first thing that happens needs to be a check for existence, before peeling off the navigator, since that might not even exist.

### DIFF
--- a/src/js/adapter_core.js
+++ b/src/js/adapter_core.js
@@ -11,5 +11,6 @@
 
 import {adapterFactory} from './adapter_factory.js';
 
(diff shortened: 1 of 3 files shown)


legastero (approved):

mck1117:
Should I bump the package version while I'm at it?

fippo:
nah, let me do that...

mck1117:
sounds good, thanks for the merge!

PR #1051, opened 2020-07-28
Bump version to 7.6.4

fippo:
including #1050 
-- should be safe to do a patch


fippo:
published -- thank you @mck1117!

PR #1052, opened 2020-08-09
Add default encodings[0] to getParameters() when missing in Firefox.

jan-ivar:
**Description**
Return at minimum an empty encoding from getParameters() when missing in Firefox.

**Purpose**
Improve getParameters() firefox shim just a tad more, so we can write JS with fewer hacks.

### DIFF
--- a/src/js/firefox/firefox_shim.js
+++ b/src/js/firefox/firefox_shim.js
@@ -263,7 +263,7 @@ export function shimGetParameters(window) {
   if (origGetParameters) {
     window.RTCRtpSender.prototype.getParameters =
       function getParameters() {
-        var params = origGetParameters.apply(this, arguments);
+        const params = origGetParameters.apply(this, arguments);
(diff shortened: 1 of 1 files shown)


fippo:
can you update the `var` to be a `const`?

benbro:
Why not fix it in Firefox?

jan-ivar:
> Why not fix it in Firefox?

For sure, but adding workarounds in adapter is not mutually exclusive with that, and has a quicker turnaround to [help users do basic things web-compatibly](https://stackoverflow.com/a/63331550/918910) in the meantime. The work to fix Firefox is tracked in [bug 1401592](https://bugzilla.mozilla.org/show_bug.cgi?id=1401592).

fippo:
looks like this fails in test/e2e/simulcast.js?

fippo (approved):

fippo:
\o/

PR #1056, opened 2020-08-19
chore: bump test dependencies

fippo:


fippo:
looks like the stability reporter is broken after the karma update... it is not filling assretionErrors anymore.

PR #1058, opened 2020-12-22
add github action for lint + unit tests

fippo:


fippo:
travis failure is consistent with recent build failures so unrelated.

PR #1060, opened 2020-12-23
[DONT MERGE] try travis-multirunner action

fippo:


fippo:
need to find time to actually do this...

PR #1061, opened 2021-01-04
add matrix build action

fippo:
and remove travis-ci. Same as #1059 but hopefully without travis hanging around.
Some TODOs:
* port nightly run (or stop doing it, it has not triggered in 2020)
* cache node_modules across jobs (due to the lack of package-lock.json probably from GITHUB_SHA and GITHUB_REF, see here)


fippo:
ok, no waiting on travis...

PR #1066, opened 2021-02-13
unify addIceCandidate(null) and addIceCandidate({candidate: "", ...})

### LINKED ISSUE #863: ignore addIceCandidate with empty candidate string

fippo:
fixes #863


fippo:
@jan-ivar can you review? I want to forget about this topic ONCE AND FOR ALL ;-)

OlivierJM (approved):
This looks good

jan-ivar (approved):

PR #1071, opened 2021-03-10
dont attempt to overwrite RTCSessionDescription.sdp

fippo:
follow-up on #1042. Shouldn't be necessary in new Safari versions but see [here](https://bugs.chromium.org/p/chromium/issues/detail?id=1186456&)


dagingaa (approved):
LGTM!

PR #1073, opened 2021-03-18
firefox: fix detection on insecure contexts

fippo:
@jan-ivar: on insecure contexts mozGetUserMedia isn't defined but mozRTCPeerConnection is :-p

### DIFF
--- a/src/js/utils.js
+++ b/src/js/utils.js
@@ -160,7 +160,7 @@ export function detectBrowser(window) {
 
   const {navigator} = window;
 
-  if (navigator.mozGetUserMedia) { // Firefox.
+  if (navigator.mozGetUserMedia || window.mozRTCPeerConnection) { // Firefox.
     result.browser = 'firefox';
     result.version = extractVersion(navigator.userAgent,
         /Firefox\/(\d+)\./, 1);


jan-ivar:
Hmm, what's this needed for? Any reason to not use UA string instead? 

We'd like to get rid of these moz ones eventually ([bug 1531812](https://bugzilla.mozilla.org/show_bug.cgi?id=1531812)).

fippo:
I thought the general agreement was that UA sniffing is even worse than using prefixes to do feature detection?
Mostly for completeness sake, if Firefox can be detected in such cases, do detect it.

jan-ivar:
> I thought the general agreement was that UA sniffing is even worse than using prefixes to do feature detection?

Needing to detect browsers at all is bad, regardless of how. This function called `detectBrowser` is a bad pattern that competes with (and thus defeats) actual "feature detection".

Is this an internal function to adapter, or is it exposed to apps? If we can't rid adapter of it, I think I'd prefer a straightforward parsing of the UA string. That way, I can at least see what features a site is enabling or disabling, by changing it in config. 

Vendor-prefixed or not, relying on non-standard APIs prevents us from ever removing them. So I'd rather we not merge this. As I mention in [Bug 1531812 (comment 3)](https://bugzilla.mozilla.org/show_bug.cgi?id=1531812#c3) I think we have a shot at removing `mozRTCPeerConnection`, more so than `mozGetUserMedia`.

fippo:
I remember mozGetUserMedia being considered last-man-standing :-)
See also #764. I wish adapter was obsolete by now but it isn't.

fippo:
not worth it probably

PR #1085, opened 2021-07-12
Shim parameterless SLD in Safari

### LINKED ISSUE #1084: RTCPeerConnection.setLocalDescription() without arguments fails on Safari 13.1 (macOS Catalina 10.15.4) with TypeError: Argument 1 ('description') to RTCPeerConnection.setLocalDescription must be an instance of RTCSessionDescription
# Please read first!

- [x] This issue only happens when adapter.js is used (**kinda, see below**)

## Versions affected

**Browser name including version (e.g. Chrome 64.0.3282.119)**
Safari Version 13.1 (15609.1.20.111.8)
Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_4) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/13.1 Safari/605.1.15

**adapter.js (e.g. 6.1.0)**
8.0.0

## Description
Calling `RTCPeerConnection.setLocalDescription()` without any arguments in Safari 13.1 (macOS Catalina 10.15.4) fails with:
> TypeError: Argument 1 ('description') to RTCPeerConnection.setLocalDescription must be an instance of RTCSessionDescription

## Steps to reproduce
https://jsfiddle.net/60ne23pb/1/
```javascript
const ctx = new (window.AudioContext || webkitAudioContext)();
const gain = ctx.createGain();
const {stream} = gain.connect(ctx.createMediaStreamDestination());
const track = stream.getAudioTracks()[0];

const pc = new RTCPeerConnection({});
pc.addEventListener('negotiationneeded', async () => {
  await pc.setLocalDescription();
});
pc.addTrack(track, stream);
```

## Expected results
It should behave according to [the spec](https://www.w3.org/TR/2021/REC-webrtc-20210126/#dom-peerconnection-setlocaldescription):

> Passing in a description is optional. If left out, then setLocalDescription will implicitly create an offer or create an answer, as needed.

## Actual results
Throws the following exception:
> TypeError: Argument 1 ('description') to RTCPeerConnection.setLocalDescription must be an instance of RTCSessionDescription

Removing adapter.js changes the behavior slightly; it instead throws:
> TypeError: Not enough arguments

jan-ivar:
Fixes https://github.com/webrtcHacks/adapter/issues/1084. Disclaimer: I haven't built it, as I don't have my npm setup for this atm.

### DIFF
--- a/src/js/safari/safari_shim.js
+++ b/src/js/safari/safari_shim.js
@@ -387,7 +387,7 @@ export function shimParameterlessSetLocalDescription(window, browserDetails) {
             break;
         }
       }
-      if (desc.sdp) {
+      if (desc.sdp || (desc.type != "offer" && desc.type != "answer")) {
         return nativeSetLocalDescription.apply(this, [desc]);
       }
       const func = desc.type == "offer" ? this.createOffer : this.createAnswer;
--- a/src/js/safari/safari_shim.js
+++ b/src/js/safari/safari_shim.js
@@ -377,20 +377,20 @@ export function shimParameterlessSetLocalDescription(window, browserDetails) {
       desc = {type: desc.type, sdp: desc.sdp};
       if (!desc.type) {
         switch (this.signalingState) {
-          case "stable":
-          case "have-local-offer":
-          case "have-remote-pranswer":
-            desc.type = "offer";
+          case 'stable':
+          case 'have-local-offer':
+          case 'have-remote-pranswer':
+            desc.type = 'offer';
             break;
           default:
-            desc.type = "answer";
+            desc.type = 'answer';
             break;
         }
       }
-      if (desc.sdp || (desc.type != "offer" && desc.type != "answer")) {
+      if (desc.sdp || (desc.type != 'offer' && desc.type != 'answer')) {
         return nativeSetLocalDescription.apply(this, [desc]);
       }
-      const func = desc.type == "offer" ? this.createOffer : this.createAnswer;
+      const func = desc.type == 'offer' ? this.createOffer : this.createAnswer;
       return func.apply(this)
         .then(desc => nativeSetLocalDescription.apply(this, [desc]));
     };
--- a/src/js/safari/safari_shim.js
+++ b/src/js/safari/safari_shim.js
@@ -364,7 +364,7 @@ export function shimParameterlessSetLocalDescription(window, browserDetails) {
   window.RTCPeerConnection.prototype.setLocalDescription =
     function setLocalDescription() {
       let desc = arguments[0] || {};
-      if (typeof desc != "object" || (desc.type && desc.sdp)) {
+      if (typeof desc !== "object" || (desc.type && desc.sdp)) {
         return nativeSetLocalDescription.apply(this, arguments);
       }
       // The remaining steps should technically happen when SLD comes off the
--- a/src/js/safari/safari_shim.js
+++ b/src/js/safari/safari_shim.js
@@ -364,7 +364,7 @@ export function shimParameterlessSetLocalDescription(window, browserDetails) {
   window.RTCPeerConnection.prototype.setLocalDescription =
     function setLocalDescription() {
       let desc = arguments[0] || {};
-      if (typeof desc !== "object" || (desc.type && desc.sdp)) {
+      if (typeof desc !== 'object' || (desc.type && desc.sdp)) {
         return nativeSetLocalDescription.apply(this, arguments);
       }
       // The remaining steps should technically happen when SLD comes off the
@@ -387,11 +387,11 @@ export function shimParameterlessSetLocalDescription(window, browserDetails) {
             break;
         }
       }
-      if (desc.sdp || (desc.type != 'offer' && desc.type != 'answer')) {
+      if (desc.sdp || (desc.type !== 'offer' && desc.type !== 'answer')) {
         return nativeSetLocalDescription.apply(this, [desc]);
       }
-      const func = desc.type == 'offer' ? this.createOffer : this.createAnswer;
+      const func = desc.type === 'offer' ? this.createOffer : this.createAnswer;
       return func.apply(this)
-        .then(desc => nativeSetLocalDescription.apply(this, [desc]));
+        .then(description => nativeSetLocalDescription.apply(this, [description]));
     };
 }
--- a/src/js/safari/safari_shim.js
+++ b/src/js/safari/safari_shim.js
@@ -350,3 +350,48 @@ export function shimAudioContext(window) {
   }
   window.AudioContext = window.webkitAudioContext;
 }
+
+// Note: Make sure to call this ahead of shimCallbacksAPI.
+export function shimParameterlessSetLocalDescription(window, browserDetails) {
+  if (!(window.RTCPeerConnection && window.RTCPeerConnection.prototype)) {
+    return;
+  }
+  const nativeSetLocalDescription =
+      window.RTCPeerConnection.prototype.setLocalDescription;
+  if (!nativeSetLocalDescription || nativeSetLocalDescription.length === 0) {
+    return;
+  }
+  window.RTCPeerConnection.prototype.setLocalDescription =
+    function setLocalDescription() {
+      let desc = arguments[0] || {};
+      if (typeof desc != "object" || (desc.type && desc.sdp)) {
+        return nativeSetLocalDescription.apply(this, arguments);
+      }
+      // The remaining steps should technically happen when SLD comes off the
+      // RTCPeerConnection's operations chain (not ahead of going on it), but
+      // this is too difficult to shim. Instead, this shim only covers the
+      // common case where the operations chain is empty. This is imperfect, but
+      // should cover many cases. Rationale: Even if we can't reduce the glare
+      // window to zero on imperfect implementations, there's value in tapping
+      // into the perfect negotiation pattern that several browsers support.
+      desc = {type: desc.type, sdp: desc.sdp};
+      if (!desc.type) {
+        switch (this.signalingState) {
+          case "stable":
+          case "have-local-offer":
+          case "have-remote-pranswer":
+            desc.type = "offer";
+            break;
+          default:
+            desc.type = "answer";
+            break;
+        }
+      }
+      if (desc.sdp) {
+        return nativeSetLocalDescription.apply(this, [desc]);
+      }
+      const func = desc.type == "offer" ? this.createOffer : this.createAnswer;
+      return func.apply(this)
+        .then(desc => nativeSetLocalDescription.apply(this, [desc]);
+    };
+}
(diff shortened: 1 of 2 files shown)


src/js/safari/safari_shim.js
```diff
@@ -350,3 +350,48 @@ export function shimAudioContext(window) {
...
+  if (!(window.RTCPeerConnection && window.RTCPeerConnection.prototype)) {
+    return;
+  }
+  const nativeSetLocalDescription =
+      window.RTCPeerConnection.prototype.setLocalDescription;
+  if (!nativeSetLocalDescription || nativeSetLocalDescription.length === 0) {
+    return;
+  }
+  window.RTCPeerConnection.prototype.setLocalDescription =
+    function setLocalDescription() {
```
rhansen, line 365: Is it necessary to support the case where `desc` is non-null but `desc.type` is falsy? If not, I think the function can be simplified:

```javascript
  window.RTCPeerConnection.prototype.setLocalDescription = async function (...args) {
    if (args.length === 0) {
      // The description should technically be created when the SLD comes off
      // the RTCPeerConnection's operations chain (not ahead of going on it),
      // but this is too difficult to shim. Instead, this shim only covers the
      // common case where the operations chain is empty. This is imperfect, but
      // should cover many cases. Rationale: Even if we can't reduce the glare
      // window to zero on imperfect implementations, there's value in tapping
      // into the perfect negotiation pattern that several browsers support.
      const createOffer =
          ['stable', 'have-local-offer', 'have-remote-pranswer'].includes(this.signalingState);
      args.push(createOffer ? await this.createOffer() : await this.createAnswer());
    }
    return await nativeSetLocalDescription.call(this, ...args);
  };
```

src/js/safari/safari_shim.js
```diff
@@ -350,3 +350,48 @@ export function shimAudioContext(window) {
...
+// Note: Make sure to call this ahead of shimCallbacksAPI.
+export function shimParameterlessSetLocalDescription(window, browserDetails) {
+  if (!(window.RTCPeerConnection && window.RTCPeerConnection.prototype)) {
+    return;
+  }
+  const nativeSetLocalDescription =
+      window.RTCPeerConnection.prototype.setLocalDescription;
+  if (!nativeSetLocalDescription || nativeSetLocalDescription.length === 0) {
+    return;
+  }
```
rhansen, line 363: Would the following be a more robust way to feature test? I think this will work even if `shimParameterlessSetLocalDescription()` is called after `shimCallbacksAPI()`. Unfortunately, it is asynchronous, which might be a dealbreaker.

```javascript
  const nativeSetLocalDescription =
      window.RTCPeerConnection &&
      window.RTCPeerConnection.prototype &&
      window.RTCPeerConnection.prototype.setLocalDescription;
  if (!nativeSetLocalDescription) return;
  try {
    await new window.RTCPeerConnection().setLocalDescription();
    return;
  } catch (err) {}
```

src/js/safari/safari_shim.js
```diff
@@ -350,3 +350,48 @@ export function shimAudioContext(window) {
...
+            desc.type = "answer";
+            break;
+        }
+      }
+      if (desc.sdp) {
+        return nativeSetLocalDescription.apply(this, [desc]);
+      }
+      const func = desc.type == "offer" ? this.createOffer : this.createAnswer;
+      return func.apply(this)
+        .then(desc => nativeSetLocalDescription.apply(this, [desc]);
```
rhansen: Missing closing paren:
```suggestion
        .then(desc => nativeSetLocalDescription.apply(this, [desc]));
```

rhansen (commented):
Thank you for this!

rhansen:
After adding the missing parenthesis it works as expected.

src/js/safari/safari_shim.js
fippo: i'm not a fan of creating peerconnections and disposing them for the sake of feature deteection.
Observing the order or pulling the "need to" check before doing any modifications should be fine.

src/js/safari/safari_shim.js
```diff
@@ -350,3 +350,48 @@ export function shimAudioContext(window) {
...
+  }
+  const nativeSetLocalDescription =
+      window.RTCPeerConnection.prototype.setLocalDescription;
+  if (!nativeSetLocalDescription || nativeSetLocalDescription.length === 0) {
+    return;
+  }
+  window.RTCPeerConnection.prototype.setLocalDescription =
+    function setLocalDescription() {
+      let desc = arguments[0] || {};
+      if (typeof desc != "object" || (desc.type && desc.sdp)) {
```
fippo, line 367: !==

fippo:
```suggestion
      if (typeof desc !== "object" || (desc.type && desc.sdp)) {
```
needs extra check for rollback?

src/js/safari/safari_shim.js
jan-ivar: Polyfills run ahead of WebIDL's input validation or normalization. WebIDL says all these must work equivalently:
```js
await pc.setLocalDescription();
await pc.setLocalDescription(null);
await pc.setLocalDescription({});
await pc.setLocalDescription({foo: 3});
```
Also, the way the [spec](https://w3c.github.io/webrtc-pc/#dom-peerconnection-setlocaldescription) is written (and [implemented in firefox](https://searchfox.org/mozilla-central/rev/4163c8f09b03871e053a61f39de020fa09e02ea9/dom/media/PeerConnection.jsm#1071,1083)), the dynamic defaults for `type` and `sdp` are provided independently, which means the following combinations should also work (or fail predictably depending on state):
```js
await pc.setLocalDescription({type: "offer"});
await pc.setLocalDescription({type: "answer"});
await pc.setLocalDescription({sdp: sdp});
```

As a backwards-compatible shim, Adapter also avoids `async`/`await`, since they don't parse on older browsers.

src/js/safari/safari_shim.js
jan-ivar: Good catch on `"rollback"`!

I'll yield to the prevailing pattern in this module on !== and ===

src/js/safari/safari_shim.js
```diff
@@ -350,3 +350,48 @@ export function shimAudioContext(window) {
...
+          case "have-local-offer":
+          case "have-remote-pranswer":
+            desc.type = "offer";
+            break;
+          default:
+            desc.type = "answer";
+            break;
+        }
+      }
+      if (desc.sdp) {
```
jan-ivar:
```suggestion
      if (desc.sdp || (desc.type != "offer" && desc.type != "answer")) {
```

src/js/safari/safari_shim.js
```diff
@@ -350,3 +350,48 @@ export function shimAudioContext(window) {
...
+            break;
+          default:
+            desc.type = "answer";
+            break;
+        }
+      }
+      if (desc.sdp || (desc.type != "offer" && desc.type != "answer")) {
+        return nativeSetLocalDescription.apply(this, [desc]);
+      }
+      const func = desc.type == "offer" ? this.createOffer : this.createAnswer;
```
jan-ivar:
```suggestion
          case 'stable':
          case 'have-local-offer':
          case 'have-remote-pranswer':
            desc.type = 'offer';
            break;
          default:
            desc.type = 'answer';
            break;
        }
      }
      if (desc.sdp || (desc.type != 'offer' && desc.type != 'answer')) {
        return nativeSetLocalDescription.apply(this, [desc]);
      }
      const func = desc.type == 'offer' ? this.createOffer : this.createAnswer;
```

src/js/safari/safari_shim.js
```diff
@@ -350,3 +350,48 @@ export function shimAudioContext(window) {
...
+  }
+  const nativeSetLocalDescription =
+      window.RTCPeerConnection.prototype.setLocalDescription;
+  if (!nativeSetLocalDescription || nativeSetLocalDescription.length === 0) {
+    return;
+  }
+  window.RTCPeerConnection.prototype.setLocalDescription =
+    function setLocalDescription() {
+      let desc = arguments[0] || {};
+      if (typeof desc !== "object" || (desc.type && desc.sdp)) {
```
jan-ivar:
```suggestion
      if (typeof desc !== 'object' || (desc.type && desc.sdp)) {
```

src/js/safari/safari_shim.js
```diff
@@ -350,3 +350,48 @@ export function shimAudioContext(window) {
...
+          case 'have-local-offer':
+          case 'have-remote-pranswer':
+            desc.type = 'offer';
+            break;
+          default:
+            desc.type = 'answer';
+            break;
+        }
+      }
+      if (desc.sdp || (desc.type != 'offer' && desc.type != 'answer')) {
```
jan-ivar:
```suggestion
      if (desc.sdp || (desc.type !== 'offer' && desc.type !== 'answer')) {
```

src/js/safari/safari_shim.js
```diff
@@ -350,3 +350,48 @@ export function shimAudioContext(window) {
...
+        return nativeSetLocalDescription.apply(this, [desc]);
+      }
+      const func = desc.type == 'offer' ? this.createOffer : this.createAnswer;
```
jan-ivar:
```suggestion
      const func = desc.type === 'offer' ? this.createOffer : this.createAnswer;
```

src/js/safari/safari_shim.js
```diff
@@ -350,3 +350,48 @@ export function shimAudioContext(window) {
...
+      return func.apply(this)
+        .then(desc => nativeSetLocalDescription.apply(this, [desc]));
```
jan-ivar:
```suggestion
        .then(description => nativeSetLocalDescription.apply(this, [description]));
```

src/js/safari/safari_shim.js
```diff
@@ -350,3 +350,48 @@ export function shimAudioContext(window) {
...
+            desc.type = 'answer';
+            break;
+        }
+      }
+      if (desc.sdp || (desc.type !== 'offer' && desc.type !== 'answer')) {
+        return nativeSetLocalDescription.apply(this, [desc]);
+      }
+      const func = desc.type === 'offer' ? this.createOffer : this.createAnswer;
+      return func.apply(this)
+        .then(description => nativeSetLocalDescription.apply(this, [description]));
```
jan-ivar:
```suggestion
        .then(d => nativeSetLocalDescription.apply(this, [d]));
```

fippo (approved):

rhansen:
I just tested the latest version and it still works well. Thank you!

rhansen:
I encountered an old Chrome that also lacks support for parameterless SLD. Can this be added to Chrome as well? Moving `shimParameterlessSetLocalDescription()` to `common_shim.js` and calling it for Chrome seems to work just fine in my limited testing.

fippo:
so Firefox added this in FF75 (https://bugzilla.mozilla.org/show_bug.cgi?id=1568292)
@henbos can you check what Chrome version this shipped in?

Given that is using proper feature detection moving it to the common shim seems like a good idea. This might be a bit trickier due to more things manipulating setLocalDescription though.

fippo:
i'll do the chrome/firefox bits (which is somewhat ironic since jib knows firefox better buuuuut...)

fippo:
*sigh*... wrong button :-(
squashed and rewrote history.

PR #1094, opened 2021-12-15
safari: fix transceiver direction in createOffer

ybandou:
When there is no transceiver, it is created with the default "sendrecv" direction.
From the WebRTC spec, it must be "recvonly".
https://w3c.github.io/webrtc-pc/#legacy-configuration-extensions

**Description**

**Purpose**


fippo (approved):

fippo:
published as 8.1.1, thank you!

PR #1096, opened 2022-02-02
chrome: add naive selectAudioOutput shim

fippo:
based on enumerateDevices. This allows apps to somewhat use
getUserMedia already to use selectAudioOutput (but will not show any prompt).
Note that this does not (yet) modify enumerateDevices output.


fippo:
@jan-ivar useful or harmful? The tests also show I didn't think about how it should behave in Firefox...

hthetiot:
naive shim is more like a like sham....
And it will fail the feature detection with false positive and resulting in the behavior to be expected by the user not working and exposed to user because false positive sham.

Please dont merge this.

PR #1101, opened 2022-05-27
Shims/fixes for `RTCDataChannel.binaryType`

Liamolucko:
**Description**
- Shims `RTCDataChannel.binaryType: 'blob'` in Chrome (which doesn't support it)
- Makes 'blob' the default value for `RTCDataChannel.binaryType` in Chrome and Safari, as the spec states it should be.

**Purpose**
Make `RTCDataChannel.binaryType` work consistently across browsers.

### DIFF
--- a/src/js/adapter_factory.js
+++ b/src/js/adapter_factory.js
@@ -70,6 +70,7 @@ export function adapterFactory({window} = {}, options = {
       commonShim.shimMaxMessageSize(window, browserDetails);
       commonShim.shimSendThrowTypeError(window, browserDetails);
       commonShim.removeExtmapAllowMixed(window, browserDetails);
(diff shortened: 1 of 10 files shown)


fippo:
the code looks nice but how will it deal with attempts to send a large blob (e.g. a whole file) over the datachannel?

(in general the resistance of the webrtc working group to adapt to the sad reality of chrome not implementing blob is surprising... the default should have been changed years ago)

Liamolucko:
> the code looks nice but how will it deal with attempts to send a large blob (e.g. a whole file) over the datachannel?

I don't think that'd really be possible, while still sending it as a single message.

My understanding is that WebRTC currently has quite a small maximum message size anyway though.

Issue #1017, opened 2020-01-14
navigator.userAgent deprecation


fippo:
see https://groups.google.com/a/chromium.org/forum/#!msg/blink-dev/-2JIRNMWJ7s/yHe4tQNLCgAJ

fippo:
`git grep browserDetails src/` isn't heavily used internally.

getUserAgent() returns a promise and is not sync which is a problem. We'd only use the major version anyway.

I'm not sure if there is much action here, adapter can continue to return the values from the UA but it needs to be documented that this isn't to be trusted. If browsers add a reliable way to detect the major version that would actually be great

fippo:
Done in 9.0.0

PR #1103, opened 2022-06-13
prefer navigator.userAgentData for browser detection

### LINKED ISSUE #764: Chrome will be identified as Safari if webkitGetUserMedia is removed

fippo:
fixes #764,#1017

### DIFF
--- a/src/js/utils.js
+++ b/src/js/utils.js
@@ -160,6 +160,16 @@ export function detectBrowser(window) {
 
   const {navigator} = window;
 
+  // Prefer navigator.userAgentData.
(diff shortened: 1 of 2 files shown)


fippo:
see also https://web.dev/migrate-to-ua-ch/

PR #1111, opened 2022-09-13
all: add getDataChannels polyfill

fippo:
implementing
  https://github.com/w3c/webrtc-extensions/issues/110

@orphis something along these lines?


Orphis (approved):
While not familiar with the implementation details, it looks like it's doing what I intend to specify.

fippo:
i'll eagerly await for your WPT then ;-)

PR #1117, opened 2022-11-28
common: replace substr with substring

fippo:
alternative to #1098


fippo:
travis-multirunner failure too

PR #1119, opened 2023-02-17
eslint and travis-multirunner updates, whitespace changes

fippo:
and a change in the Safari url->urls transition.
No functional changes intended.

Best viewed ignoring whitespace, e.g. using gits `-b`


fippo:
(note to self: when doing a patch release because why not bump the sdp dependency)

saghul (approved):

PR #1123, opened 2023-02-22
Firefox: do not attempt to modify read-only property in ice candidate shim

### LINKED ISSUE #1122: RTCIceCandidate constructor shim throwing error if ufrag given for Firefox beta (111)
# Please read first!

- [ x] I have provided steps to reproduce (e.g. a link to a [jsfiddle](https://jsfiddle.net/))
- [ x] I have provided browser name, version and adapter.js version
- [ x] This issue only happens when adapter.js is used

## Versions affected

Firefox (Beta) version 111.0b3 (64-bit)
webrtc-adapter@8.2.1

## Description
RTCIceCandidate() constructor throws the below error when ufrag supplied on the candidate line.

TypeError: setting getter-only property "usernameFragment"
    RTCIceCandidate common_shim.js:35
    <anonymous> debugger eval code:1

Error is not seen in current stable Firefox (v110).  No error is thrown if ufrag is not given.

## Steps to reproduce

Try creating RTCIceCandidate using the following candidateInfo: 

```
new RTCIceCandidate({ candidate: 'candidate:4234997325 1 udp 2043278322 192.168.0.56 44323 typ host generation 0 ufrag iTuJ network-id 1', sdpMid: 0 })
```

## Expected results
Newly created RTCIceCandidate object returned.

## Actual results
The following error is thrown:
TypeError: setting getter-only property "usernameFragment"
    RTCIceCandidate common_shim.js:35
    <anonymous> debugger eval code:1

fippo:
fixes #1122

### DIFF
--- a/src/js/common_shim.js
+++ b/src/js/common_shim.js
@@ -32,20 +32,23 @@ export function shimRTCIceCandidate(window) {
       // Augment the native candidate with the parsed fields.
       const nativeCandidate = new NativeRTCIceCandidate(args);
       const parsedCandidate = SDPUtils.parseCandidate(args.candidate);
-      const augmentedCandidate = Object.assign({}, nativeCandidate,
-        parsedCandidate);
-      augmentedCandidate.prototype = nativeCandidate.prototype;
+      for (const key in parsedCandidate) {
+        if (!(key in nativeCandidate)) {
+          Object.defineProperty(nativeCandidate, key,
+            {value: parsedCandidate[key]});
+        }
+      }
 
-      // Add a serializer that does not serialize the extra attributes.
-      augmentedCandidate.toJSON = function toJSON() {
+      // Override serializer to not serialize the extra attributes.
+      nativeCandidate.toJSON = function toJSON() {
         return {
-          candidate: augmentedCandidate.candidate,
-          sdpMid: augmentedCandidate.sdpMid,
-          sdpMLineIndex: augmentedCandidate.sdpMLineIndex,
-          usernameFragment: augmentedCandidate.usernameFragment,
+          candidate: nativeCandidate.candidate,
+          sdpMid: nativeCandidate.sdpMid,
+          sdpMLineIndex: nativeCandidate.sdpMLineIndex,
+          usernameFragment: nativeCandidate.usernameFragment,
         };
       };
-      return augmentedCandidate;
+      return nativeCandidate;
     }
     return new NativeRTCIceCandidate(args);
   };
--- a/src/js/common_shim.js
+++ b/src/js/common_shim.js
@@ -32,8 +32,9 @@ export function shimRTCIceCandidate(window) {
       // Augment the native candidate with the parsed fields.
       const nativeCandidate = new NativeRTCIceCandidate(args);
       const parsedCandidate = SDPUtils.parseCandidate(args.candidate);
-      const augmentedCandidate = Object.assign(nativeCandidate,
+      const augmentedCandidate = Object.assign({}, nativeCandidate,
         parsedCandidate);
+      augmentedCandidate.prototype = nativeCandidate.prototype;
 
       // Add a serializer that does not serialize the extra attributes.
       augmentedCandidate.toJSON = function toJSON() {


src/js/common_shim.js
```diff
@@ -32,8 +32,9 @@ export function shimRTCIceCandidate(window) {
       // Augment the native candidate with the parsed fields.
       const nativeCandidate = new NativeRTCIceCandidate(args);
       const parsedCandidate = SDPUtils.parseCandidate(args.candidate);
-      const augmentedCandidate = Object.assign(nativeCandidate,
+      const augmentedCandidate = Object.assign({}, nativeCandidate,
         parsedCandidate);
+      augmentedCandidate.prototype = nativeCandidate.prototype;
```
jan-ivar: `nativeCandidate.prototype` is `undefined` here so this won't work.

I'd prefer not to mess with prototypes and instead modify and return the real platform object. This [should work](https://jsfiddle.net/jib1/4rhLe5y3/41/):
```js
    const nativeCandidate = new NativeRTCIceCandidate(args);
    const parsedCandidate = SDPUtils.parseCandidate(args.candidate);
    for (const key in parsedCandidate) {
      nativeCandidate[key] = parsedCandidate[key];
    }
```

jan-ivar: The above should provide parity with existing code. If we want to match spec, we'd [add read-only properties](https://jsfiddle.net/jib1/4rhLe5y3/38):
```js
    const nativeCandidate = new RTCIceCandidate(options);
    const parsedCandidate = SDPUtils.parseCandidate(options.candidate);
    for (const key in parsedCandidate) {
      if (!(key in nativeCandidate)) {
        Object.defineProperty(nativeCandidate, key, {value: parsedCandidate[key]});
      }
    }
```

jan-ivar (changes requested):

src/js/common_shim.js
fippo: That is better indeed! I wonder if the native Firefox toJSON would do the right thing out of the box but technically this code path can still be hit very *very* old Chrome

fippo:
Sadly the patch doesn't apply clean to v7 or v6 (that one is still ES5) but it is easy enough to backport.

src/js/common_shim.js
```diff
@@ -32,19 +32,23 @@ export function shimRTCIceCandidate(window) {
...
+      for (const key in parsedCandidate) {
+        if (!(key in nativeCandidate)) {
+          Object.defineProperty(nativeCandidate, key,
+            {value: parsedCandidate[key]});
+        }
+      }
 
-      // Add a serializer that does not serialize the extra attributes.
-      augmentedCandidate.toJSON = function toJSON() {
+      // Override serializer to not serialize the extra attributes.
```
jan-ivar, line 42: Hmm, this should be unnecessary, since [enumerable](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty#enumerable) defaults to `false`. Did it not work in ES5? If the default changed, does the following help?
```js
Object.defineProperty(nativeCandidate, key, {
  value: parsedCandidate[key],
  enumerable: false
});
```

jan-ivar: If anything, [my pilot error](https://github.com/webrtcHacks/adapter/issues/1122#issuecomment-1440795147) was caused by the custom toJSON on the native object persisting, causing me to mistakenly think the properties hadn't been copied, because I was using JSON.stringify to dump things to console in my first fiddle attempts, and even new enumerable properties weren't showing up for me.

jan-ivar: To clarify, I believe even with `enumerable: true` I would not expect new properties to appear in JSON, because of the native [toJSON for RTCIceCandidate](https://w3c.github.io/webrtc-pc/#dom-rtcicecandidate-tojson) is already custom and limited.

fippo: True but this code still has to deal with the theoretical possibility of Chromes native toJSON predating the right thing

jan-ivar: Oh I missed that this was fixing up existing code. LGTM!

jan-ivar (approved):

Issue #1065, opened 2021-01-28
look at version distribution


fippo:
https://api.npmjs.org/versions/webrtc-adapter/last-week shows the download distribution. Looks like a considerable amount of 6.x usage

fippo:
grabbing that data into `s` and then doing
```
majorVersions = {};
Object.keys(s.downloads).forEach(version => {
  const majorVersion = parseInt(version, 10);
  if (!majorVersions[majorVersion]) majorVersions[majorVersion] = 0;
  majorVersions[majorVersion] += s.downloads[version];
});
```
shows
```
Object { 0: 1, 1: 1615, 2: 2512, 3: 5898, 4: 1787, 5: 1046, 6: 25177, 7: 98195 }
```
So 20% of dependents are still stuck on version 6.

fippo:
```
Object.keys(s.downloads).forEach(version => {
  const majorVersion = parseInt(version, 10);
  if (majorVersion < 6) return;
  console.log(version, s.downloads[version]);
})
```
at least mostly using the latest versions of 6 + 7

fippo:
```
Object { 0: 14, 1: 3260, 2: 1958, 3: 4323, 4: 641, 5: 261, 6: 16590, 7: 119796, 8: 33792 }
```
Quite surprising 7.x is so popular given the big removal for 8.x was legacy Edge

fippo:
```
Object { 0: 18, 1: 5263, 2: 1952, 3: 7406, 4: 610, 5: 566, 6: 49052, 7: 146731, 8: 51352 }
```
https://npmtrends.com/webrtc-adapter shows quite a large bump recently

fippo:
```Object {0: 10, 1: 7247, 2: 1491, 3: 4003, 4: 767, 5: 642, 6: 32823, 7: 190426, 8: 63210}```

fippo:
`{0: 35, 1: 4334, 2: 891, 3: 2312, 4: 993, 5: 285, 6: 8877, 7: 112479, 8: 66910}`
At least 6.x dropped in usage

fippo:
`{0: 3, 1: 30803, 2: 1342, 3: 2960, 4: 559, 5: 260, 6: 15729, 7: 133953, 8: 168522, 9: 7048}`
v8 is slightly ahead now

fippo:
[...]

PR #1127, opened 2023-05-16
Extract Safari version instead of WebKit version

### LINKED ISSUE #1082: Safari <=12.0 media establishing does not work due to `removeExtmapAllowMixed` shim not being applied
# Please read first!

## Versions affected

**Safari 12.0**

**adapter.js 7.7.0, 7.7.1, 8.0**

## Description

`removeExtmapAllowMixed` shim is not applied to iOS Safari 12.
Based on this [PR's comments](https://github.com/webrtcHacks/adapter/pull/1042) the shim should be applied to Safari < 13.1, however it is not applied in iOS Safari 12.

I think the culprit is this check https://github.com/webrtcHacks/adapter/blob/master/src/js/common_shim.js#L327

```javascript
  if (browserDetails.browser === 'safari' && browserDetails.version >= 605) {
    return;
  }
```

`browserDetails.version` returns `AppleWebKit` version and this version is 605 in both iOS Safari 12 and iOS Safari 13.
For example here is UserAgent from iOS Safari 12:
`Mozilla/5.0 (iPhone; CPU iPhone OS 12_1 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/12.0 Mobile/15E148 Safari/604.1`
here is UserAgent from iOS Safari 13:
`Mozilla/5.0 (iPhone; CPU iPhone OS 13_2_3 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/13.0.3 Mobile/15E148 Safari/604.1`

Based on local testing this shim is only needed in Safari 12.0 and below, as all works correctly in Safari 12.1+

## Steps to reproduce
1. Go to https://appr.tc/ in Chrome 91 and create a room
2. Go to that room from iOS Safari 12, allow camera, in console paste the latest adapter release https://github.com/webrtcHacks/adapter/blob/master/release/adapter.js. This is because `https://appr.tc/` does not use the newest version at the moment.
3. In iOS Safari press Join

## Expected results
2-way video is established

## Actual results
Error: `setRemoteDescription: OperationError: Expects at least 2 fields.`
Screenshot 2021-06-22 at 09 23 04

lukasIO:
**Description**
This PR changes the version extraction for Safari to look at the actual Safari version instead of the WebKit version.
Minor versions (e.g. `10.3`) get parsed as float in order to keep straight forward version comparison as before.
In case there would be a patch version defined in the ua string (e.g. `10.3.4`) the patch version gets dropped. 

The oldest Safari versions I could find still work with that approach:
- macOS Safari 4.0 - `Mozilla/5.0 (Macintosh; U; Intel Mac 05 X 10_6_8; en-us) ApplewebKit/531.22.7 (KHTML, like Gecko) Version/4.0.5 Safari/531.22.7`
- iOS Safari 7.0 - `Mozilla/5.0 (iPhone; CPU iPhone OS 7_1 like Mac OS X) AppleWebKit/537.51.2 (KHTML, like Gecko) Version/7.0 Mobile/11D167 Safari/9537.53`

**Purpose**
fixes #1082

Safari has stopped updating WebKit versions in the user agent string. E.g. Safari 16.3 still shows `AppleWebKit/605.1.15`.

### DIFF
--- a/src/js/utils.js
+++ b/src/js/utils.js
@@ -177,7 +177,7 @@ export function detectBrowser(window) {
       navigator.userAgent.match(/AppleWebKit\/(\d+)\./)) { // Safari.
     result.browser = 'safari';
     result.version = extractVersion(navigator.userAgent,
-      /version\/(\d+(\.?\d))/i, 1);
+      /version\/(\d+(\.?\d+))/i, 1);
     result.supportsUnifiedPlan = window.RTCRtpTransceiver &&
         'currentDirection' in window.RTCRtpTransceiver.prototype;
   } else { // Default fallthrough: not supported.
--- a/src/js/utils.js
+++ b/src/js/utils.js
@@ -21,7 +21,11 @@ let deprecationWarnings_ = true;
  */
 export function extractVersion(uastring, expr, pos) {
   const match = uastring.match(expr);
-  return match && match.length >= pos && parseInt(match[pos], 10);
+  if (match === null) {
+    return null;
+  }
+  const versionParts = match[pos].split('.').slice(0, 2);
+  return match && match.length >= pos && parseFloat(versionParts.join('.'), 10);
 }
 
 // Wraps the peerconnection event eventNameToWrap in a function
@@ -177,7 +181,8 @@ export function detectBrowser(window) {
       navigator.userAgent.match(/AppleWebKit\/(\d+)\./)) { // Safari.
     result.browser = 'safari';
     result.version = extractVersion(navigator.userAgent,
-      /AppleWebKit\/(\d+)\./, 1);
+      /version\/(\d+(\.?_?\d+)+)/i, 1);
+    console.log(result.version);
     result.supportsUnifiedPlan = window.RTCRtpTransceiver &&
         'currentDirection' in window.RTCRtpTransceiver.prototype;
   } else { // Default fallthrough: not supported.
--- a/src/js/utils.js
+++ b/src/js/utils.js
@@ -21,11 +21,7 @@ let deprecationWarnings_ = true;
  */
 export function extractVersion(uastring, expr, pos) {
   const match = uastring.match(expr);
-  if (match === null) {
-    return null;
-  }
-  const versionParts = match[pos].split('.').slice(0, 2);
-  return match && match.length >= pos && parseFloat(versionParts.join('.'), 10);
+  return match && match.length >= pos && parseFloat(match[pos], 10);
 }
 
 // Wraps the peerconnection event eventNameToWrap in a function
@@ -181,8 +177,7 @@ export function detectBrowser(window) {
       navigator.userAgent.match(/AppleWebKit\/(\d+)\./)) { // Safari.
     result.browser = 'safari';
     result.version = extractVersion(navigator.userAgent,
-      /version\/(\d+(\.?_?\d+)+)/i, 1);
-    console.log(result.version);
+      /version\/(\d+(\.?\d))/i, 1);
     result.supportsUnifiedPlan = window.RTCRtpTransceiver &&
         'currentDirection' in window.RTCRtpTransceiver.prototype;
   } else { // Default fallthrough: not supported.
--- a/src/js/utils.js
+++ b/src/js/utils.js
@@ -177,7 +177,7 @@ export function detectBrowser(window) {
       navigator.userAgent.match(/AppleWebKit\/(\d+)\./)) { // Safari.
     result.browser = 'safari';
     result.version = extractVersion(navigator.userAgent,
-      /version\/(\d+(\.?\d+))/i, 1);
+      /Version\/(\d+(\.?\d+))/i, 1);
     result.supportsUnifiedPlan = window.RTCRtpTransceiver &&
         'currentDirection' in window.RTCRtpTransceiver.prototype;
   } else { // Default fallthrough: not supported.
--- a/test/unit/extractVersion.js
+++ b/test/unit/extractVersion.js
@@ -117,10 +117,17 @@ describe('extractVersion', () => {
 
   describe('Safari regular expression', () => {
     const expr = /version\/(\d+(\.?_?\d+)+)/i;
-    it('matches the webkit version', () => {
+    it('matches the Safari version', () => {
       ua = 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12_4) ' +
           'AppleWebKit/604.1.6 (KHTML, like Gecko) Version/10.2 Safari/604.1.6';
       expect(extractVersion(ua, expr, 1)).to.equal(10.2);
+    });
+
+    it('drops patch version numbers', () => {
+      ua = 'Mozilla/5.0 (iPhone; CPU iPhone OS 7_1 like Mac OS X) ' +
+          'AppleWebKit/537.51.2 (KHTML, like Gecko) Version/7.0.4 ' +
+          'Mobile/11D167 Safari/9537.53';
+      expect(extractVersion(ua, expr, 1)).to.equal(7.0);
     });
 
     it('matches the iphone simulator', () => {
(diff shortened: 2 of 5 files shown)


fippo:
🤔 this would be a breaking change (which means losing 30% of users, see #1065)
I don't think the safari/webkit version is important enough for that, I never gated features on it.

lukasIO:
thanks for the review!

> I never gated features on it

maybe I'm misunderstanding what you mean by that, but there's https://github.com/webrtcHacks/adapter/blob/main/src/js/common_shim.js#L356 and #1082 which details an uncaught (by adapter) bug related to the version parsing for Safari 12.

lukasIO:
@fippo was not including this in 9.0.0 an oversight or are you not planning to merge this altogether?

fippo:
I think the main concern here is still that even with a breaking change this would folks to update their code... which will reduce the adoption rate (not great already). #1160 is the next breaking change so this might ride along. But the alternative would be to expose the Safari version with a different name and update the adapter code?

src/js/utils.js
```diff
@@ -177,7 +177,7 @@ export function detectBrowser(window) {
       navigator.userAgent.match(/AppleWebKit\/(\d+)\./)) { // Safari.
     result.browser = 'safari';
     result.version = extractVersion(navigator.userAgent,
-      /AppleWebKit\/(\d+)\./, 1);
+      /version\/(\d+(\.?\d+))/i, 1);
```
uxmaster: Why is this Reg Exp case-insensitive?
The version always appears as `Version` in the UA string.

test/unit/extractVersion.js
```diff
@@ -116,30 +116,37 @@ describe('extractVersion', () => {
   });
 
   describe('Safari regular expression', () => {
-    const expr = /AppleWebKit\/(\d+)/;
-    it('matches the webkit version', () => {
+    const expr = /version\/(\d+(\.?\d+))/i;
```
uxmaster: Same here.

uxmaster (commented):
Ockham razor ;-)

uxmaster:
BTW @fippo, I think nobody uses the adapter's Safari version, since it is useless being always 605 (I parse it myself for Safari while for other browsers I lean on adapter). Thus, this is not so breaking change as it looks.

fippo:
I know it is useless but changing such behavior is a breaking change in semver for better or worse.

lukasIO:
> I think the main concern here is still that even with a breaking change this would folks to update their code... 

I don't see how that's true if the current behaviour is just broken/useless as you mention. No one using a adapter today can rely on the Safari version parsing, so I don't see why they would have to change their code 🤔 

> But the alternative would be to expose the Safari version with a different name and update the adapter code?

Totally your call, I think everything is better than it not being useful at all, but exposing a custom property for the safari version (not sure if you had something else in mind) sounds less obvious to me than fixing the version parsing

dagingaa:
Sorry for taking so long on weighing in here. I'd like to avoid feature gating based on user agent parsing altogether, and I think adding the proper Safari version now without a clear benefit to adapter is a breaking change I'd like to avoid.

Not saying we won't need this in the future, but until then I'd like us to keep it as stable as possible given the intention of this project. The bug you mentioned is valid, but also has a clear workaround. If anything that bug should be fixed by feature testing and fallback rather than user agent parsing to begin with, but I might be able to be swayed on that.

In any case, I love the contribution, and think we should leave it open. As @fippo mentioned we might include this with the next breaking change, or as part of fixing the above-mentioned bug.

lukasIO:
yeah, definitely agree that this should only be included in a major version update! 

An initial intention was to use adapter's browser parsing which is mentioned in the project's README: 

> You might want to use adapters browser detection which detects which webrtc quirks are required. ... for webrtc engine detection (which will for example detect Opera or the Chromium based Edge as 'chrome') and `adapter.browserDetails.version` for the version according to the user-agent string.

and that's not very useful for Safari as is.

dagingaa:
Yeah I think we need to remove that from the README, from when this project was created to now, a lot has happened to browser detection, user agent strings etc. to the point where it's much better to use a library dedicated to this, rather than relying on adapter.

Still we need to support the legacy for a while longer, and probably do a major release at some point that will remove exposing these fields and provide a good alternative route for people that rely on this today.

uxmaster:
@dagingaa, it seems you concentrate on a global view, while I think this should be treated just as a fix for #1082. When you fix the issue reported almost 4 years ago, you may then wonder about possible refactoring. First and foremost, the lib should just work, not be politically correct (according to not-use-the-user-agent-string orthodoxes).

dagingaa:
> not be politically correct (according to not-use-the-user-agent-string orthodoxes).

The user agent string is not a reliable method to get information anymore, and thus should not be used as such. Further, relying on adapter for this is backwards and will actually just hurt your application rather than help it when there are good alternatives.

Internally, adapter needs to use any means necessary to add workarounds, such as checking the user agent string. We do that today, and will continue to do that tomorrow. However, exposing that to end users is a mistake I personally think we should stop doing. Changing the detection code everywhere means additional risk which I think is unnecessary.

> it seems you concentrate on a global view, while I think this should be treated just as a fix for https://github.com/webrtcHacks/adapter/issues/1082.

This PR is changing the global view, thus we need to care about it (for the aforementioned reasons). Fixing #1082 in isolation should be a much simpler fix, which we very much welcome, but introducing a major release breaking change is a huge ask which does need to be taken into very careful consideration.

In the meantime we are lucky there exists a workaround.

fippo:
#1082 is an issue that was fixed in libWebRTC [back in late 2018](https://chromiumdash.appspot.com/commits?commit=211856b956d8099ef99b9dbe16ce77dcf78d0bb4). Safari has been slow updating it but that means 2019 in Safari 13 still. The issue was marginal in 2021 already, in 2025 I doubt it is relevant? (and if not my condolences to the users on that browser)

Using adapter for UA detection was convenient in the early days (see the [2015 version here](https://github.com/webrtcHacks/adapter/blob/c30d0da53aa96dccc85255c6ee3f28dbad42a651/samples/web/js/adapter.js)), in particular since one wanted the "webrtc version" which was the same in Chrome and Opera. I haven't seen much use of that in recent years (and even the samples [use it sparingly](https://github.com/search?q=repo%3Awebrtc%2Fsamples+browserdetails&type=code)). For the most part one can write decent feature detection (please do not create a peerconnection for that... well, if you do in Safari I don't mind)

Is relying on adapters version something you should do in ~2023~2025? Probably not👎 
If version detection is broken for Safari and that discourages usage of adapter for this purpose then 👍
Do it yourself (like [here](https://github.com/livekit/client-sdk-js/pull/1117)) 👍

kor0000:
> **Description**
> This PR changes the version extraction for Safari to look at the actual Safari version instead of the WebKit version.
> Minor versions (e.g. `10.3`) get parsed as float in order to keep straight forward version comparison as before.
> In case there would be a patch version defined in the ua string (e.g. `10.3.4`) the patch version gets dropped. 
> 
> The oldest Safari versions I could find still work with that approach:
> - macOS Safari 4.0 - `Mozilla/5.0 (Macintosh; U; Intel Mac 05 X 10_6_8; en-us) ApplewebKit/531.22.7 (KHTML, like Gecko) Version/4.0.5 Safari/531.22.7`
> - iOS Safari 7.0 - `Mozilla/5.0 (iPhone; CPU iPhone OS 7_1 like Mac OS X) AppleWebKit/537.51.2 (KHTML, like Gecko) Version/7.0 Mobile/11D167 Safari/9537.53`
> 
> **Purpose**
> fixes #1082
> 
> Safari has stopped updating WebKit versions in the user agent string. E.g. Safari 16.3 still shows `AppleWebKit/605.1.15`. 
> 
>

fippo:
#1161 is now part of v9.0.2

PR #1131, opened 2023-06-14
feat: add support for react-native

anion155:
**Description**
Handle react-native environment, without any shims.

**Purpose**
https://github.com/webrtcHacks/adapter/issues/1129


saghul:
Why is this necessary? It already does nothing on RN. Are you running into any problem?

anion155:
Yep. It throws in react native, 'cause there is no `navigator.userAgent`. So any module that using `webrtc-adapter` internally is practically unusable without patching.

saghul:
Gotcha. @fippo I guess adapter shouldn't do anything if `navigator.userAgent` is not a thing.

fippo:
#1132 seems a bit more minimal for that and not specific to RN

fippo:
Can you please give the new version 8.2.3 a try?

PR #1133, opened 2023-06-15
use @puppeteer/browsers instead of travis-multirunner

fippo:
which reduces the maintenance burden


fippo:
Almost there, need to bring back Firefox though.

saghul (approved):
Nice!

fippo:
Lets do this, nightly-only should be ok

PR #1146, opened 2024-03-07
prefer navigator.userAgentData for browser detection

### LINKED ISSUE #764: Chrome will be identified as Safari if webkitGetUserMedia is removed

fippo:
fixes #764,#1017

(previously known as #1103 but needed a new PR due to the switch from master to main)

### DIFF
--- a/src/js/utils.js
+++ b/src/js/utils.js
@@ -161,6 +161,16 @@ export function detectBrowser(window) {
 
   const {navigator} = window;
 
+  // Prefer navigator.userAgentData.
+  if (navigator.userAgentData && navigator.userAgentData.brands) {
+    const chromium = navigator.userAgentData.brands.find((brand) => {
+      return brand.brand === 'Chromium';
+    });
+    if (chromium) {
+      return {browser: 'chrome', version: chromium.version};
+    }
+  }
+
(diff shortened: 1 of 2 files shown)


fippo:
@yoavweiss and @miketaylr does that look forward-compatible?

(I'll yell at the tests separately...)

src/js/utils.js
```diff
@@ -161,6 +161,16 @@ export function detectBrowser(window) {
 
   const {navigator} = window;
 
+  // Prefer navigator.userAgentData.
+  if (navigator.userAgentData && navigator.userAgentData.brands) {
```
miketaylr, line 165: If you wanted to get fancy, you could do something like `navigator.userAgentdata?.brands`

miketaylr (approved):

PR #1147, opened 2024-03-24
update test setup and bring back chai as a dependency

fippo:
using chai@3.5.0 which is an older version that still works with karma-chai

### DIFF
--- a/.github/workflows/e2e-tests.yaml
+++ b/.github/workflows/e2e-tests.yaml
@@ -13,8 +13,8 @@ jobs:
           - browser: firefox
             version: nightly
     steps:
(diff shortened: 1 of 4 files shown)


fippo:
@jan-ivar @pehrsons it seems FF Nightly rejects modifications to RTCSessionDescription.sdp.
This has been tried in Chromium before and had to be pulled out. The test fix here is trivial but you may want to communicate this better than "we added a console warning" (which does not even have a bugzilla link)

https://webrtchacks.github.io/chromestatus/?buckets=1676,3451 shows data. This is ~1/3 of setRemoteDescription

PR #1152, opened 2024-03-28
Update browserDetails.version to return int

### LINKED ISSUE #1151: browserDetails.version returns string after PR#1150
- [X] I have provided browser name, version and adapter.js version
- [X] This issue only happens when adapter.js is used

## Versions affected 

**Chrome and Edge Browser all versions

**adapter.js after PR#1150

## Description
browserDetails.version returns string after PR#1150

xingri:
fixes #1151

### DIFF
--- a/src/js/utils.js
+++ b/src/js/utils.js
@@ -167,7 +167,7 @@ export function detectBrowser(window) {
       return brand.brand === 'Chromium';
     });
     if (chromium) {
-      return {browser: 'chrome', version: parseInt(chromium.version)};
+      return {browser: 'chrome', version: parseInt(chromium.version, 10)};
     }
   }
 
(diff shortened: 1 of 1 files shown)


src/js/utils.js
```diff
@@ -167,7 +167,7 @@ export function detectBrowser(window) {
       return brand.brand === 'Chromium';
     });
     if (chromium) {
-      return {browser: 'chrome', version: chromium.version};
+      return {browser: 'chrome', version: parseInt(chromium.version)};
```
fippo:
```suggestion
      return {browser: 'chrome', version: parseInt(chromium.version, 10)};
```

fippo:
Good catch, thank you!

fippo (approved):

fippo:
@miketaylr let me guess, this returned the full version because it used to be a string before the UA reduction and now we can't change the API anymore? ;-)

miketaylr:
> and now we can't change the API anymore? ;-)

heh, pretty much. besides, everybody loves strings!

PR #1160, opened 2025-02-27
Detect browser without relying on vendor prefixes

### LINKED ISSUE #1156: Firefox will be identified as "Not a supported browser." if mozGetUserMedia is removed
https://github.com/webrtcHacks/adapter/issues/764 identified problematic `detectBrowser` code which prevents browsers from removing non-standard APIs:
```js
    if (navigator.mozGetUserMedia) { // Firefox.
      result.browser = 'firefox';
          /Firefox\/(\d+)\./, 1);
    } else if (navigator.webkitGetUserMedia) {
      // Chrome, Chromium, Webview, Opera.
      // Version matches Chrome/WebRTC version.
      result.browser = 'chrome';
          /Chrom(e|ium)\/(\d+)\./, 2);
    } else if (navigator.mediaDevices &&
        navigator.userAgent.match(/Edge\/(\d+).(\d+)$/)) { // Edge.
      result.browser = 'edge';
          /Edge\/(\d+).(\d+)$/, 2);
    } else if (navigator.mediaDevices &&
        navigator.userAgent.match(/AppleWebKit\/(\d+)\./)) { // Safari.
      result.browser = 'safari';
          /AppleWebKit\/(\d+)\./, 1);
    } else { // Default fallthrough: not supported.
      result.browser = 'Not a supported browser.';
      return result;
    }
```
E.g. Firefox cannot remove `mozGetUserMedia` without this returning "Not a supported browser."

The https://github.com/webrtcHacks/adapter/pull/1146 fix to https://github.com/webrtcHacks/adapter/issues/764 is unsatisfactory, relying on an [API that is not standards-track](https://wicg.github.io/ua-client-hints/#dom-navigatorua-useragentdata) to solve this problem only for Chromium.

The core problem, that this function treats browsers without specific non-standard APIs as "Not a supported browser." remains unaddressed.

jan-ivar:
Fixes #1156.

### DIFF
--- a/src/js/utils.js
+++ b/src/js/utils.js
@@ -149,16 +149,11 @@ export function deprecated(oldMethod, newMethod) {
  *     properties.
  */
 export function detectBrowser(window) {
-  // Returned result object.
-  const result = {browser: null, version: null};
-
   // Fail early if it's not a browser
   if (typeof window === 'undefined' || !window.navigator ||
       !window.navigator.userAgent) {
-    result.browser = 'Not a browser.';
-    return result;
+    return {browser: 'Not a browser.', version: null};
   }
-
   const {navigator} = window;
 
   // Prefer navigator.userAgentData.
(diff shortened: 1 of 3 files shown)


fippo:
isn't waiting for https://bugzilla.mozilla.org/show_bug.cgi?id=1750143 easier?
That will not require a major version bump which traditionally takes years (see #1065)

jan-ivar:
Merging this PR should have the same net effect (of making the check UA-string based).

It might also end up being faster given Mozilla's [neutral](https://mozilla.github.io/standards-positions/#ua-client-hints) position on UA client hints.

I should have done this a year ago when we discovered this, but we gotta plant the seed at some point.

jan-ivar:
Firefox (Nightly) 138 now has `media.navigator.mozgetusermedia.enabled` pref in about:config that if you set it to `false` causes adapter to return `"Not a supported browser."`.

While most major video conferencing websites I've tested so far appear to work, we're in contact with one affected by this. They're applying a fix on their end, but getting this merged would probably help them and other consumers of adapter.

fippo:
Just trying to get it into the same release as #1127

https://www.reddit.com/r/firefox/comments/1j8401w/another_media_service_fallen_f1tv_is_a_costly/?share_id=qEAEN4EyxZnU9fGVFNaDI&utm_content=2&utm_medium=android_app&utm_name=androidcss&utm_source=share&utm_term=1&rdt=56689 should have demonstrated that UA sniffing is far more costly to Firefox than helping do actual feature detection but...

jan-ivar:
Thanks for the link. It's a good example of why we want to get rid of things like `mozGetUserMedia`.

Streamlining UA sniffing helps us respond quickly with out-of-band interventions or dot-releases, which might need to include overriding UA strings in cases where websites reactively blocked Firefox.

kor0000:
> Fixes #1156.

fippo:
rebase needed

fippo:
@jan-ivar ^

I still think that providing a minimal UserAgentData that saves developers from the pain of parsing the UA string would be easier to adopt.

PR #1174, opened 2026-02-04
fix: add null check for localDescription getter

wbxl2000:
## Summary

Fix `TypeError: Cannot read properties of null (reading 'type')` when accessing `localDescription` in Chrome DevTools iOS device simulation.

## Reproduction

1. Open Chrome DevTools → Toggle device toolbar → Select iPhone/iPad
2. Load any page using webrtc-adapter
3. Access `RTCPeerConnection.localDescription` before connection is established

**Live demo:**
- Bug: https://bug.qer.im/chrome-webrtc-adapter-localDescription-null-in-ios-emulator/original
- Fixed: https://bug.qer.im/chrome-webrtc-adapter-localDescription-null-in-ios-emulator/fixed

## Root Cause

When simulating iOS devices, Chrome DevTools changes the UA string but keeps browser APIs intact. This causes:
1. Browser detected as Chrome (via `navigator.webkitGetUserMedia`)
2. Version parsing fails (UA has no `Chrome/XX` pattern) → `version: NaN`
3. `NaN >= 65` is `false` → falls into OLD shim path
4. OLD path wraps `localDescription` getter without null check → `null.type` throws

## Fix

Add null check before accessing `description.type`:

```diff
- if (description.type === '') {
+ if (!description || description.type === '') {
```

## Test Plan

- [x] Added unit test for null `localDescription` scenario
- [x] Verified fix with live demo in Chrome DevTools iOS simulation


fippo:
Interesting find! This only happens in devtools? Chrome on iOS should not expose webkitGetUserMedia.
From what I recall those Chrome versions returns `{type: '', sdp: ''}` for pc.localDescription before it was set.

I think the fix should be different though, bailing out on NaN [here](https://github.com/webrtcHacks/adapter/blob/main/src/js/adapter_factory.js#L45) or [here](https://github.com/webrtcHacks/adapter/blob/main/src/js/utils.js#L186)

wbxl2000:
Thanks for the review! You're right, I've updated the fix to `utils.js` as you suggested:                                                                                                       
                                                                                   
  ```diff                                                                                                                                                                                         
    result.version = parseInt(extractVersion(navigator.userAgent,                                                                                                                                 
  -   /Chrom(e|ium)\/(\d+)\./, 2));
  +   /Chrom(e|ium)\/(\d+)\./, 2)) || null;
  ```
  
This converts NaN to null so the existing version === null guard in adapter_factory.js bails out correctly. The localDescription null check in chrome_shim.js has been reverted.

And yes, this only happens in DevTools device emulation.

fippo:
Almost there, can you move the test to test/unit/detectBrowser.test.js please? Should be a lot simpler:
```
  it('does not misdetect Chrome devtools something something', () => {
    navigator.userAgent = 'Mozilla/5.0 (iPhone; CPU iPhone OS 18_5 ' +
        'like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) ' +
        'Version/18.5 Mobile/15E148 Safari/604.1';
    navigator.webkitGetUserMedia = function() {};
    window.webkitRTCPeerConnection = function() {};

    const browserDetails = detectBrowser(window);
    expect(browserDetails.browser).toEqual('chrome');
    expect(browserDetails.version).toEqual(null);
  });
```

wbxl2000:
Sure, done!

fippo:
Thank you, Will merge and publish a new release next week!

fippo (approved):

fippo:
Published as part of 9.0.4

Issue #1177, opened 2026-03-06
wrapPeerConnectionEvent crashes application when addEventListener is non-writable


kalvin807:
# Please read first!
Please use [discuss-webrtc](https://groups.google.com/forum/#!forum/discuss-webrtc) for general technical discussions and questions.

- [x] I have provided steps to reproduce (e.g. a link to a [jsfiddle](https://jsfiddle.net/))
- [x] I have provided browser name, version and adapter.js version
- [x] This issue only happens when adapter.js is used

**Note: If the checkboxes above are not checked (which you do after the issue is posted), the issue will be closed.**

## Versions affected

**Browser name including version (e.g. Chrome 64.0.3282.119)**
Any browser I believes, as long as it is a wix site (I reproduced with firefox/chrome/safari)

**adapter.js (e.g. 6.1.0)**
9.0.4

## Description
`wrapPeerConnectionEvent` in `src/js/utils.js` crashes when `EventTarget.prototype.addEventListener` is made non-writable by the host environment (e.g. Wix's site). Since `adapterFactory()` runs all shims immediately on import, this crashes the application at load time if they import this package.

We are currently have a working around by patching with a try-catch guard before the assignment

```js
const nativeAddEventListener = proto.addEventListener;
try { proto.addEventListener = nativeAddEventListener; }
catch (e) { return; }
```
 
- related downstream issue [livekit/client-sdk-js#1806](https://github.com/livekit/client-sdk-js/issues/1806)

## Steps to reproduce

  1. Open DevTools console on any page
  2. Go to one of the wix's example site (https://www.wix.com/explore/websites)
  3. Load adapter.js:
```js
  var s = document.createElement('script');
  s.src = 'https://unpkg.com/webrtc-adapter@9.0.4/out/adapter.js';
  document.head.appendChild(s);
  // → TypeError: Cannot assign to read only property 'addEventListener'
```

## Expected results

adapter.js loads without error, gracefully skipping patches it cannot apply.

## Actual results

TypeError: Cannot assign to read only property 'addEventListener' of object '#<RTCPeerConnection>'
Crashes at module initialization, prevents any library that bundles adapter.js from loading.

fippo:
From inserting a
```
Object.defineProperty(EventTarget.prototype, 'addEventListener', {
    writable: false
 });
```
into the test page it looks like this happens in three places:
* Chrome shim's  shimOnTrack which should probably not attempt to do anything in versions that unshipped plan-b (M102)
[...]

PR #1178, opened 2026-03-06
use stricter guards for some ancient polyfills

fippo:
use stricter guards for some ancient polyfills

* shimOnTrack which is no longer useful in unified-plan which is the only remaining mode as of M102.
* fixNegotiationNeeded which is working in unified-plan
* shimSendThrowTypeError which is no longer required in Firefox60+. Still required in Chrome and Safari as of WebRTC M148, fixed in M149.

Also do not override the event listener if it is not writable which avoids (but does not fix) #1177

### DIFF
--- a/src/js/chrome/chrome_shim.js
+++ b/src/js/chrome/chrome_shim.js
@@ -15,7 +15,11 @@ export function shimMediaStream(window) {
   window.MediaStream = window.MediaStream || window.webkitMediaStream;
 }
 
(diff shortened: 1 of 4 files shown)


fippo:
https://jsfiddle.net/fippo/dftamhpq/1/ -- turns out Chrome is not spec compliant and thus does not throw an error...

fippo:
should ride along with #1180 on a patch version for easy upgrade

PR #1180, opened 2026-04-10
firefox: do not throw when calling getStats on a closed peerconnection

### LINKED ISSUE #1179: Firefox guard getStats() shim against closed PeerConnection
- [x] I have provided steps to reproduce ([CodeSandbox](https://codesandbox.io/p/sandbox/lp8pf4))
- [ ] This issue only happens when adapter.js is used

## Versions affected

Firefox 149.0.2 (aarch64). Chrome is not affected.

**adapter.js (9.0.4)**

## Description

Calling `getStats()` on a closed PeerConnection in Firefox returns a rejected promise with a `DOMException: An attempt was made to use an object that is not, or is no longer, usable`. This is native Firefox behavior — it happens with and without adapter.js.

However, since the adapter already shims `getStats` in `shimPeerConnection` (`src/js/firefox/firefox_shim.js`, line 61), it could normalize this by returning an empty `Map` instead of propagating the rejection — the same pattern already used in `shimSenderGetStats` for the no-track case:

```javascript
// shimSenderGetStats already returns empty Map when there's no track:
return this.track ? this._pc.getStats(this.track) : Promise.resolve(new Map());
```

This is common in applications that tear down and recreate PeerConnections (room transitions, reconnection) while polling stats on a timer. The Chrome shim already has `signalingState === 'closed'` guards on `setLocalDescription` (line 429) and `setRemoteDescription` (line 555).

## Steps to reproduce

[CodeSandbox](https://codesandbox.io/p/sandbox/lp8pf4)

## Expected results

`getStats()` on a closed PeerConnection should not throw.

## Actual results

Unhandled promise rejection:

```
DOMException: An attempt was made to use an object that is not, or is no longer, usable
    getStats firefox_shim.js:63
```

### Suggested fix

Add a `connectionState` guard to `shimPeerConnectionGetStats` — the single point all three getStats shims (`shimPeerConnectionGetStats`, `shimSenderGetStats`, `shimReceiverGetStats`) flow through:

```javascript
window.RTCPeerConnection.prototype.getStats = function getStats() {
  const [selector, onSucc, onErr] = arguments;

  if (this.connectionState === 'closed') {
    if (onSucc) {
      onSucc(new Map());
    }
    return Promise.resolve(new Map());
  }

  return nativeGetStats.apply(this, [selector || null])
    .then(stats => {
      // ... existing stat type normalization ...
    })
    .then(onSucc, onErr);
};
```

fippo:
which is fixed in Firefox 151+, see
  https://bugzilla.mozilla.org/show_bug.cgi?id=1056433

Fixes #1179

### DIFF
--- a/src/js/adapter_factory.js
+++ b/src/js/adapter_factory.js
@@ -86,6 +86,7 @@ export function adapterFactory({window} = {}, options = {
 
       firefoxShim.shimGetUserMedia(window, browserDetails);
       firefoxShim.shimPeerConnection(window, browserDetails);
+      firefoxShim.shimGetStats(window, browserDetails);
       firefoxShim.shimOnTrack(window, browserDetails);
       firefoxShim.shimRemoveStream(window, browserDetails);
       firefoxShim.shimSenderGetStats(window, browserDetails);
--- a/src/js/firefox/firefox_shim.js
+++ b/src/js/firefox/firefox_shim.js
@@ -48,6 +48,17 @@ export function shimPeerConnection(window, browserDetails) {
         window.RTCPeerConnection.prototype[method] = methodObj[method];
       });
   }
+}
+
+export function shimGetStats(window, browserDetails) {
+  if (typeof window !== 'object' ||
+      !(window.RTCPeerConnection || window.mozRTCPeerConnection)) {
+    return; // probably media.peerconnection.enabled=false in about:config
+  }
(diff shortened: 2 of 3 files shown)


fippo:
nightly failure should resolve itself :soon:

fippo:
Fixed by https://phabricator.services.mozilla.com/D293467

PR #1181, opened 2026-04-16
Bump version to 9.0.5

fippo:
picking up #1178 and #1180 which makes adapter more obsolete (yay)


fippo:
will need to wait until chrome dev is at least 149.0.7783.0 to pick up https://chromiumdash.appspot.com/commit/30f3df63e7c9f6d81bfbd429552984b541e35651

Issue #1182, opened 2026-04-18
determine how to use polyfills less


fippo:
currently adding logging before "native" calls still shows some places where a polyfill is used (Chrome here)
```
     13 ERROR: 'adapter.js: native addStream called (shimAddTrackRemoveTrackWithNative)'
     35 ERROR: 'adapter.js: native addTrack called (shimAddTrackRemoveTrackWithNative)'
     17 ERROR: 'adapter.js: native removeTrack called (shimAddTrackRemoveTrackWithNative)'
      3 ERROR: 'adapter.js: native getStats called (shimSenderReceiverGetStats)'
      7 ERROR: 'adapter.js: native createDataChannel called (shimSendThrowTypeError)'
      1 ERROR: 'adapter.js: native RTCDataChannel.send called (shimSendThrowTypeError)'
     56 ERROR: 'adapter.js: native setRemoteDescription called (shimMaxMessageSize)'
```

`shimSendThrowTypeError` will be obsolete as of Chrome 149.
`shimMaxMessageSize` needs some guards as it should be obsolete but needs a version check - unified plan (M105?) for Chrome (done in #1187 ) and [FF](https://bugzilla.mozilla.org/show_bug.cgi?id=2019361)
`shimSenderReceiverGetStats` is the `shim RTCPeerConnection.getStats(track)` part which should be obsolete as of [M67](https://chromiumdash.appspot.com/commits?commit=ac7139070a15296cc490158eb9dee7d23f2165a4&platform=Windows) which also shipped [sender/receiver getStats](https://chromiumdash.appspot.com/commits?commit=fe8a5e81a3b68c6b3c388728288591165a5a8c63&platform=Windows), fixed by #1183

fippo:
Firefox is down to shimMaxMessageSize after #1185

fippo:
Chrome's [getUserMedia polyfill](https://github.com/webrtcHacks/adapter/blob/main/src/js/chrome/getusermedia.js) also seems largely obsolete as of M64. The stuff at the bottom is related to calling `GUM({audio: true, video: true})` which could return a stream without video if no camera was present. From the [PSA](https://groups.google.com/g/discuss-webrtc/c/DMPTKcXVUQ4/m/85FaASoFBwAJ?) (h/t @guidou) that also shipped in M64 back in 2017, #1186 takes care of that.

fippo:
So shimAddTrackRemoveTrackWithNative was added in #557. In a nutshell it does this:
```
[...]

PR #1187, opened 2026-04-20
chrome: don't polyfill .sctp and max-message-size

fippo:
in Chrome 102+ with only unified plan

See #1182


fippo:
The test is actually correct (surprise) and Firefox needs a little more work:
https://bugzilla.mozilla.org/show_bug.cgi?id=2019361
Thanks @docfaraday!

fippo:
Interesting, native Chrome rejects 2GB max-message size while Firefox accepts it:
```
a=max-message-size:2147483648
```

PR #1190, opened 2026-05-15
test: add safari runner

fippo:
**Description**

**Purpose**


fippo:
sad to see Safari still not being easily testable a decade later.

PR #1192, opened 2026-05-25
fix: ignore incorrect userAgentData in SLBrowser

leochliu107:
# PR Content

## Title

fix: add null check for localDescription to prevent crash on SLBrowser

### Problem

Some Chromium-based browsers (e.g. Lenovo SLBrowser) report an incorrect Chromium version in `navigator.userAgentData.brands`. For example, SLBrowser reports version `9.0.8.3131` (its own browser version) instead of the actual Chromium engine version (`141`).

This causes `detectBrowser` to return a very low version number (e.g. `9`), which triggers legacy shim code paths in `shimAddTrackRemoveTrack`. In this code path, `origLocalDescription.get()` can return `null` before any description is set, and the existing code accesses `.type` on it without a null check, resulting in:

```
TypeError: Cannot read properties of null (reading 'type')
```

**Reproduction environment:**
- Browser: SLBrowser/9.0.8.3131 (Chromium 141) — download: https://browser.lenovo.com.cn/
- UA: `Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/141.0.0.0 Safari/537.36 SLBrowser/9.0.8.3131 SLBChan/103 SLBVPV/64-bit`
- UAData brands: `Chromium/9.0.8.3131, Not?A_Brand/8.0.0.0`

### Fix

Add a null guard for `description` before accessing `.type` in the `localDescription` getter within `shimAddTrackRemoveTrack`:

```js
// Before
if (description.type === '') {

// After
if (!description || description.type === '') {
```

This is a safe and minimal fix — when `localDescription` is `null` (no local description has been set yet), we simply return `null` as-is, which is the correct WebRTC behavior.

### Testing

- Verified that existing unit tests (`npm run lint-and-unit-tests`) continue to pass.
- Manually confirmed the fix resolves the crash on SLBrowser.

### DIFF
--- a/src/js/chrome/chrome_shim.js
+++ b/src/js/chrome/chrome_shim.js
@@ -550,7 +550,7 @@ export function shimAddTrackRemoveTrack(window, browserDetails) {
     'localDescription', {
       get() {
         const description = origLocalDescription.get.apply(this);
-        if (description.type === '') {
+        if (!description || description.type === '') {
           return description;
         }
         return replaceInternalStreamId(this, description);
--- a/src/js/chrome/chrome_shim.js
+++ b/src/js/chrome/chrome_shim.js
@@ -550,7 +550,7 @@ export function shimAddTrackRemoveTrack(window, browserDetails) {
     'localDescription', {
       get() {
         const description = origLocalDescription.get.apply(this);
-        if (!description || description.type === '') {
+        if (description.type === '') {
           return description;
         }
         return replaceInternalStreamId(this, description);
--- a/src/js/utils.js
+++ b/src/js/utils.js
@@ -174,7 +174,12 @@ export function detectBrowser(window) {
       return brand.brand === 'Chromium';
     });
     if (chromium) {
-      return {browser: 'chrome', version: parseInt(chromium.version, 10)};
+      const version = parseInt(chromium.version, 10);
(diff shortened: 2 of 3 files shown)


fippo:
I don't think this is the right workaround. What about gating in Chromium >= 90 in [utils.js](https://github.com/webrtcHacks/adapter/blob/main/src/js/utils.js#L176) -- 90 is when this shipped in Chromium according to https://developer.mozilla.org/en-US/docs/Web/API/Navigator/userAgentData

cc @miketaylr @yoavweiss who probably have opinions on whether such a modification of userAgentData has wider implications and should be fixed at the source.

leochliu107:
@fippo  Thanks for the suggestion!

I've updated the PR to gate on Chromium >= 90 in utils.js — since navigator.userAgentData itself shipped in Chromium 90, any reported version below that is inherently invalid. This means we now gracefully fall through to UA string parsing in those cases.

I've also removed the !description null check in chrome_shim.js since it's no longer needed with this approach.

fippo (approved):

fippo:
Thank you @leochliu! Will release a new patch version next week.

miketaylr:
The workaround seems sensible here (and whatever SLBrowser is doing, it's just a bug they should fix).

PR #1200, opened 2026-08-27
safari : drop local streams in Safari when the last track is removed

fippo:
Safari's getLocalStreams shim kept streams around after all their tracks had been removed via removeTrack while Chrome's legacy shim drops them.


fippo:
(hopefully nobody uses or relies on this...)

